PR #197, opened 2014-04-10
Add body package_method apt_get_safe 

brontolinux:
This package method use more "prudential" options for apt/dpkg to reduce the likeliness of unexpected upgrades/removals of packages.


brontolinux:
Hei alle sammen :-) Is this going to be merged or not? I had nearly forgotten about it!

estenberg:
Can we just take it as part of the apt_get body for 3.7 instead of new body please?
Too many dead bodies already. :)

brontolinux:
@estenberg : I am not sure it is appropriate for generic use, just look at the options used with apt and you'll probably agree... I'm not going to fight over this anyway, the product is yours, I must know what's appropriate for CFEngine better than anyone.

kacfengine:
I agree that we should favor finding ways to fix the default package body rather than adding a new one, so I'll close this for now.

PR #203, opened 2014-04-14
dpackages.cf: call package managers directly

tzz:
Proposed direct mechanism for calling package managers


tzz:
Looks like this one isn't happening... closing.

PR #292, opened 2014-07-21
Add: Local user inventory module

nickanderson:
The local user inventory module parses and reports local user information as
discovered in /etc/passwd and /etc/shadow. Additional bundles are provided to
augment the Mission Portal Inventory UI with interesting information such as
all local users, users with locked accounts, and users that have emtpy
passwords.

All local user information is collected for reporting. Additionaly inventory
item for users last password change data is availbale, but disabled by default
until we get epoch time formatting in Mission Portal.

Ref: https://dev.cfengine.com/issues/6213

### DIFF
--- a/def.cf
+++ b/def.cf
@@ -224,6 +224,18 @@ bundle common inventory_control
       "mtab" string => "/etc/mtab";
       "proc" string => "/proc";
 
+      # Note: This controls local user discovery, by affects the inventory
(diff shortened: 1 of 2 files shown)


estenberg:
Content repo candidate?

nickanderson:
yes I think it is

nickanderson:
Closing this as I will just publish it seperately

PR #312, opened 2014-08-28
add shortcut masterfiles as suggested by Nick Anderson, see redmine #3697

basvandervlies:


kacfengine:
It appears @nickanderson has made it backwards-compatible, but in reviewing this I also discovered that ``shortcut`` is undocumented. Not being involved in developing that functionality, I don't know how it works, and this would be the same for our customers. So I think the documentation needs to be fixed before this can be merged (https://dev.cfengine.com/issues/7074).

kacfengine:
Sorry, my bad, it *is* documented, I was just looking in the wrong place.

kacfengine:
Merged as https://github.com/cfengine/masterfiles/commit/93ff48f09c72ca85847fbefe2d901607e8968591.

PR #339, opened 2014-10-08
Change: Make paths bundle dynamic

nickanderson:
DO NOT MERGE: THIS IS ONLY FOR CONSIDERATION AND FEEDBACK

Instead of having a list of each path that we want to know about we could find
bins based on a pre defined path, just like PATH behaves on the cli.

This is risky, possibly too risky. I am unsure if this is offset by making the
paths bundle more usable.

Additionally some bins have characters in the name that make them generate
erros. This could probably be worked around.

Currently this implementation relies on GNU find. That may be an unacceptable
dependancy.

### DIFF
--- a/lib/3.6/paths.cf
+++ b/lib/3.6/paths.cf
@@ -47,7 +47,7 @@
 # purchasing a commercial version of the software.#
 ###################################################
 
-bundle common paths
+bundle agent paths
 # @brief Defines an array `path` with common paths to standard binaries,
 # and classes for defined and existing paths.
 #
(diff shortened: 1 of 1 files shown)


tzz:
I think this should be prioritized, it would make life easier.  I would use `findfiles` though.  If it's slow, let's fix that.

nickanderson:
I tried again for a bit to use findfiles()
It is not apparent to me how I can reliably pick the first or last path listed in the event of a duplicate bin like `/bin/echo` vs `/usr/local/bin/echo`.

```
bundle agent dynamic_paths
{
  vars:
#    any::
      "PATH"
        slist => {
                  "/tmp/one/*",
                  "/tmp/three/*",
                  "/tmp/two/*",
                 };
#                  "/var/cfengine/bin",
#                  "/usr/local/sbin",
#                  "/usr/local/bin",
#                  "/usr/sbin",
#                  "/sbin",
#                  "/bin",
#                };
# 
#      "REV_PATH"
#        slist => reverse("PATH"),
#        comment => "We reverse the path so that the first listed
#                    path is what wins";

      #"paths" slist => findfiles("/var/cfengine/bin/*", "/usr/local/sbin/*", "/usr/local/bin/*", "/usr/sbin/*", "/sbin/*", "/bin/*");

      # This produces an array with the path as the index, and the
      # value as the canonified form of the path.
      "canonified_path_map[$(PATH)]" string => canonify($(PATH));

      # This produces an array with the canonified form of the path as
      # the index, and the path as the value.
      "all_paths[$(canonified_path_map[$(PATH)])]" slist => findfiles("$(PATH)");

      # Generate a list of paths for each path we are searching.
      # This list will have to be further filtered to ensure the files are executables      "paths_$(canonified_path_map[$(PATH)])" slist => { "@(all_paths[$(canonified_path_map[$(PATH)])])" }; 

      # Would need to use lastnode() and build an array of paths[lastnod()]
      # 

  reports:
    # This will report each path that is found.
    "#paths_$(canonified_path_map[$(PATH)]) = $(paths_$(canonified_path_map[$(PATH)]))";
    "paths__tmp_one__ =  $(paths__tmp_one__)";
}
```

neilhwatson:
How is find any different that setting a path?

nickanderson:
I'm wondering if we do really want this feature if it would be best implemented in the C as a language level feature. 

body agent control
{
  default_command_paths_body => no_paths;
} 

body command_paths no_paths
{
  # Current behavior 
} 

body command_paths common_paths
{
  # first one wins 
  search_path => {} 
} 

I don't know.

tzz:
IMO the path list source should be a special environment variable, not `PATH`, and should be set by the user in `def.cf` (so they can make it `PATH` if they want).

```
bundle agent main
{
  vars:
      "paths" slist => string_split(getenv("CFENGINE_SEARCH_PATH", 99k), ":", 1k);
      "pfound[$(paths)]" slist => findfiles("$(paths)/*");
      "found_all" slist => getvalues(pfound);
      "found" slist => filter(".*[[].*", found_all, true, true, 100k); # some systems have the [ executable
      "base[$(found)]" string => lastnode($(found), "/");
      "path[$(base[$(found)])]" string => $(found);

  classes:
      "have_$(base[$(found)])" expression => "any";

  reports:
      "Searching path: $(paths)";
      "Found executables: $(found) (basename $(base[$(found)]))";
      "Path for $(base[$(found)]) =  $(path[$(base[$(found)])])";
}
```

output (note the first searched path wins):

```
% CFENGINE_SEARCH_PATH=/tmp/1:/tmp/2 cf-agent -KI -f ~/Dropbox/cf/test/test_dynamic_paths.cf          [master] 
R: Searching path: /tmp/1
R: Searching path: /tmp/2
R: Found executables: /tmp/2/ping (basename ping)
R: Found executables: /tmp/1/ping (basename ping)
R: Path for ping =  /tmp/1/ping
```

IMO it's not necessary to check for executable bits.

This is fairly slow currently with large lists.

Also, IMVHO, the proper way to speed up this and many other use cases is to implement variable persistence, like we have classes persistence.  It's a common need and writing C code just for dynamic paths seems suboptimal.

nickanderson:
ENV would have to be set in policy as well wouldn't it?
https://github.com/cfengine/masterfiles/blob/master/controls/cf_agent.cf#L35

Without persistent variables currently, couldn't we cache this list of files in $(workdir)/state/paths.cache or something only regenerating that file when the software cache has changed?

tzz:
The user can choose to set the environment externally or from policy...

I think caching this list would work, just save it in a JSON file.  It's not great but would work, especially if the `bundlestate` function is accepted (because you'd save the state of the whole `paths` bundle).

nickanderson:
If ENV can be modified from outside the agent I would not use that, at least by default. Perhaps a commented out example.

tzz:
The choice of the environment variable is up to the user.  If you don't want it to be externally configurable, just make it a static slist in `def.cf`.  The operation of the code doesn't change.

estenberg:
Created a feature ticket for a native implementation here: https://dev.cfengine.com/issues/6915

Please continue the discussion there, as this pull request will not be merged anyway.

PR #361, opened 2014-11-07
Backport yum package body from 3.6 cfe packages.cf

worr:
- tested on RHEL6.5 x86_64 cfe 3.5.2

### DIFF
--- a/lib/3.5/packages.cf
+++ b/lib/3.5/packages.cf
@@ -84,6 +84,31 @@ bundle common debian_knowledge
       "patch_version_regex" string => "^Inst\s+(\S+)\s+.*";
 }
 
+bundle common rpm_knowledge
+# @depends paths
+# @brief common RPM knowledge bundle
+#
+# This common bundle has useful information about platforms using RPM
+{
+  vars:
+      "call_rpm" string => "$(paths.path[rpm])";
+
+      "rpm_output_format" string => "i | repos | %{name} | %{version}-%{release} | %{arch}\n";
+      "rpm_name_regex" string => "[^|]+\|[^|]+\|\s+([^\s|]+).*";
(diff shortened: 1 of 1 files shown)


nickanderson:
@worr ping did you see the note from @kacfengine  about `$(redhat_knowledge.check_update_postproc)`?

worr:
@nickanderson Sorry about that! Updated.

lib/3.5/packages.cf
```diff
@@ -113,6 +138,25 @@ bundle common redhat_knowledge
       "rpm3_name_regex" string => "([^.]+).*";
       "rpm3_version_regex" string => "[^\s]\s+([^\s]+).*";
       "rpm3_arch_regex" string => "[^.]+\.([^\s]+).*";
```
kacfengine: The whole section with "rpm*_regex" type of variables is now duplicated in both ``redhat_knowledge`` and ``rpm_knowledge``. I think we should stick to one, so we don't introduce subtle bugs later because one was being used instead of the other. The ``rpm_knowledge`` is the one that is used in 3.6 onwards, so it's probably more up to date.

kacfengine:
Otherwise I think this is a good change.

kacfengine:
Closing due to unfixed issues, and not a very urgent bugfix since it is 3.5 only. Please reopen if you want to keep working on this.

worr:
`rpm3_*` is not duplicated. Also, I didn't want to break backwards compat by removing variables that *may* be used by consumers. I can resubmit with that removed, if you still think they should be removed.

kacfengine:
Yes, I think they should be removed. They are very unlikely to be used outside this file, so it's better to avoid duplication I think.

PR #362, opened 2014-11-11
Backport fixes to getfacl, npm, and pip paths

nickanderson:


nickanderson:
Looks like this is already in 3.6.x

PR #363, opened 2014-11-14
Initial tagged test

estenberg:


kacfengine:
@estenberg, do we still want this? I haven't tested if this passes, but the process cleanup has been implemented, so we don't risk anything by trying.

estenberg:
Yes, please. This should auto-visualize now with the tool of @pasinskim

kacfengine:
trigger build

kacfengine:
The test passes, but the test harness fails afterwards because you haven't cleaned up the processes you created. You should modify the test to stop the service again after you have checked the pass conditions.

kacfengine:
> process cleanup has been implemented, so we don't risk anything by trying.

And btw, just in case you are wondering about my seemingly contradictory statements about process cleanup: The test framework will clean up after you, so there is no harm in trying things that may leave lingering processes. However its policy is to fail if it finds any such process, since it can influence the results of other tests unintentionally. Hence you have to clean up inside the test.

kacfengine:
Fixed up and rebased in #449.

estenberg:
Oh, thanks! :)

PR #385, opened 2014-12-11
Add: Select files by symlink target

nickanderson:
This body allows selection of symlinks based on their target. This can be
useful if you want to make a promise about a symlink only if it points to a
specific location. For example deleting an old symlink.


nickanderson:
@estenberg @tzz thoughts?

tzz:
That's useful.  Normally I'd have to call `filestat` and create temporary classes.

estenberg:
Nice!

PR #393, opened 2015-01-02
Fix: Mission Portal Port Assignment from policy

nickanderson:
The policy to ensure the correct port is set if not 80 needed to be updated
after the appsettings.php file was updated.

Additionally this fixes a weak regular expression when setting the port in
httpd.conf

Ref: https://dev.cfengine.com/issues/6910

### DIFF
--- a/cfe_internal/CFE_hub_specific.cf
+++ b/cfe_internal/CFE_hub_specific.cf
@@ -249,7 +249,7 @@ bundle edit_line change_port(p)
 
     any::
 
-      "^Listen(\s+)(?!$(p)).*"
+      "^\s*Listen(\s+)(?!$(p)$).*$"
       comment => "Match Listen line and replace",
       handle => "cfe_internal_change_port_edit_line_listen_port",
       replace_with => value("Listen $(p)");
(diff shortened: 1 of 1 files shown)


nickanderson:
This should be picked into 3.6.x

cfe_internal/CFE_hub_specific.cf
```diff
@@ -249,7 +249,7 @@ bundle edit_line change_port(p)
 
     any::
 
-      "^Listen(\s+)(?!$(p)).*"
+      "^\s?Listen(\s+)(?!$(p)$).*"
```
kacfengine: Should it start with ``^\s*`` to be even more robust?

kacfengine: Also, doesn't the the final ``$`` contradict the ``.*`` that follows it? I suggest:

<pre>
^\s*Listen(\s+)(?!$(p))\s*$
</pre>

Which makes it even more robust by matching trailing whitespace as well. I think the regex is anchored, but good to have ``$`` at the end still.

nickanderson:
I'm by no means a regex guru. 

I was trying to account for the case of not matching the  promised content which leads to error messages.

kacfengine:
Ok, then I suggest you just switch it for the expression I proposed.

nickanderson:
I thought \s?  Was zero or more spaces.  Listen in the Apache config does not have to be lead by a space.

kacfengine:
No, '?' is zero or one (not more), '*' is zero or more.

cfe_internal/CFE_hub_specific.cf
nickanderson: That doesn't match. At least when I set the port to 8080.
This works though.
```
"^\s*Listen(\s+)(?!$(p)$).*$"
```

nickanderson:
I couldnt get your suggested regex to work. But I did update it a bit.

You suggested

```
^\s*Listen(\s+)(?!$(p))\s*$
```

But that didn't work in my testing.

This is what I updated it to and it seems to work well.

```
"^\s*Listen(\s+)(?!$(p)$).*$"
```

nickanderson:
ping @kacfengine See last note

PR #394, opened 2015-01-05
Fix paths to ping on FreeBSD 9.3 and later

rootwyrm:
FreeBSD relocated ping to /sbin/ping for 9.3, 10.x, 11.x. 

Will need to be addressed again in June when 8.4 goes end of life, reversing behavior so that `/sbin/ping` is the default and `/usr/bin/ping` is the exception for FreeBSD.

### DIFF
--- a/lib/3.6/paths.cf
+++ b/lib/3.6/paths.cf
@@ -207,7 +207,6 @@ bundle common paths
     freebsd_9_3|freebsd_10|freebsd_11::
   
       "path[ping]"     string => "/sbin/ping";
-      "path[sysctl]"   string => "/sbin/sysctl";
   
     freebsd|netbsd::
 
--- a/lib/3.6/paths.cf
+++ b/lib/3.6/paths.cf
@@ -194,13 +194,21 @@ bundle common paths
       "path[grep]"     string => "/usr/bin/grep";
       "path[ls]"       string => "/bin/ls";
       "path[netstat]"  string => "/usr/bin/netstat";
-      "path[ping]"     string => "/usr/bin/ping";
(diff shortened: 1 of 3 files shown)


lib/3.6/paths.cf
```diff
@@ -194,13 +194,21 @@ bundle common paths
...
       "path[tr]"       string => "/usr/bin/tr";
 
+    freebsd.!(freebsd_9_3|freebsd_10|freebsd_11)|netbsd|openbsd::
+  
+      "path[ping]"     string => "/usr/bin/ping";
+  
+    freebsd_9_3|freebsd_10|freebsd_11::
+  
+      "path[ping]"     string => "/sbin/ping";
+      "path[sysctl]"   string => "/sbin/sysctl";
```
rootwyrm: Accidental omission on 3.7, deliberate omission on 3.5. "path[sysctl]" is of course present, but I am planning to use it for future inventory work that depends on 3.6+ features. Corrected 3.7 so it's there. :)

nickanderson: The path for sysctl is /sbin/sysctl everywhere except archlinux and darwin. Unless I am missing something we shouldn't need a different path for those versions of freebsd.

rootwyrm: Ah-ha, I missed line 422 actually. Latest (2e67d52, acd2a19) have it removed. Will have to investigate NetBSD though, as a path of /usr/sbin/sysctl may still be valid on some systems.

nickanderson:
Looks clean would be great if you can squash the commits for easier cherry picking into 3.6.x.

`git rebase -i HEAD~6` on that branch

PR #395, opened 2015-01-07
Redmine #6865: Fix broken package promise as a result of inventory.

kacfengine:
We don't need to install the package, we only need to make sure that
the package list as been generated. However, we want to include the
package manager name, so that the list of packages is named correctly.
Hence the "<mgr> --help ; /bin/true".


kacfengine:
Don't merge yet. I need to test this with the packages it generates.

kacfengine:
trigger build

kacfengine:
Update with an acceptance test. We now test that inventory is generated properly.

kacfengine:
trigger build

PR #397, opened 2015-01-08
def.cf: use services/def.json to override some variables tagged "defvar"

tzz:
This lets you load any classes, any input files, and some variables from `services/def.json` (note that you only specify the things you want to override).

I tested as follows:

```console
% cf-agent -KI -f ./def.cf -b def -v|grep 'R: def: '                                                           *[master] 
R: def: override request domain to 'mydomain-com'; new value 'mydomain-com'
R: def: override request mailto to 'people@mydomain-com'; new value 'people@mydomain-com'
R: def: override request mailfrom to 'me@mydomain-com'; new value 'me@mydomain-com'
R: def: override request smtpserver to 'smtp.mydomain-com'; new value 'smtp.mydomain-com'
R: def: override class services_autorun
R: def: default:def.mailto = people@mydomain-com
R: def: default:def.mailfrom = me@mydomain-com
R: def: default:def.domain = mydomain-com
R: def: default:def.smtpserver = smtp.mydomain-com
```


tzz:
I added the ability to specify inputs in `services/def.json` as well.  Which is useful.

basvandervlies:
Nice one

nickanderson:
Initial conversations are leaning to using this implementation but not until 3.7 unless we get overwhelming user support to have it in 3.6.

We may use this implementation as a backup/alternative https://gist.github.com/nickanderson/6f52ba76566bedddb235
in the event people need classed based settings.

tzz:
I added the ability to specify roles dynamically in two ways:

* by class regex, with a list of roles associated with each regex
* by roles, with a list of class regexes associated with each role

In my experience, users are 50/50 between those two ways to classify hosts, so I think it makes sense to allow both.

tzz:
(Note that the need to classify hosts is consistently the first things users ask for when they learn CFEngine.)

tzz:
I added some more things:

* `update.cf` can load values from `services/def.json` as well now, see how `input_name_patterns` is extended to add `*.js` files
* only data containers are used in `def.cf` and `update.cf` for the overrides, supporting...
* strings or slists can be overridden now, and potentially more!
* finally, reporting improvements.  Test with:

```console
% cf-agent -Kv -f ./def.cf -b def|grep 'R: '
% cf-agent -Kv -f ./update.cf -b update_def|grep 'R:'
```

The reports can show empty variables sometimes, and for slists you get 1 line per item, but I didn't bother with perfecting them because they are for debugging only.

tzz:
(mentioned in the mailing list as well)

I was thinking how convenient it would be if the C code supported loading variables and classes from a JSON file automatically. It's related to persistence, but you're only loading the data, not persisting it.  I don't know if the two could be combined.

neilhwatson:
I like this idea. BTW, while JSON does not permit comment syntax, you can make a comment field that the rest of the policy will ignore e.g.

    "comment" : "This defines the file types that will be updated in inputs",

I encourage this use so that the JSON does not become too inscrutable.

tzz:
The CFEngine JSON parser should be modified to allow comments, if we want those.  It's pretty common to allow `//` syntax (we allow it in Design Center JSON, for instance).

Putting comments inside data is IMO the worse approach.  I see `def.json` as a pure augmentation of the policy (which you can find in `def.cf` and `update.cf`), and something that should be updated by tools like the Mission Portal rather that manual editing.

neilhwatson:
> The CFEngine JSON parser should be modified to allow comments,

I don't like the ideas of supporting a non compliant JSON syntax, but leave that to others with more CS theory than I.

tzz:
@neilhwatson I kind of like the https://github.com/timjansen/hanson syntax.  But I really don't think it's all that important for this PR :)

nickanderson:
But if you put a "reason" field in the data, it can be tied together with the policy into a "comment" in policy. Same goes for stakeholders. 

I find a lot of value in those features of cfengine. IIRC Neil has been pretty consistent about keeping that knowledge very close to his data in EFL. Looking now I don't see comment/reason, but I do see promisee in the example data.

tzz:
Well, you're trying to turn data into policy... who's reading the data?  Shouldn't they be reading the policy?  It feels like extra baggage to me.  But I know what you mean, it's nice to have some connection with this feature.

Now, YAML supports comments!  Should we just switch to `def.yaml` or support both `def.json` and `def.yaml`?  Would that be enough, or do you think the glue has to be stronger?

nickanderson:
But there is a slight difference from having comment support and being able to associate a comment with a particular piece of data.

There is a reason that that particular data is defined as so. I am not sure we need to add yaml support for this feature. Though personally I think yaml is slightly easier to read.

tzz:
added `services/extra_classes.txt` where you simply add classes to be defined, one class per line.

The `masterfiles_perm_mode` override had issues on bootstrap so I disabled it.

update.cf
```diff
@@ -43,19 +43,46 @@ body agent control
...
+      "$(override_classes)" expression => "any";
 
+  vars:
       "current_version" string => "3.7.0";
+    any::
+      "defvars" slist => variablesmatching("default:def\..*", "defvar");
+
+      "extra_classes" slist => readstringlist("$(this.promise_dirname)/services/extra_classes.txt", "#.*", "\s", 10k, 100k);
+
+      "augments" data => readjson("$(this.promise_dirname)/services/def.json", 100k);
```
nickanderson: I think this should be guarded to only when the file exists.

services/def.json
```diff
@@ -0,0 +1,32 @@
+{
```
nickanderson: I don't think that this file should be part of masterfiles itself. But maybe exist as example_def.json OOTB for people to get inspiration from. ?

services/def.json
```diff
@@ -0,0 +1,32 @@
...
+    "roles":
```
nickanderson: Why "roles" instead of "classes" or "extra_classes"? I didnt see any special role_ prefix or meta tags attached to the classes defined from there. Maybe it would make more sense if the classes were prefixed with role_ and tagged?

services/extra_classes.txt
```diff
@@ -0,0 +1 @@
+services_autorun
```
nickanderson: Why did you choose to put this in a different file instead of inside of def.json?

update.cf
```diff
@@ -43,19 +43,46 @@ body agent control
...
+
+      # all override classes are defined true
+      "$(override_classes)" expression => "any";
 
+  vars:
       "current_version" string => "3.7.0";
+    any::
+      "defvars" slist => variablesmatching("default:def\..*", "defvar");
+
+      "extra_classes" slist => readstringlist("$(this.promise_dirname)/services/extra_classes.txt", "#.*", "\s", 10k, 100k);
```
nickanderson: Probably should be guarded so that its only read when the file is present.

services/def.json
nickanderson: OR put it in a namespace.
So i could use it like this
```
roles:webserver.rhel_6:: 
```

def.cf
```diff
@@ -4,30 +4,73 @@
...
+      "$(roles_byrole_keys)" expression => classmatch("$(extra_roles[byrole][$(roles_byrole_keys)])");
 
   vars:
 
     any::
+      "defvars" slist => variablesmatching("default:def\..*", "defvar");
+
+      "extra_classes" slist => readstringlist("$(this.promise_dirname)/services/extra_classes.txt", "#.*", "\s", 10k, 100k);
+
+      "augments" data => readjson("$(this.promise_dirname)/services/def.json", 100k);
```
nickanderson: It might be nice if def.json were relocatable, also I think it should only be read if it exists.

update.cf
tzz: OK; done

services/def.json
tzz: OK; done

services/def.json
tzz: I tried the namespace and it was not working right.  I left it as the user expects and added a `role` meta tag.

services/extra_classes.txt
tzz: Because it's really convenient to say `echo services_autorun >> extra_classes.txt` :)

I commented it out but left the file in place, so it's empty and ready to use.

update.cf
tzz: OK; done

def.cf
tzz: OK; done.  With command-line arguments I could do better :)

tzz:
Take a look, it's as you suggested except for the namespace.  **I TRIED OH GOD I TRIED**.

tzz:
What's missing or needs to be modified to make this OK for merging?

nickanderson:
This is waiting until we shift focus to 3.7. Other than being re-based it looked pretty good to me the last time I read through it.

kacfengine:
This makes the file a little intimidating for those beginners who are just going to edit a few defaults. Is there any way that the stuff between "Begin change" and "End change" could be kept simple and at the top?

tzz:
I'm OK with moving things around, sure.  The goal is to make it easy to edit.  So maybe a YAML version of `def.json` would be better?

kacfengine:
I think it's nice to keep the default editing in the ``def.cf`` file, ``def.json`` is more for special needs. Also, this file is not version protected, so it would be nice to keep it compatible with at least 3.6, which means ``readyaml`` will have to be outsourced to a different file (yes, I can hear "future syntax" echoing in the background :-) ). If you do that, you might as well do ``readjson`` as well, and it'll be compatible with 3.5 as well for extra points!

And I think the default editing of things like ``mailfrom`` should be easy, so don't wrap those in an ``ifelse`` or anything like that. However, you can change the name, for example ``mailfrom`` -> ``default_mailfrom`` and then an ``ifelse`` somewhere else can take care of the logic to make sure ``mailfrom`` is populated. The key thing is just to keep part that the user wants to edit clean, so that we don't mix in a bunch of logic unrelated to their task and which uses syntax they may not have learned yet.

nickanderson:
Well, the idea will be that going forward users will only make the modifications in def.json. The file will *likely* be version controlled at the customer site, its just not a file that we ship so upgrades to the policy framework should be easier, because hopefully *most* user modifications will be contained to def.json and since we don't ship it there should be no conflicts.

kacfengine:
Ok, so the user is not expected to edit ``def.cf`` at all anymore? Then it makes sense, but I suppose we should at the very least ship an example file of ``def.json`` so new users can easily get going, or?

nickanderson:
With the addition of def.json or def.yaml there is no need for the user to edit the policy directly for the most common changes. We can include a def.json.example and or def.yaml.example but the whole point is that we should not ship the file so that there is no conflict or diffing to do for that file during framework upgrades.

tzz:
I gave an `example_def.json` that can be edited to reflect what's currently in `def.cf`.  Currently they are not the same defaults.  As far as complexity, I tried to keep it simple but the bigger responsibility to the user is to make it resilient, I think.  @nickanderson has a good feel for what users consider too complex, so maybe he has ideas on simplifying this.

Ideally the JSON would come out of MP in Enterprise, not manual editing.  But that's a different task...

kacfengine:
> I gave an ``example_def.json``

Right, I missed that somehow.

So it looks pretty good, although I haven't reviewed the actual code in detail. If @nickanderson is ok with this, then I am ok with this, so merge at will.

tzz:
@nickanderson are you good with the scope, naming, and behavior of this?  I'm happy to adjust or expand it further.

update.cf
```diff
@@ -43,19 +43,61 @@ body agent control
...
+      "augments_are_yaml" expression => regcmp(".+\.yaml", $(augments_file)), scope => "bundle";
+      "have_yaml_augments" and => { "have_augments_file", "augments_are_yaml" }, scope => "bundle";
+
+  vars:
       "current_version" string => "3.7.0";
+    any::
+      "augments_file" string => ifelse(fileexists("$(this.promise_dirname)/services/def.yaml"), "$(this.promise_dirname)/services/def.yaml",
+                                      "$(this.promise_dirname)/services/def.json");
+
+      "defvars" slist => variablesmatching("default:def\..*", "defvar");
```
nickanderson: Should this one be update_def\.* ?

tzz: I think so.  Can you test with that change?

def.cf
```diff
@@ -4,30 +4,95 @@
...
+      "augments_file" string => ifelse(fileexists("$(this.promise_dirname)/services/def.yaml"), "$(this.promise_dirname)/services/def.yaml",
+                                      "$(this.promise_dirname)/services/def.json");
+
+      "defvars" slist => variablesmatching("default:def\..*", "defvar");
+
+      "extra_classes" slist => readstringlist("$(this.promise_dirname)/services/extra_classes.txt", "#.*", "\s", 10k, 100k);
+
+    have_augments_file::
+      "augments" data => readjson($(augments_file), 100k), ifvarclass => "augments_are_json";
+      "augments" data => readyaml($(augments_file), 100k), ifvarclass => "augments_are_yaml";
```
nickanderson: We cant have yamlstuff until 3.8 since it breaks compatibility with 3.6. I think it's fine to have it there but commented out.

update.cf
```diff
@@ -43,19 +43,61 @@ body agent control
...
+
+      "extra_classes" slist => readstringlist("$(this.promise_dirname)/services/extra_classes.txt", "#.*", "\s", 10k, 100k);
+
+    have_augments_file::
+      "augments" data => readjson($(augments_file), 100k), ifvarclass => "augments_are_json";
+      "augments" data => readyaml($(augments_file), 100k), ifvarclass => "augments_are_yaml";
```
nickanderson: We cant have yamlstuff until 3.8 since it breaks compatibility with 3.6. I think it's fine to have it there but commented out.

def.cf
```diff
@@ -4,30 +4,95 @@
...
+      "have_inputs" expression => isvariable("augments[inputs]"), scope => "bundle";
+
+    have_roles::
+      "$(extra_roles[byname][$(roles_byname_keys)])" expression => classmatch("$(roles_byname_keys)"), meta => { "role" };
+      "$(roles_byrole_keys)" expression => classmatch("$(extra_roles[byrole][$(roles_byrole_keys)])"), meta => { "role" };
 
   vars:
 
     any::
+      "augments_file" string => ifelse(fileexists("$(this.promise_dirname)/services/def.yaml"), "$(this.promise_dirname)/services/def.yaml",
```
nickanderson: I think its a bit odd to have these files inside of the services directory, instead of sitting next to def.cf. Do you have a specific reason for it?

update.cf
nickanderson: Yep, it needs to be "default:update_def\..*" instead of default:def\..*. I dont get the masterfiles_perms_mode printed out unless its update_def.
 
R: update_def: default:update_def.masterfiles_perms_mode = 0600

def.cf
tzz: I thought it was tidier, because the other things in `services` are disposable, but I've move it now.

nickanderson:
@estenberg @tzz readjson isn't available in 3.5, in order to preserve backwards compatibility for 2 minor versions, we need to split def.cf into lib/VER.

https://github.com/tzz/masterfiles/pull/2

nickanderson:
Ah, I just realized the same is also true for update.cf since it also has the override support. But in this case update.cf is the entry, so I guess we should just break update_def out into its own file, and move that into lib/VER OR since update.cf has long been *NOT* standalone, we could have update.cf load def.cf and consolidate all of that stuff together. There is extra risk any time someone edits def.cf, but that shouldn't happen very often. 

Thoughts?

tzz:
Breaking `def.json` is not as tragic as breaking `def.cf`, that's one of the nice features of the new approach.  We can abort more gently.

nickanderson:
Right, i was just salivating at removing some of the duplication between update_def and def :-P but there is some reason to keep them separate. And its really only a pain for those who maintain them, since users will *hopefully* start using def.json.

nickanderson:
Split update_def into lib/VER https://github.com/tzz/masterfiles/pull/3/files

nickanderson:
can you rebase this?

tzz:
I've merged the master branch into it, which makes it mergeable...

estenberg:
Quite complicated structure/names.
Propose simplifying the class definition part with simply

 "classes":
 {
   "services_autorun": [ "any" ],
   "my_apache" : [ "server1", "server2", "redhat.*" ]
   "my_other_apache" : [ "server[34]", "debian.*" ]
 },

This aligns with most, if not all, programming languages where you define something on the left based on something on the right.

tzz:
Unfortunately half the people like `byrole` (what you prefer, list the roles and for each one define what's in it) and the other half like `byname` (given a host name, define its roles), based on my experience.  Furthermore **both** groups believe their way is best!

I think it's not a big deal to support both and hope you'll accept that.  But instead of `classes -> byname` and `classes -> byrole` maybe they would be easier to understand as `host_classes` and `roles` at the top level?

estenberg:
In order to make CFEngine simple and easy to learn we should not have both. A few reasons why I believe that the original "byrole" way is the best

* It's how all common programming languages do it (define on the left, value to the right), hence it is intuitive to anyone with some programming experience
* "Special" hosts are going away with cloud, containers, and so on, so it doesn't make sense to specify something for one particular host anymore (pet versus cattle and so on)
* The "byrole" approach allows you to define several classes on one host by having several lines, if you really want to

tzz:
OK, my latest commit does the following:

* only support JSON (anticipating the acceptance of https://github.com/cfengine/core/pull/2162 to give us the `readdata()` function, which we'll then use in the 3.7 branch)
* drop `roles/byname` and only have a single top-level `roles` key, simplifying the logic as @estenberg suggested.
* use `dirname(def.cf)/def.json` first, then `dirname(def.cf)/../../def.json`.  The effect is that you can have `lib/3.7/def.json` and also `def.json` at the root level.  We could also **merge** the two files?
* add `.yaml` to the default list of inputs everywhere in 3.6. and 3.7

tzz:
ping?

nickanderson:
Pong. I havent forgotten, hope to get back to revewing this week or next. I think its pretty close. @estenberg @kacfengine Have you had a chance to review readdata()?

nickanderson:
- add .yaml to the default list of inputs everywhere in 3.6. and 3.7
  - Why add to 3.6? It cant read yaml
- multiple def.json files
  - I'm not sure we need multiple places, we already have defaults set in policy, have other defaults defined in lib which can be overridden with the def.json defined at the root level seems to add little value. Did you have something specific in mind?

nickanderson:
Finally

PR #400, opened 2015-01-09
Change: Reorganize cfe_internal

nickanderson:
Not all policy that ships with CFEngine is applicable or even useful to
community users. This work is to seperate the enterprise specific policy from
the core policy and only load the enterprise specific policy when necessary.

While these changes are compatible with 3.6 but will likely be kept in master
until 3.7 due to it being a large change.


kacfengine:
I haven't looked at this in detail yet, but I like the idea of splitting up the enterprise vs community parts.

> While these changes are compatible with 3.6 but will likely be kept in master until 3.7 due to it being a large change.

And since it reorganizes a lot of code too, I vote for waiting until focus shifts to 3.7 in order to make cherry-picking between branches easier.

kacfengine:
Rebased in #448.

PR #401, opened 2015-01-15
Add: edit_line body to insert line before regular expression

nickanderson:
Sometimes one thing must be found in a file before another.

I think this should be picked to 3.6 as well.


kacfengine:

PR #402, opened 2015-01-16
Improve inline documentation

danielhoherd:
Updating a self referential "see also" section to point to something useful.


nickanderson:
Thanks!
I pulled your change in via https://github.com/cfengine/masterfiles/pull/403 to add support to the 3.7 tree of the library.

PR #405, opened 2015-01-20
factor out current version to variable to make upgrading less error-prone

estenberg:

PR #406, opened 2015-01-20
factor out current version to variable to make upgrading less error-prone (3.6.x)

estenberg:


estenberg:
@nickanderson review please?

estenberg:
Please merge this and #405 at the same time.

nickanderson:
Shouldn't we set current_package_build or something too in order to handle The .0 and -1 packages.

estenberg:
I don't think we change this very often (or shouldn't at least). But perhaps someone else has input?

nickanderson:
We do not change it very often. Only if we have a packaging problem, which we have had in the past. 
We can skip it for now since its generally not used.

PR #408, opened 2015-01-21
Add package method for SmartOS pkg_add

nickanderson:
This package method interacts with SmartOS pkg_add to install from local or
remote repositories. It is slightly different than the FreeBSD `pkg_add`.

 This example installs `perl5` from a remote repository.

     packages:
     "perl5"
     package_policy   =>  "add",
     package_method   =>  smartos_pkg_add("http://pkg.example.com/packages/");

 This example installs `perl5` from a local package cache.

     packages:
     "perl5"
     package_policy   =>  "add",
     package_method   =>  smartos_pkg_add("/opt/custom/packages");

(cherry picked from commit 076bee22e1808e693021d6d1c5b8ffc19c567b63)

Conflicts:
lib/3.7/packages.cf

### DIFF
--- a/lib/3.6/packages.cf
+++ b/lib/3.6/packages.cf
@@ -1247,6 +1247,36 @@ body package_method smartos
       package_update_command =>  "/opt/local/bin/pkgin -y install";
 }
 
+body package_method smartos_pkg_add(repo)
+# @brief SmartOS pkg_add installation package method
+#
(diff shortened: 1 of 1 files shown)


estenberg:
New feature so the default should be not to put it in 3.6.x unless there are specific reasons.

PR #409, opened 2015-01-21
Change: Remove def.domain from default acl

nickanderson:
Accepting def.domain might be insecure and imply performance penalty in form
of DNS lookup. Redmine #5556

(cherry picked from commit d902e8b88b9bb5588814e8b9600126fb0d557c90)

### DIFF
--- a/def.cf
+++ b/def.cf
@@ -34,7 +34,7 @@ bundle common def
                            # Note that this:
                            # 1. requires def.domain to be correctly set
                            # 2. will cause a DNS lookup for every access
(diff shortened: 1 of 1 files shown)


jimis:
I was originally pushing for this, so +1 from me.

estenberg:
No, definitively too risky for 3.6.x. What if someone is relying on def.domain to authenticate and now suddenly no clients can talk to the policy server after upgrading (masterfiles) to 3.6.4?

PR #410, opened 2015-01-23
Enterprsie reports are accumulated on the client till they are collected...

maciejmrowiec:
... by cf-hub.

In case if not collecting the reports for extended period of time, they may consume too much space,
this policy will purge accumulated "diff" reports if they exceed 50MB in total size. As the result,
report history will become inconcistent with next successful report collection and cf-hub
will go for rebase query (fetching full state next time). This cleanup apply only to diff reports
which include class, variable, promise execution, software installed, software patch, lastseen reports.

Redmine: #1150

PR #411, opened 2015-01-23
Enterprsie reports are accumulated on the client till they are collected by

nickanderson:
cf-hub.

In case if not collecting the reports for extended period of time, they may
consume too much space, this policy will purge accumulated "diff" reports if
the exceed 50MB in total size. As the result, report history will become
inconcistent with next successful report collection and cf-hub will go for
rebase query (fetching full state next time). This cleanup apply only to diff
reports which include class, variable, promise execution, software installed,
software patch, lastseen reports.

Redmine: #1150

### DIFF
--- a/def.cf
+++ b/def.cf
@@ -150,6 +150,11 @@ bundle common def
         "$(sys.workdir)/httpd/logs/error_log", # Mission Portal
       };
 
+      "max_client_history_size" -> { "cf-hub", "CFEngine Enterprise" }
+        int => "50M",
+        comment => "The threshold of report diffs which will trigger purging of
+                    diff files.";
+
     enterprise.!am_policy_hub::
       # CFEngine's own log files
       "cfe_log_files" slist => { @(base_log_files), @(enterprise_log_files) };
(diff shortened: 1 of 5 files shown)


nickanderson:
Supersedes https://github.com/cfengine/masterfiles/pull/410

estenberg:
Nice.

def.cf
```diff
@@ -150,6 +150,11 @@ bundle common def
         "$(sys.workdir)/httpd/logs/error_log", # Mission Portal
       };
 
+      "max_client_history_size" -> { "cf-hub", "CFEngine Enterprise" }
+        int => "50M",
```
estenberg: Does this work (50M = 50 MB?) ? Just never seen it before

nickanderson: According to [the language concepts documentation on variables](https://docs.cfengine.com/latest/guide-language-concepts-variables.html)
suffixing an integer value with 'M' will multiply the value by 1024.

#+beign_quote
  Since computing systems such as storage and memory are based on
  binary values, CFEngine also provide the following uppercase
  suffixes to create integer values for common powers of 1024.

    'K' = value times 1024.
    'M' = value times 10242
    'G' = value times 10243
#+end_quote

PR #412, opened 2015-01-23
Enterprsie reports are accumulated on the client till they are collected by

nickanderson:
cf-hub.

In case if not collecting the reports for extended period of time, they may
consume too much space, this policy will purge accumulated "diff" reports if
the exceed 50MB in total size. As the result, report history will become
inconcistent with next successful report collection and cf-hub will go for
rebase query (fetching full state next time). This cleanup apply only to diff
reports which include class, variable, promise execution, software installed,
software patch, lastseen reports.

Redmine: #1150
(cherry picked from commit 65b96a5a28e478a2e68f4d4bbdd899a811563ca1)

Conflicts:
lib/3.7/cfe_internal.cf


nickanderson:
Please merge with https://github.com/cfengine/masterfiles/pull/411

nickanderson:
@estenberg ping

Issue #413, opened 2015-01-25
Reading assumed real value 'cf_null', anomalous remainder '' 


leoliu:
Running the `3.6.x` branch with cfengine 3.6.3 on centos 7, I am seeing many such errors lately:

2015-01-25T10:03:51+0800    error: /default/cfe_internal_cleanup_agent_reports/vars/'total_report_size'[0]: Reading assumed real value 'cf_null', anomalous remainder ''

nickanderson:
Thanks for the report @leoliu.

In the future please submit issues to redmine at dev.cfengine.com.

Thanks!

leoliu:
Hi @nickanderson,

I can do that but unfortunately all these bug trackers require registration and give me a headache :(

Leo

PR #414, opened 2015-01-26
Change: Avoid generating errors in logs

nickanderson:
Diff state purging is only applicable when diff state files actually exist.
This adds a guard to ensure we only try to caclulate the size of the diff
state reports when there is something to calculate.

Ref: https://github.com/cfengine/masterfiles/issues/413


tzz:
Thanks, this was annoying.  I was hoping to see `eval` used :)

PR #417, opened 2015-01-28
Change: Remove redundant package promise

nickanderson:
There is no need to have two package promises to upgrade Enterprise. I chose
to remove the package promises guarded by stopped_cfprocs because in order for
those promises to be activated the software must be downloaded from the hub,
and cfengine processes must be stopped within the same agent run. This creates
a brittle chain.

Ref: https://dev.cfengine.com/issues/6964

### DIFF
--- a/update/update_bins.cf
+++ b/update/update_bins.cf
@@ -145,66 +145,7 @@ bundle agent cfe_internal_update_bins
 
       #
 
-  processes:
-
-    bin_newpkg.!bin_update_success.enterprise::
-
-      "$(cf_components)" signals => { "$(stop_signal)" },
-      comment => "Stop cfengine running processes before binary update",
-      handle => "cfe_internal_update_bins_processes_stop_cfengine",
-      classes => u_if_repaired("stopped_cfprocs");
-
-    bin_newpkg.!bin_update_success.!windows.enterprise::
-
-      "cf-execd"  signals => { "$(stop_signal)" },
-      comment => "Stop cf-execd running process before binary update",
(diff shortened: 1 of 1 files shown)


kacfengine:
Are there any ill effects of leaving this as is, other than slightly better performance when upgrading?

I'm not saying I support the current way, but I'd really like to have the install tests in place before we attempt to mess too much with the automatic upgrades. They are notoriously time consuming to test manually everywhere.

nickanderson:
This is causing the package upgrade to be attempted twice. The second attempt fails resulting in a restoration from backup which in turn leads to inodes changing for files in inputs, and if you are copying files from inputs using ctime (default copy_from behaviour), services that use said files could be unnecessarily and unexpectedly restarted causing service disruption.

kacfengine:
@estenberg: What do you think about this one? I think the reasons are good, but it's quite risky. I'm thinking we should wait until we have install tests.

estenberg:
It is a bug and fits well into 3.6.x scope.

If the install tests will help make testing more efficient it can wait, but won't we need upgrade tests for this?

kacfengine:
> If the install tests will help make testing more efficient it can wait, but won't we need upgrade tests for this?

We do. They are planned together with install tests, but they will definitely be "second iteration", so it might be a little while.

I will merge this then, and we'll have to put some focus on upgrade testing the platforms well for 3.6.5.

kacfengine:
I do always like removal of code though! :-)

nickanderson:
@kacfengine what about https://github.com/cfengine/masterfiles/pull/418 for 3.6.x?

PR #418, opened 2015-01-28
Change: Remove redundant package promise

nickanderson:
There is no need to have two package promises to upgrade Enterprise. I chose
to remove the package promises guarded by stopped_cfprocs because in order for
those promises to be activated the software must be downloaded from the hub,
and cfengine processes must be stopped within the same agent run. This creates
a brittle chain.

Ref: https://dev.cfengine.com/issues/6964
(cherry picked from commit 403af5633e5b61ac7d087a52c84f380f1341cbac)

### DIFF
--- a/update/update_bins.cf
+++ b/update/update_bins.cf
@@ -145,66 +145,7 @@ bundle agent cfe_internal_update_bins
 
       #
 
-  processes:
-
-    bin_newpkg.!bin_update_success.enterprise::
-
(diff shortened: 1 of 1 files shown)


jimis:
I don't have knowledge in this area but anything that removes code from the bloated update policy gets a +1 from me!

kacfengine:
Ah sorry, this was cherry-picked manually already!

PR #419, opened 2015-01-28
package inventory policy ran on every agent run

estenberg:
Tested on CentOS 6.
yum is run shortly after bootstrap and package installed & update inventory works.
yum is not run after that (yet) - has still not tested for 24h.

Please backport to 3.6.x if accepted.


estenberg:
@nickanderson commentus please.

nickanderson:
Yes, great!

PR #421, opened 2015-02-08
More fixes to paths.cf for FreeBSD

rootwyrm:
This is a fairly big one. 

Biggest change here is what I'm calling the Major Version Concatenator; catchier names welcome. CFEngine doesn't create a hardclass for the major, only major_minor. Fixing that would be very intrusive to core for little benefit, and it is needed for cleanup plus pkgng. paths.cf seemed the best place to put it, as it will also prevent future cluttering in paths.cf but I'm open to better suggestions as to methods and/or placements. This just seemed the least intrusive solution.

Other notable fixes addressed here:
* Attempt to address perl <5.20 being at /usr/bin/perl and >=5.20 NOT having /usr/bin/perl with an `slist`. Tested okay here, but more eyes are better.
* Add `setfacl` (POSIX.1e), `getfmac` (Mandatory Access Control), `setfmac` - MAC paths are handy as core basically has zero MAC handling.
* Taking another swing at `path[ping]` since transition seems to be somewhere in 9.2 patchlevels _and_ 9.2->9.3.
* Added `path[pkg]` for ongoing pkgng work (pull reqs for that coming Soon<sup>TM</sup>) which also covers DragonflyBSD.


kacfengine:
Thanks, some issues before this can be merged:

* Please squash the last two commits, the second to last commit is pretty useless by itself. Same goes if you make more commits to the ``platform_versions.cf`` file.
* The ``path[perl]`` slists still need to use strings.
* The classes you added to ``platform_versions.cf`` should be removed from elsewhere.
* ``platform_versions.cf`` should be added to the ``stdlib.cf`` file in the same directory, so that it is auto-loaded.

lib/3.6/platform_versions.cf
```diff
@@ -0,0 +1,99 @@
...
+      perl_5_18::
+        "perl_version[path]" string => "/usr/local/bin/perl5.18";
+      perl_5_20::
+        "perl_version[path]" string => "/usr/local/bin/perl5.20";
+        
+    classes:
+      "perl_5_16" expression => fileexists("/usr/local/bin/perl5.16");
+      "perl_5_18" expression => fileexists("/usr/local/bin/perl5.18");
+      # /usr/bin/perl is guaranteed ABSENT after this point...
+      "perl_5_20" expression => fileexists("/usr/local/bin/perl5.20");
```
tzz: I think it's better to `findfiles("/usr/local/bin/perl5\.[0-9]+")` and then canonify the basename of the last found executable.  That way you don't have to list all the possible Perl versions.

lib/3.6/platform_versions.cf
```diff
@@ -0,0 +1,99 @@
...
+# major releases. Developers: this is updated (removal of oldest,
+# addition of newest) when -CURRENT changes major numbers.
+{
+  vars:
+    freebsd::
+      "freebsd_7_mvc" slist => classesmatching("freebsd_7*");
+      "freebsd_8_mvc" slist => classesmatching("freebsd_8*");
+      "freebsd_9_mvc" slist => classesmatching("freebsd_9*");
+      "freebsd_10_mvc" slist => classesmatching("freebsd_10*");
+      "freebsd_11_mvc" slist => classesmatching("freebsd_11*"); # -CURRENT
```
tzz: You can just say:

```
classes:
  "freebsd_7_mvc" expression => classmatch("freebsd_7.*");
```

Note the regex I wrote above; `freebsd_7*` as you have it will match `freebsd_`, `freebsd_7`, `freebsd_77`, etc.

That being said, I think this belongs more in `inventory/freebsd.cf` than in the stdlib.

lib/3.6/platform_versions.cf
rootwyrm: As I mentioned on IRC, I pretty much agree with you on that. However, canonifying in that fashion isn't an option because that puts it in opposition to intended behavior/capability of having multiple perl5s which do not conflict. Optimally it should examine `/etc/make.conf` and `/usr/local/etc/perl5_version` but then you get `DEFAULT_VERSIONS = perl5=5.18`, `PERL5_DEFAULT=5.18`, and `PERL_VERSION=5.18.4` as an example.

If there's no objections, I think the best option may instead be to accept potential breakage for 8.x systems (EOSL in June anyway) and rely on `/usr/local/etc/perl5_version`. Implement it roughly as `path[perl] string => "/usr/local/bin/perl$(perl_version)";` which should be fine for 5.16+.

rootwyrm:
@kacfengine I've actually been working with my local copy (and forgetting to commit, insert sheepish look here) trying to get the AIX soft-classes in as well to extend `aix_6.1` to `aix_6.1_6` (TL6) for example; but without having to rely on (extremely) expensive `lslpp` calls to ensure it keys off committed and not installed. Long story short, now also trying to figure out why there are 6+ instances of `bos` being picked up.
If preferred, I can go ahead and clean up FreeBSD MVC + perl_headache<sup>TM</sup>, bring in `vios_platform_detect` and `vios_platform_version` - which should work correctly from padmin now - and squash it down this week.

lib/3.6/platform_versions.cf
tzz: Yeah, I agree it's best to assume 9.x or higher and do it with `perl5_version`.

tzz:
@kacfengine I agree about running commands unnecessarily from a candidate slist.  `findfiles(option1, option2, option3, ...)` works best.  I suggested adding options to `findfiles` to find executable files, for instance (like `/usr/bin/find` options).  Here it would be very useful.

rootwyrm:
Sorry about the delay! Should *finally* be squashed with VIOS added (vastly simplified from what I had last week). 

Additional testing is **STRONGLY** recommended for `aix_vios_level` as I can only check one older version managed by IVM. It *may* require a longer read for later versions. CPU effect was negligible.

Switched to the more generic `perl_version` (since this may apply to additional OSes) and abused regular expressions to extract the expected version and stuff it into the path. `paths.cf` then looks to `$(perl_version.bsd_path_perl)` to avoid the `slist`.

kacfengine:
I see that my bullet point "The ``path[perl]`` slists still need to use strings." is still a problem (I haven't checked the other points). This will not be merged until that is fixed.

Also, you'll have to rebase the PR, since it has conflicts now.

rootwyrm:
@kacfengine Yeah, I fubar'd the squash there. (I do have a fixed version.) This has turned into a bit of a rat's nest WRT commits, so I'm gonna close this pull and open fresh ones this weekend which can merge clean. `platform_versions.cf` should likely be off on it's own anyway since it's bulked up substantially with AIX and VIOS fixes and it's a new file.

PR #422, opened 2015-02-10
turn off cleanup of several cf-execd and cf-monitord by default

estenberg:
May kill CFEngine processes inside containers.
Redmine: https://dev.cfengine.com/issues/6906.

This is a workaround, ideal solution is to have a way to filter process in the current namespace (zone/cgroup/etc.).

Consider backport to 3.6.x.


kacfengine:
Quite hacky fix, but I can't think of a better one ATM that isn't nontrivial.

kacfengine:

PR #423, opened 2015-02-11
Add: Purging of old scheduled reports for Mission Portal

nickanderson:
Scheduled reports need to be cleaned up after some time so that they do not
fill up the disk on CFEngine Enterprise Mission Portal hubs.

Ref: https://dev.cfengine.com/issues/6819


cfe_internal/CFE_cfengine.cf
```diff
@@ -31,6 +31,14 @@ bundle agent cfe_internal_management
 
     am_policy_hub.enterprise::
 
+      "hub" -> { "Mission Portal", "CFEngine Enterprise" }
+        usebundle => cfe_internal_purge_scheduled_reports_older_than_days( $(def.purge_scheduled_reports_older_than_days) ),
+        ifvarclass => "enable_cfe_internal_purge_scheduled_reports_older_than_days",
```
estenberg: Don't think ifvarclass is necessary, assume almost everyone would want this (and they can adjust the threshold in days).

def.cf
```diff
@@ -253,6 +258,11 @@ bundle common def
         comment => "If reports are not collected for an extended period of time
                     the disk may fill up or cause additional collection
                     issues.";
+
+      "enable_cfe_internal_purge_scheduled_reports_older_than_days" -> { "Mission Portal", "CFEngine Enterprise" }
```
estenberg: Likewise, would not define this class, just make it always run and let people customize with threshold def.

estenberg:
Very nice!

cfe_internal/CFE_cfengine.cf
nickanderson: My reasoning here is that this makes it easier to roll the change across a subset of hosts. Imagine 10k virtual machines sharing the same storage back end all deciding to delete a bunch of old files within the same 5 minutes. Without this enable class, I would have to dig deep into the policy in order to instrument a phased roll out.

nickanderson: Granted, this is probably not a good example since you probably wouldnt have 10k hubs. But its a general pattern that I was trying to establish. Things ought to be specifically enabled.

estenberg: Yes, I know what you mean for phased roll-out across nodes.

However in this case I would just remove it as it won't be a problem for hubs and configurability has a tradeoff with complexity.

nickanderson:
Removed class guard @estenberg

estenberg:
Looks great! Thanks!

estenberg:
Cherry-picked 3.6.x: 8b5188b6cf9bfd70c359b28af292b796cd116e11

PR #425, opened 2015-02-19
Add path to dtrace, zfs and zpool on FreeBSD

skreuzer:
Remove trailing whitespace


nickanderson:
This looks good to me, and should be picked into 3.6.x

kacfengine:
Absolutely! Merged to both master and 3.6.x.

PR #426, opened 2015-02-23
Redmine #6401, #6402: Remove '/' member in test packages.

kacfengine:
On RHEL7 this entry causes a conflict with a filesystem package.


kacfengine:
trigger build

PR #427, opened 2015-02-25
Fix: rm_rf_depth should also remove the named directory

nickanderson:
Previously it was only removing the contents of the directory.

Ref: https://dev.cfengine.com/issues/7009

### DIFF
--- a/lib/3.5/bundles.cf
+++ b/lib/3.5/bundles.cf
@@ -131,6 +131,9 @@ bundle agent rm_rf_depth(name,depth)
       depth_search => recurse_with_base($(depth)),
       delete => tidy;
 
+      "$(name)/."
+        delete => tidy;
+
     !isdir::
       "$(name)" delete => tidy;
 }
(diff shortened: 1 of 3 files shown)


nickanderson:
This should also be picked to 3.6

basvandervlies:
+1

kacfengine:
And cherry-picked to 3.6.x.

zerkms:
Actually it brings some inconsistency now:

      "$(name)"
      file_select => all,
      depth_search => recurse_with_base($(depth)),
      delete => tidy;

      "$(name)/."
        delete => tidy;

As soon as you have a not infinite depth - you cannot guarantee that the directory is empty, hence you cannot perform directory removal.

tzz:
I think the promise outcome may be wrong, but the convergence passes will take care of the removal in either case.

Can you show a counter-example of how this might fail?

zerkms:
For some reason I thought that it will be a warning about being not able to remove the directory.

Just checked it (should have done it before actually) and it does not. So, sorry and nevermind, I was wrong :-)

tzz:
No problem :)  I was just as curious as you, but I've used it so I remembered it worked...

PR #428, opened 2015-02-25
Change: Replace hard coded time based classes

nickanderson:
Hard coding time based classes makes it difficult to allow policy execution
outside of that window without modifying the policy. This preserves the same
functionality while increasing flexibility which can be useful especially when
troubleshooting.

Ref: https://dev.cfengine.com/issues/7010

### DIFF
--- a/def.cf
+++ b/def.cf
@@ -258,7 +258,6 @@ bundle common def
         comment => "If reports are not collected for an extended period of time
                     the disk may fill up or cause additional collection
                     issues.";
-
 }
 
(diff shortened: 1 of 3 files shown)


nickanderson:
I think this should also be picked into 3.6

tzz:
It's good, except IMO it should use `and` on a list instead of `x.y.z` inside the expression.  But that's no big deal either way.

nickanderson:
thank you !

PR #429, opened 2015-02-27
Remove: *_not_repaired classes are a misnomer

nickanderson:
Each attribute is checked individually, one attribute not being repaired says
nothing about another attributes potential outcome.

Perhaps in the future we could have a conveniance combined outcome or
"compliance" outcome, where classes could be defined if all promsie attributes
were kept or repaired. Or perhaps a $(this.attribute) variable that would
allow automatic identification of the attribute that had the specified
outcome.

Ref: https://dev.cfengine.com/issues/7022

### DIFF
--- a/lib/3.5/common.cf
+++ b/lib/3.5/common.cf
@@ -213,10 +213,10 @@ body classes classes_generic(x)
 # Define x prefixed/suffixed with promise outcome
 {
       promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
-      repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
-      repair_denied => { "repair_denied_$(x)", "$(x)_denied", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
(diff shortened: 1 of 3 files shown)


kacfengine:
I agree fully with your reasoning, but isn't the same thing true for "not_kept"? A promise can have both "kept" and "not_kept" if you use both create and edit_line at the same time.

I'm a little bit reluctant to remove this because it has been there for a long time already (the whole 3.5 and 3.6 series), and each such change risks breaking the policy of someone that uses it. /me still has memories of the *the comma* change!

nickanderson:
yes but _not_kept is set when we know a repair failed in some way, and kept is set when we know that it was kept in some way. not_repaired, is special in that its set when the promise is seen to be kept in some way. Its hard to articulate, but basically not_repaired is an assumption where not_kept is a fact (it just does not exclude also being kept in some other way).

if I recall correctly, @fanf may have used this _not_repaired.

@estenberg - thoughts?

estenberg:
In general I don't think this is very important and am hesitant to remove stuff like this that can break backward compat to "clean up".

However _not_repaired seems to be defined in all cases, so I can't see how anyone wants to rely on this. This looks like a bug.

So I'd say this one is safe to merge, but don't quite understand the motivation behind it.

kacfengine:
> However _not_repaired seems to be defined in all cases, so I can't see how anyone wants to rely on this. This looks like a bug.

You're right, I suppose ``promise_kept`` will always be triggered when ``promise_repaired`` is, so it'll end up always being set. Ok, let's remove it then.

kacfengine:
Not critical by any means, so I'm not cherry-picking this to 3.6.x.

PR #430, opened 2015-03-03
Redmine #6401, #6402: Add systemd support to update_processes.cf.

kacfengine:
This will launch CFEngine using systemd when calling "cf-agent -f
update.cf" instead of launching them directly. This has the advantage
of bringing all daemons under systemd's control.

### DIFF
--- a/update/update_processes.cf
+++ b/update/update_processes.cf
@@ -54,14 +54,14 @@ bundle agent cfe_internal_update_processes
 
   methods:
 
-    am_policy_hub.enterprise::
+    am_policy_hub.enterprise.!systemd::
 
       "TAKING CARE CFE HUB PROCESSES"
       usebundle => maintain_cfe_hub_process,
       comment => "Call a bundle to maintian HUB processes",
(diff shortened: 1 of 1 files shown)


update/update_processes.cf
```diff
@@ -366,6 +372,22 @@ bundle agent maintain_cfe_windows
...
+  classes:
+    systemd::
+      "restart_cfe"
+      not => returnszero("/usr/bin/systemctl -q is-active cfengine3", "noshell"),
+      comment => "Check running status of CFEngine using systemd",
+      handle => "cfe_internal_maintain_cfe_systemd_classes_restart_cfe";
+
+  commands:
+    restart_cfe::
+      "/usr/bin/systemctl -q start cfengine3"
```
tzz: Maybe also `enable` the service?

kacfengine: I opted not to, because we enable it during package install, and the user might have removed CFEngine from the startup on purpose. Unlikely I guess, but it's how it works for the non-systemd platforms.

PR #432, opened 2015-03-09
Change: Stop replace_or_add bundle from adding extra lines

nickanderson:
Classes defined as the result of a promise outcome are by default namespace
scoped. This causes the classes to live on beyond the individual bundle
activation, resulting in bad behaviour for classes being defined unexpectedly.

Ref: https://dev.cfengine.com/issues/7035

This should be picked to 3.6.x.

Let me know if you think this should be updated to include all classes defined in edit_line bundles or all classes defined using the always classes body in edit_line bundles.


kacfengine:
trigger build

kacfengine:
+1, makes a lot of sense!

kacfengine:

PR #433, opened 2015-03-11
Change: Improve CPU info inventory

nickanderson:
dmidecode is unreliable, sometimes it outputs very badly formated processor
version information. Instead of using dmidecode, we now get it from /proc/cpu
(where we should have gotten it in the first place).

Additionally inventory for physical sockets, physical cores, and hyperthreding
has been added.

Ref: https://dev.cfengine.com/issues/7014

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -151,6 +151,59 @@ bundle agent cfe_autorun_inventory_proc
       "cpuinfo[$(cpuinfo_array[$(cpuinfo_idx)][0])]" string => "$(cpuinfo_array[$(cpuinfo_idx)][1])";
       "cpuinfo_keys" slist => getindices("cpuinfo");
 
+      "cpuinfo_hyperthreading"
+        string => ifelse("cpuinfo_hyperthreading_disabled", "disabled", "cpuinfo_hyperthreading_enabled", "enabled", "unknown"),
+        meta => { "inventory", "attribute_name=Hyperthreading", "derived-from=$(files[cpuinfo])" };
+
+      "cpuinfo_physical_cores"
+        string => "$(cpuinfo[cpu cores])",
+        ifvarclass => "have_cpuinfo_cpu_cores",
+        meta => { "inventory", "attribute_name=CPU physical cores", "derived-from=$(files[cpuinfo])" };
+
+      "cpuinfo_cpu_model_name"
+        string => "$(cpuinfo[model name])",
+        ifvarclass => "have_cpuinfo_model_name",
+        meta => { "inventory", "attribute_name=CPU model", "derived-from=$(files[cpuinfo])" };
+
+    # We need to be able to count the number of unique physical id lines in
+    # /proc/cpu in order to get a physical processor count.
+      "cpuinfo_lines" slist => readstringlist(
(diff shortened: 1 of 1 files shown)


estenberg:
What is the platform support for this now? Aren't we cutting it a bit by relying on /proc?
dmidecode is more portable, no?

Also, not sure if we should have "hyperthreading" by default - it seems a bit unreliable to detect and most CPUs have that (so it's not that interesting)?

inventory/any.cf
```diff
@@ -184,6 +265,16 @@ bundle agent cfe_autorun_inventory_proc
       scope => "namespace";
 
   reports:
+    DEBUG|DEBUG_cfe_autorun_inventory_proc::
```
estenberg: Not a big fan of this DEBUG/verbose blocks, it clutters the policy quite a bit.

estenberg:
Overall great.

inventory/any.cf
nickanderson: I don't know of another way to see variable values. Perhaps the new verbose output addresses this somehow. Either way being able to enable extra output is very useful for debugging and understanding what's going on.

tzz: `verbose_mode`?

nickanderson: I don't like to use verbose_mode alone because then its difficult to get just debugging output for an individual bundle. In its current state I also don't use verbose mode much since its hard to read.

nickanderson:
@estenberg regarding platform support:

According to the [dmidecode website](http://www.nongnu.org/dmidecode/):

```
Dmidecode is known to work on the following systems:
    Linux i386, x86-64, ia64
    FreeBSD i386, amd64
    NetBSD i386, amd64
    OpenBSD i386, amd64
    BeOS i386
    Cygwin i386
    Solaris x86
    Haiku i586
```
Of those only Linux is officially supported

So I don't think we are losing a lot, especially given that the inventory policy is most useful with Mission Portal in Enterprise.

Regarding hyperthreading, I have little opinion. It was just something that I thought was interesting while I was digging out other cpu details.

kacfengine:
I think it's not good to switch over to something known not to work on many platforms, even if those platforms are not officially supported by us. We might support them in the future, or people might be building them manually. Can't we scope it to Linux only and keep dmidecode for the others?

nickanderson:
I think CPU info needs to be its own abstract policy, it could use info provided by /proc/cpuinfo parsing, or from dmidecode based on platform. CPUinfo is the service, it may use dmidecode, it may use something else, but this is also part of a larger discussion on how I want to re-organize this policy. @estenberg keeps running from me.

kacfengine:
Ok, I talked a bit to Eystein, so the road ahead for the pull request is either:

1. Include this fix as is, and scope it to Linux only.
2. Think bigger and switch to CPUinfo with several backend implementations, depending on platform.

Not sure how much time you have, if you want to go for number 2, I suggest we close this pull request, since the end result is going to be quite different.

Personally, I think that option 2 sounds better in the long run.

nickanderson:
Superseded (I hope) by https://github.com/cfengine/masterfiles/pull/454

PR #438, opened 2015-03-23
Correct inventory_lslpp() package name regexp

rootwyrm:
The existing regexp is incorrect on AIX resulting in invalid package inventory.

Here is the format for lslpp -Lqc (using bos.64bit as the example.)
`bos.64bit:bos.64bit:6.1.8.0: : :C:F:Base Operating System 64 bit Runtime: : : : : : :0:0:/:1241`
`^^^^^^^^^` - this is hit
Problem: field 1 is not the package name. Field 1 is the package family. Thus, another example...
`sysmgt.websm:sysmgt.websm.apps:6.1.8.0: : :C:F:Web-based System Manager Applications: : : : : : :0:0:/:1241`
`sysmgt.websm:sysmgt.websm.diag:6.1.0.0: : :C: :Web-based System Manager Diagnostic Applications : : : : : : :1:0:/:0748`
The packages are `sysmgt.websm.apps` and `sysmgt.websm.diag` - however the existing regexp inventories both as `sysmgt.websm` with contradicting versions. This also occurs in bos.* and sys.* families. 

Note; this may misbehave on 5.3 but I do not have any 5.3 to check. If it does, 5.3 should be handled as a more specific class.

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -542,7 +542,7 @@ body package_method inventory_lslpp(update_interval)
       package_list_update_ifelapsed => $(update_interval);
 
       package_list_command       => "/usr/bin/lslpp -Lqc"; # list RPMs too
-      package_list_name_regex    => "([^:]+):.*";
+      package_list_name_regex    => ".:+([^:]+):.*";
       package_list_version_regex => "[^:]+:[^:]+:([^:]+):.*";
       package_installed_regex    => "[^:]+:[^:]+:[^:]+:[^:]+:[^:]+:([CA]):.*";
 


kacfengine:
5.3 appears to be the same, however packages with GNU software (gzip, for example) has a slightly different scheme where the version is included in the name. This is quite difficult to match correctly actually, because you need to match the name and at the same time ensure it doesn't contain a version, which regular expressions are not that good at. However, I was able to get a working expression by using non-greedy and lookahead operators. This is the expression I ended up with:

<pre>
package_list_name_regex => "[^:]+:([^:]+?(?=-[0-9])|[^:]+)[^:]*:.*";
</pre>

Can you try if that works for you?

rootwyrm:
@kacfengine The problem is that if somebody once again obnoxiously included the version in the string? That must be included. The whole thing.

Here's an example:
```
unzip:unzip-5.51-1:5.51-1: : :C:R:A utility for unpacking zip files.: :/bin/rpm -e unzip: : : : :0: :/opt/freeware:Mon Sep 27 15:40:54 EDT 2004

scd4 # lslpp -L unzip
  Fileset                      Level  State  Type  Description (Uninstaller)
  ----------------------------------------------------------------------------
  unzip                       5.51-1    C     R    A utility for unpacking zip files. (/bin/rpm)
```
Okay seems sane, right? Sigh.. don't I wish.

``` 
scd4 # rpm -qa
bash-3.2-1
screen-3.9.10-2
unzip-5.51-1
AIX-rpm-7.1.1.15-10
```

So we don't want to strip the version from the lslpp -Lqc output until such point as we're identifying the origin (fileset vs product vs component vs feature vs RPM), since the '-version' is the only way we currently have to tag something as an RPM requiring different handling; ```rpm -U``` vs. ```installp -u``` since installp can't/shouldn't operate on it. Basically tagging up inventory by looking for the hyphen in the package name with your policies and praying none of your other products did the same thing. Which somebody has or will. (So we can't just say "if it's -[0-9.*] it's RPM.")

Bit of a kludge, but it's a 3.6.x thing... it'll be much better in 3.7.x if I have to drag folks kicking and screaming. ;)

kacfengine:
Ah, so RPMs are listed by lslpp as well. That's evil... I'm not really sure how we should handle that. I suppose the only real way is to strip the RPM packages out and treat RPM as a separate package manager, with a separate package body? I think that's the only special treatment we can support, because CFEngine doesn't have a way to separate between different types of packages (that's where you would use a different package body, basically).

rootwyrm:
Innit though? And yes, that would be my preferred and recommended method (since the RPMs are almost never from IBM anyway.) But even the `name-version` is **not guaranteed** with the RPMs. Depends who packaged it. And unless we're taking inventory from `lslpp -Lqc` there is no guarantee of not introducing conflicting commits. So what I'll have to do longer term is introduce some way for the total inventory to also split it into installp/RPM categories. For now, it's more logical to just treat it as "tell us everything, we'll sort it later."

kacfengine:
Right, but for RPM we will just use the regular ``rpm`` command to do listings, so the whole ``name-version`` problem with ``lslpp -Lqc`` goes away. Alright, I will just split them then. Use ``lslpp`` body for bffs and ``rpm`` for RPMs.

rootwyrm:
For now this is the only safe fix, no matter how much I dislike that fact. The ONLY safe way to ID rpms is to get that list, then parse them out of lslpp. But there's no clear identifier for rpms other than rpm. Sigh. So it'll have to be on the admin for now.

kacfengine:
Yep.

This is my proposed fix then (I was able to get rid of the non-greedy/lookahead operators):

<pre>
diff --git a/inventory/any.cf b/inventory/any.cf
index df7b6d7..8858176 100644
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -542,9 +542,11 @@ body package_method inventory_lslpp(update_interval)
       package_list_update_ifelapsed => $(update_interval);
 
       package_list_command       => "/usr/bin/lslpp -Lqc"; # list RPMs too
-      package_list_name_regex    => "([^:]+):.*";
       package_list_version_regex => "[^:]+:[^:]+:([^:]+):.*";
-      package_installed_regex    => "[^:]+:[^:]+:[^:]+:[^:]+:[^:]+:([CA]):.*";
+      # Make sure version is not included in the name, that indicates RPM
+      # packages, which we should ignore.
+      package_list_name_regex    => "[^:]+:(([^-:]|-[^0-9])+):.*";
+      package_installed_regex    => "[^:]+:(([^-:]|-[^0-9])+):[^:]+:[^:]+:[^:]+
 
       package_name_convention    => "$(name)-$(version).+";
</pre>

Could you try if that produces the right list for you?

kacfengine:
Slight typo by me in the latest proposal, but I have verified that this works, and gives a better package overview. Submitted to master and 3.7.x.

PR #440, opened 2015-03-25
lib/3.7/files.cf: summarize file_make and file_make_mog with long strings

tzz:
@nickanderson for your review

I can backport this to 3.6, of course.  It's a pretty trivial change but makes the reports a lot more readable for large strings.

### DIFF
--- a/lib/3.7/files.cf
+++ b/lib/3.7/files.cf
@@ -2100,6 +2100,15 @@ bundle agent file_make(file, str)
 # and some more text here");
 # ```
 {
+  vars:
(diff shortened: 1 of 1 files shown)


kacfengine:
12 characters is maybe a bit too conservative, but looks useful otherwise. What do you think?

tzz:
I used 12 because it seemed reasonable :)  Any number is fine, really.

kacfengine:
Thanks, I upped it to 40 and submitted it. It makes strings like ``The file contains the information`` be a little more verbose than ``The ...tion``.

PR #441, opened 2015-03-31
Change: Exclude oracle hosts from redhat_pure

nickanderson:
Oracle Enterprise Linux hosts are not pure redhat and should be
excluded.

Ref: https://dev.cfengine.com/issues/7081

### DIFF
--- a/inventory/redhat.cf
+++ b/inventory/redhat.cf
@@ -4,7 +4,7 @@ bundle common inventory_redhat
 # This common bundle is for Red Hat Linux inventory work.
 {
   classes:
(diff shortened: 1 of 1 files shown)


tzz:
YES

estenberg:
Looks good to me.
As you noted earlier this would also add exclude nodes with hostname "oracle" but I don't think this is any different than other OSes (hostname debian, raspberry and so on) - so the risk should be acceptable.

nickanderson:
The only thing being that I dont know of any large enterprise databases named debian or raspberry, but I do know one named Oracle.

PR #443, opened 2015-03-31
Add FreeBSD inventory module

skreuzer:
In the inventory_control agent, define default values for dmidecoder,
lldpctl_exec, lsb_exec, mtab and proc and redefine the dmidecoder variable on
freebsd with the correct path to the dmidecode binary

Explicitly define the disable_inventory_proc class on FreeBSD since procfs will
create a /proc directory but none of the files defined in the basefiles slist
exist on FreeBSD

### DIFF
--- a/def.cf
+++ b/def.cf
@@ -266,12 +266,19 @@ bundle common inventory_control
 # This common bundle is for controlling whether some inventory bundles
 # are disabled.
 {
-  vars:
+  defaults:
+
       "dmidecoder" string => "/usr/sbin/dmidecode";
       "lldpctl_exec" string => "/usr/bin/lldpctl";
       "lsb_exec" string => "/usr/bin/lsb_release";
       "mtab" string => "/etc/mtab";
       "proc" string => "/proc";
+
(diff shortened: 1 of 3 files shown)


FlorianHeigl:
Please assign at least?

def.cf
```diff
@@ -266,13 +266,20 @@ bundle common inventory_control
...
       "lldpctl_exec" string => "/usr/bin/lldpctl";
       "lsb_exec" string => "/usr/bin/lsb_release";
       "mtab" string => "/etc/mtab";
       "proc" string => "/proc";
 
+  vars:
+
+    freebsd::
+
+      "dmidecoder" string => "/usr/local/sbin/dmidecode";
```
tzz: Better IMO, because `defaults` can have issues:

```
"dmidecoder" string => ifelse("freebsd", "/usr/local/sbin/dmidecode", "/usr/sbin/dmidecode");
```

kacfengine:
Merged manually. Thanks, and sorry for the delay.

PR #444, opened 2015-04-07
Add: IPv4 address inventory

nickanderson:
Previously this inventory came directly from the agent. However in order to
provide easier user selectable filtering it has moved to policy.

Ref: https://dev.cfengine.com/issues/7097

Should be merged with change to core:
https://github.com/cfengine/core/pull/2163

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -54,6 +54,10 @@ bundle agent inventory_autorun
 
       "loadaverage" usebundle => cfe_autorun_inventory_loadaverage(),
       handle => "cfe_internal_autorun_loadaverage";
+
+      "ipv4 addresses"
+        usebundle => cfe_autorun_inventory_ipv4_addresses,
+        handle => "cfe_internal_autorun_ipv4_addresses";
 }
 
 bundle agent cfe_autorun_inventory_listening_ports
@@ -65,6 +69,32 @@ bundle agent cfe_autorun_inventory_listening_ports
   vars:
       "ports" slist => sort( "mon.listening_ports", "int"),
       meta => { "inventory", "attribute_name=Ports listening" };
+}
+
+bundle agent cfe_autorun_inventory_ipv4_addresses
+# @brief Inventory ipv4 addresses
+# This will filter the loopback address (127.0.0.1, as it is likely not very interesting)
+{
+  vars:
+    "filter_reg"
+      string => "127\.0\.0\.1",
+      comment => "Addresses that match this regular expression will be filtered
+                  from the inventory for ipv4 addresses";
+
+    # Strings are displayed more beautifully in Mission Portal than lists, so
+    # we first generate the list of addresses to be inventoried and then do
+    # inventory using an array.
+    "ipv4_addresses"
+      slist => filter( $(filter_reg), "sys.ip_addresses", "true", "true", 999999999);
+
+    "ipv4[$(ipv4_addresses)]"
+      string => "$(ipv4_addresses)",
+      meta => { "inventory", "attribute_name=IPv4 addresses" };
+
+  reports:
+    DEBUG|DEBUG_cfe_autorun_inventory_ipv4_addresses::
+      "DEBUG $(this.bundle)";
+      "$(const.t)Inventorying: '$(ipv4_addresses)'";
 }
 
 bundle agent cfe_autorun_inventory_disk


def.cf
```diff
@@ -258,6 +258,11 @@ bundle common def
         comment => "If reports are not collected for an extended period of time
                     the disk may fill up or cause additional collection
                     issues.";
+
+      "enable_inventory_ipv4_addresses" -> { "Mission Portal" }
```
estenberg: Don't see any reason to make this configurable, at least not on the top level.. it should be very performant.

inventory/def.cf
```diff
@@ -0,0 +1,26 @@
+bundle agent inventory_def_file_control
+{
+  vars:
+  # Each input should be guarded with both an enable and !disable class so that
```
estenberg: Can we avoid this? enable makes the most sense, this will become a mess over time.

inventory/ipv4_addresses/inventory_ipv4_addresses.cf
```diff
@@ -0,0 +1,24 @@
+bundle agent inventory_ipv4_addresses
+{
+  vars:
+    "filter_reg"
+      string => "127\.0\.0\.1",
```
estenberg: Please filter for all localhost, not just the one 127.0.0.1 address by default.

estenberg:
Looks good, but should be cut down a bit and kept a bit simpler & maintainable, I think.

Also the structure of the inventory is interesting, do we have a plan/idea for this?
I see inventory/ipv4_addresses/inventory_ipv4_addresses.cf as a bit too much - will we have that much inventory about ipv4 addresses?

inventory/ipv4_addresses/inventory_ipv4_addresses.cf
nickanderson: I would say that loopback addresses that are not 127.0.0.1 would be interesting, its definately not common. But I can change it to 127\.\d{0,3}\.\d{0,3}\.\d{0,3} if you prefer.

inventory/def.cf
nickanderson: If it's enabled by default, how should someone who does not need or want it disable it? Cancelling a class is not straightforward.

I can cite [this conversation on the help list](https://groups.google.com/d/msg/help-cfengine/aEjW_IqaFKw/S1T99ElDC4sJ) where at least I convinced @neilhwatson

def.cf
nickanderson: Yes, it should be performant, and for Enterprise users it makes sense to have it on by default, but it arguably provides very little benefit for non enterprise users.

Classes being in def.cf are less important to me than variables, the classes defined in def.json would apply regardless of where we document them in policy. In which file do you suggest this class be set?

inventory/ipv4_addresses/inventory_ipv4_addresses.cf
estenberg: OK, makes sense. I thought you tried to filter out localhost, not just a pattern of common addresses. But I see now what you're doing, so it makes sense!

inventory/def.cf
estenberg: I'd say that wanting to disable this is such an uncommon occurrence that it should not be optimized for. Optimizing for it has costs, as is pretty clear in this example.

nickanderson:
Regarding the structure, yes it seems a bit much, but that's because this is an extremely simple policy. I made it go under a directory to support additionaly ipv4 inventory things. Perhaps I should call the directory ipv4 instead of ipv4_addresses.

I would like to move the rest of inventory into the same structure, organized by service, and the policy is only loaded if enable.!disable. So that users can easily enable or disable what they don't want by defining extra classes in def.json. This way policy is only parsed if the policy is going to be used. The general plan for the layout is similar to my policies listed in the [content registry](https://github.com/cfengine-content/). But as these policies ship inside of masterfiles proper the extra sub directories seemed a bit of overkill.

For example proc might be more complex. It might have one file for proc, and one for each other specific thing inside proc like cpuinfo.

inventory/def.cf
nickanderson: Its not about optimizing this case, its about setting a standard practice to follow, so that any policy can be easily toggled, and you can know what to expect because they are all enabled or disabled the same way.

tzz: Rather that classes as we did in 3.6, maybe the inventory control should be through a central slist.  Then it's easy to control its contents from a central UI in MP, and there's no need for enable/disable classes.

tzz:
I would use `inventory/network.cf` and put all network-related stuff in there.  Until that inventory policy grows above 1000 lines, I wouldn't worry about it (that's about the limit of what I consider manageable in a single policy file).

inventory/def.cf
nickanderson: I think classes are the right way to signal this. It allows you to collect inventory for a subset of your infrastructure instead of for the whole thing. To implement the same then each host would have to have its own centrally defined list.

nickanderson:
I dont like scrolling so I prefer to keep files small, and only the bundles that I need in that one file so that I can avoid parsing and pre-eval for bundles that I wont use controlled using the enable.!disable classes.

nickanderson:
Ugh, I cant use body file control inputs because it wasn't in 3.5, this will need to be changed around to support 3.5

nickanderson:
Oh, never mind, inventory policy isnt loaded on 3.5 clients anyway, i can.

nickanderson:
@estenberg refactored to fall in line with current layout.

nickanderson:
trigger build

estenberg:
Looks nice, only minor thing is that I'd put it in network.cf like Ted suggests, but no strong opinion.

Also we need to figure out the UI stuff. And before you ask, you don't have a category from policy...yet. :)

nickanderson:
Oh common,  my birthday is coming up in Sept.

nickanderson:
Oh and I chose any.cf because that's where ports listening was.

PR #445, opened 2015-04-09
Change: Lexicographically sort autorun bundles

nickanderson:
It is nice to be able to predict the order in which bundles are activated and
at times necessary for class dependancies.

Ref: https://dev.cfengine.com/issues/7102


tzz:
YES

PR #452, opened 2015-04-17
Change: Reorganize policy

nickanderson:
Not all policy that ships with CFEngine is applicable or even useful to
community users. This work is to seperate the enterprise specific policy from
the core policy and only load the enterprise specific policy when necessary.

Additionally this is intended to make user entry points more clear, and easier to upgrade the policy framework in the future

PR #454, opened 2015-04-21
Change: Improve CPU info inventory

nickanderson:
dmidecode is unreliable, sometimes it outputs very badly formated processer
version information, sometimes it cannot get processor information at all.
/proc/cpuinfo is a better source (at least on linux), and windows doesnt use
dmidecode, it uses another method but offers the data as if it were from
dmidecode.

This change abstracts cpuinfo into its own bundle. If /proc/cpuinfo is 
avaialable it uses that (based on parsing proc), and if only dmidecode is
available it will use that.

This does not completly detangle windows from dmidecode, it only detangles CPU
model name from being inventoried directly in the dmidecode bundle. This can
serve as initial work to continue detangling other information gathered in the
dmidecode bundle that is not actually from dmidecode.

Additionally there is stubbed out support for inventorying hyperthreading. It
is left commented out as its detection is not completely straight forward, and
it has questionable value.

Ref: https://dev.cfengine.com/issues/7014

Refactored based on commentary in https://github.com/cfengine/masterfiles/pull/433

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -215,34 +215,8 @@ bundle agent cfe_autorun_inventory_proc_cpuinfo
       "have_cpuinfo_cpu_cores" expression => strcmp("cpu cores", "$(default:cfe_autorun_inventory_proc.cpuinfo_array[$(default:cfe_autorun_inventory_proc.cpuinfo_idx)][0])");
       "have_cpuinfo_model_name" expression => strcmp("model name", "$(default:cfe_autorun_inventory_proc.cpuinfo_array[$(default:cfe_autorun_inventory_proc.cpuinfo_idx)][0])");
 
-      # Stubbed out hyperthreading, works but difficult to detect reliably? Limited value?
-        ## If siblings is greater (usually twice as much) than cores, HT is likely enabled
-        #"cpuinfo_siblings_greater_than_cores"
-        #  expression => isgreaterthan("$(default:cfe_autorun_inventory_proc.cpuinfo[siblings])", "$(default:cfe_autorun_inventory_proc.cpuinfo[cpu cores])"),
-        #  comment => "If number of cores is equal to number of siblings then
-        #              hyperthreading is OFF.
-        #              Ref: http://linuxhunt.blogspot.com/2010/03/understanding-proccpuinfo.html";
-
-        ## HT can't be enabled if the flag isn't present
-        #"cpuinfo_flag_ht"
-        #  expression => regcmp("\bht\b", "$(default:cfe_autorun_inventory_proc.cpuinfo[flags])");
-
-        #"cpuinfo_hyperthreading_enabled"
-        #  and => { "cpuinfo_flag_ht", "cpuinfo_siblings_greater_than_cores" };
-
-        #"cpuinfo_hyperthreading_disabled"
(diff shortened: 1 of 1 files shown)


nickanderson:
@estenberg @kacfengine 
CPUinfo abstracted to its own bundle. Model name is provided by proc/cpuinfo if avaialble, if not then its provided by the dmidecode bundle

nickanderson:
trigger_build

kacfengine:
@nickanderson: You were not in the builder group earlier, but I have added you now, so it should work in the future. I also triggered this one and #452 again.

Note that the string needs to be exactly "trigger build", with nothing else in the comment.

kacfengine:
trigger build

nickanderson:
Ah thanks

kacfengine:
HP-UX failed, but it's an unrelated spurious failure, so should be good.

inventory/any.cf
```diff
@@ -196,6 +203,141 @@ bundle agent cfe_autorun_inventory_proc
...
+    "_have_cpuinfo" expression => isvariable("default:cfe_autorun_inventory_proc.cpuinfo_idx");
+
+      # So that we don't inventory non dereferenced variables we check to see
+      # if we have the info first This is only necessary because its currently
+      # invalid to do isvariable on an array key that contains a space
+      # Ref: redmine#7088 https://dev.cfengine.com/issues/7088
+      "have_cpuinfo_cpu_cores" expression => strcmp("cpu cores", "$(default:cfe_autorun_inventory_proc.cpuinfo_array[$(default:cfe_autorun_inventory_proc.cpuinfo_idx)][0])");
+      "have_cpuinfo_model_name" expression => strcmp("model name", "$(default:cfe_autorun_inventory_proc.cpuinfo_array[$(default:cfe_autorun_inventory_proc.cpuinfo_idx)][0])");
+
+      # Stubbed out hyperthreading, works but difficult to detect reliably? Limited value?
```
estenberg: I would not leave hyperthreading in a dangling state in comments. Just take it out, or make sure that it works 100% everywhere and take it in.

inventory/any.cf
```diff
@@ -337,8 +486,8 @@ bundle agent cfe_autorun_inventory_dmidecode
       "dmi[system-version]" string => $(bios_array[4]),
       meta => { "inventory", "attribute_name=System version" };
 
-      "dmi[processor-version]" string => $(processor_array[1]),
-      meta => { "inventory", "attribute_name=CPU model" };
+      "dmi[processor-version]" string => $(processor_array[1]);
+      #meta => { "inventory", "attribute_name=CPU model" };
```
estenberg: Remove this line?

inventory/any.cf
```diff
@@ -196,6 +203,141 @@ bundle agent cfe_autorun_inventory_proc
...
+    # We only accept dmidecode values that don't look like cfengine variables,
+    # (starting with dollar), or that have an apparent empty value.
+    "_have_dmidecode_cpu_model_name"
+      not => regcmp("($(const.dollar)\(.*\)|^$)", "$(default:cfe_autorun_inventory_dmidecode.dmi[processor-version])");
+
+  vars:
+    _have_proc_cpu_physical_cores::
+      "cpuinfo_physical_cores"
+        string => "$(default:cfe_autorun_inventory_proc.cpuinfo[cpu cores])",
+        #ifvarclass => "have_cpuinfo_cpu_cores",
```
estenberg: Remove?

nickanderson:
trigger build

PR #457, opened 2015-05-07
Policy changes to use collect calls by default and not listen from clients.

ediosyncratic:
Blocked by https://dev.cfengine.com/issues/7156
Do not merge until that is resolved !
However, I wanted to document how to resolve 7157 and 7158 in policy, which this does.


nickanderson:
@estenberg we have not yet moved toward client initiated reporting shall we close out this old PR?

estenberg:
Yes, this is only for reference (a Redmine ticket pointing to it).

PR #458, opened 2015-05-11
Change: Class to guard linux inventory output

nickanderson:
Inform mode is supposed to be for changes that occur to the system. Printing
out the parsed contents of /etc/os-release does not really fit this ideal. I
have moved it to DEBUG|DEBUG_bundlename so policy maintainers can easily get
policy debug output as a whole or per bundle without also having other
additional inform or verbose agent output.


kacfengine:
trigger build

PR #459, opened 2015-05-12
Keys must always be pre-generated.

jimis:


jimis:
@nickanderson please review

nickanderson:
Ref: https://dev.cfengine.com/issues/7167

jimis:
So does it make sense to have `cf-key` being run from `update.cf`? I don't think so, personally...

nickanderson:
@jimis Does this need to goto 3.7.x also?

jimis:
cherry-picked to 3.7.x.

PR #462, opened 2015-05-19
Fix: Superflous reporting about chkconfig usage

nickanderson:
Inform mode is supposed to be for changes, previously a report was printed
each time the agent was run in inform mode even if there were no changes. This
guards the chkconfig report so that it is only printed under the same
conditions that trigger the command to be run.

Ref: https://dev.cfengine.com/issues/7196


nickanderson:
trigger build

kacfengine:
trigger build

kacfengine:
Build disappeared in Jenkins for some reason, running again. Looks good to me though.

PR #464, opened 2015-05-19
Change: Switch inform_mode reports to DEBUG|DEBUG_bundlename

nickanderson:
Inform mode is intended to show changes to the system. The current use of
inform_mode throughout the framework can lead to much more information than
needed.

Using a DEBUG|DEBUG_bundlename class guard allows this policy level debug
information to be reported for the entire policy run or for individual bundles
in a consistent way.

Ref: https://dev.cfengine.com/issues/7197

### DIFF
--- a/lib/3.7/commands.cf
+++ b/lib/3.7/commands.cf
@@ -39,7 +39,7 @@
 # the online docs
 #
 
-# For CFEngine Core: 3.7.0 to 3.7.x
+# For CFEngine Core: 3.6.0 to 3.6.x
 # Commands bodies
 
 ###################################################
@@ -66,8 +66,9 @@ bundle agent daemonize(command)
         contain => in_shell;
 
   reports:
-    windows.(inform_mode|verbose_mode)::
-      "$(this.bundle): This bundle does not support Windows";
+    DEBUG|DEBUG_daemonize::
+      "DEBUG $(this.bundle): This bundle does not support Windows"
+        ifvarclass => "windows";
 }
 
 ##-------------------------------------------------------
--- a/lib/3.7/commands.cf
+++ b/lib/3.7/commands.cf
@@ -39,7 +39,7 @@
 # the online docs
 #
 
-# For CFEngine Core: 3.6.0 to 3.6.x
+# For CFEngine Core: 3.7.0 to 3.7.x
 # Commands bodies
 
 ###################################################
@@ -66,7 +66,7 @@ bundle agent daemonize(command)
         contain => in_shell;
 
   reports:
-    DEBUG|DEBUG_daemonize::
+    DEBUG|DEBUG_$(this.bundle)::
       "DEBUG $(this.bundle): This bundle does not support Windows"
         ifvarclass => "windows";
 }
--- a/cfe_internal/enterprise/CFE_hub_specific.cf
+++ b/cfe_internal/enterprise/CFE_hub_specific.cf
@@ -132,8 +132,7 @@ bundle agent cfe_internal_apache_sudoer
 
   reports:
     (DEBUG|DEBUG_cfe_internal_apache_sudoer).!cfengine_internal_sudoers_editing_enable::
-      "DEBUG $(this.bundle)";
-      "$(const.t)editing of the sudoers file is disabled; the Apache user may not be able to run passwordless sudo cf-runagent";
+      "DEBUG $(this.bundle): editing of the sudoers file is disabled; the Apache user may not be able to run passwordless sudo cf-runagent";
 }
 
 #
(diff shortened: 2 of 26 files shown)


tzz:
YES to the classes.

I disagree with replacing `this.bundle` with `const.t`.  The tab is much less useful when you grep logs.  Please put the bundle name back.

nickanderson:
I used const.t on bundles that were to be activated a single time, I like the visual separation. For bundles that are activated multiple times I kept this.bundle on each line because you would lose the DEBUG this.bundle line at the beginning. I could probably be convinced of single line reports, but I am probably a bit timid on that because the single line verbose output has been difficult to read. More opinions welcome.

kacfengine:
I agree with @nickanderson about the ``const.t``, and this is more in line with how the logging format has changed for 3.7: Bundles are now printed as headers instead of line prefixes. It's only natural that verbose messages from policy do the same.

kacfengine:
trigger build

tzz:
I want to find the debug messages from a specific bundle in an ocean of information, I'd have to scan for the bundle header and then hope the reports are not obscured by other information.  `grep BUNDLENAME` is a much more direct way to get the data.  Other log collection tools like Splunk work similarly; the bundle header is not connected to the rest of the report.  So if you search for the bundle name, you have a hard time extracting the reports it generated.

The complaints I've heard about masterfiles reporting are about the inventory policy and that it's noisy on `inform_mode`.  And that's nicely addressed by this PR.  But I haven't heard anyone complaining about the bundle name in the report.

Look, you've changed the CFEngine log format in 3.7 to omit data, and that's fine.  But now you're changing the **masterfiles policy** report format from 3.6 to 3.7 for cosmetic ("visual separation") reasons and dropping a key piece of information to identify the source of the report.  Together with the other 3.7 logging changes, it turns reports into pretty text that's hard to grep and search.  I don't think that change is justified or functionally useful, and it's not even applied consistently as @nickanderson mentioned (multiple vs. single invocation).

nickanderson:
trigger build

nickanderson:
@tzz ping

tzz:
Looks very good.  I wondered if it makes sense to backport this change, since we tried to keep 3.5 stuff static except for critical bug fixes.  I think backporting is justified in this specific case, but it should be well documented.

@nickanderson I do have one other suggestion, which you'll **love** :)

No need to write the bundle name explicitly!  In 3.7, use the new variable string contexts.  Test the following with `-DDEBUG` and `-DDEBUG_main` and you'll see it works well:

```
bundle agent main
{
  reports:
    "DEBUG|DEBUG_$(this.bundle)"::
      "$(this.bundle): debug ON";
}
```

nickanderson:
Yeah I had considered doing that for 3.7 split files.

lib/3.7/commands.cf
```diff
@@ -39,7 +39,7 @@
 # the online docs
 #
 
-# For CFEngine Core: 3.7.0 to 3.7.x
+# For CFEngine Core: 3.6.0 to 3.6.x
```
tzz: probably unintended editing error

tzz:
Look great except for that minor version mismatch.

nickanderson:
ack fixed, thanks

nickanderson:
trigger build

kacfengine:
```
Error Message

FAIL (UNEXPECTED FAILURE) ./promises/promises.cf

Stacktrace

  notice: Q: "...-promises" -c -": ../../controls/3.7/def.cf:300:9: error: syntax error
Q: "...-promises" -c -":     DEBUG|DEBUG_$(this.bundle)::
Q: "...-promises" -c -":         ^
Q: "...-promises" -c -": ../../controls/3.7/def.cf:300:9: error: Expected promiser string, got "DEBUG"
Q: "...-promises" -c -":     DEBUG|DEBUG_$(this.bundle)::
Q: "...-promises" -c -":         ^
Q: "...-promises" -c -": ../../controls/3.7/def.cf:300:10: error: Expected ";", got "|"
Q: "...-promises" -c -":     DEBUG|DEBUG_$(this.bundle)::
Q: "...-promises" -c -":          ^
Q: "...-promises" -c -": ../../controls/3.7/def.cf:300:16: error: Expected promiser string, got "DEBUG_"
Q: "...-promises" -c -":     DEBUG|DEBUG_$(this.bundle)::
```

Looks like you'll need quotes around the classes to use ``$(this.bundle)``.

kacfengine:
I don't think we should backport this. It's a behavior change, and if we use the variable class contexts then we need to make the two changes different since variable class contexts won't work on pre 3.7.

nickanderson:
Superceeded by #503

PR #465, opened 2015-05-21
New package promise

kacfengine:
Package modules for apt and yum (see https://github.com/cfengine/core/pull/2215). At the time of writing only apt has been extensively tested, yum is still undergoing testing.

I will add commits to this branch as I find issues, but in the end we should squash it before merging.


kacfengine:
@ediosyncratic: Here are the package modules containing the Python code. The most important are the modules themselves (inside ``package_modules``). The ``the_great_package_test_generator.py`` script is less important, since it is not used in production.

The core changes are here: https://github.com/cfengine/core/pull/2215

Makefile.am
```diff
@@ -1,9 +1,12 @@
+SUBDIRS = \
+	package_modules \
+	tests/unit
```
ediosyncratic: Why is tests/unit/ in SUBDIRS but tests/acceptance is in TESTS_SUBDIRS ?
Surely the logic should be:
```
SUBDIRS = package_modules
TESTS_SUBDIRS = tests/unit
if HAVE_CORE
    TESTS_SUBDIRS += tests/acceptance
endif
SUBDIRS += $(TESTS_SUBDIRS)
```

configure.ac
```diff
@@ -111,6 +113,11 @@ AS_IF([test x"$prefix" = xNONE],
 )
 datadir=${prefix}/masterfiles
 
+dnl ######################################################################
+dnl OS specific stuff
```
ediosyncratic: (Petty) should be hyphenated: OS-specific

Makefile.am
kacfengine: Actually, ``TESTS_SUBDIRS`` should not be there at all, should just assign directly to ``SUBDIRS``. Will fix.

configure.ac
```diff
@@ -158,7 +165,9 @@ dnl Now make the Makefiles
 dnl ######################################################################
 
 AC_CONFIG_FILES([Makefile
-                tests/acceptance/Makefile])
+                package_modules/Makefile
+                tests/acceptance/Makefile
+                tests/unit/Makefile])
```
ediosyncratic: Can you use make variable expansion ? $(SUBDIRS:%=%/Makefile) would save the need for changes if you add more SUBDIRS.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+import subprocess
+import re
+
+dpkg_cmd = "/usr/bin/dpkg"
+dpkg_deb_cmd = "/usr/bin/dpkg-deb"
+dpkg_query_cmd = "/usr/bin/dpkg-query"
+dpkg_output_format = "Name=${Package}\nVersion=${Version}\nArchitecture=${Architecture}\n"
+dpkg_status_format = "Status=${Status}\n" + dpkg_output_format
+
+apt_get_options = ["-o", "Dpkg::Options::=--force-confold", "-o", "Dpkg::Options::=--force-confdef", "-y", "--force-yes"]
```
ediosyncratic: This would read more clearly with some line-breaks, so that each option and optional parameter appear on one line together, separate from the others.
```
apt_get_options = ["-o", "Dpkg::Options::=--force-confold",
                 "-o", "Dpkg::Options::=-force-confdef",
                 "-y",
                 "--force-yes"]
```
which would even make it easy for you add comments explaining what each is for and why we need it.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+apt_get_cmd = "/usr/bin/apt-get"
+
+os.environ['DEBIAN_FRONTEND'] = "noninteractive"
+os.environ['LC_ALL'] = "C"
+
+if (os.environ.get('CFENGINE_TEST_DPKG_CMD')):
+    dpkg_cmd = os.environ['CFENGINE_TEST_DPKG_CMD']
```
ediosyncratic: It would be cleaner and simpler to write
```
dpkg_cmd = os.environ.get('CFENGINE_TEST_DPKG_CMD', "/usr/bin/dpkg")
```
instead of initially setting to `"/usr/bin/dpkg"` and then conditionally calling `os.environ.get()` to decide whether to then over-write the initial setting with the `os.environ[]` value.
The .get(key, default) of a dictionary supplies the given default as value if the key is absent.

Also, in python, if doesn't need parentheses around its condition !
(... but the above simplification shall not only speed up this code but also get rid of the need for the conditional.)

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+if (os.environ.get('CFENGINE_TEST_DPKG_QUERY_CMD')):
+    dpkg_query_cmd = os.environ['CFENGINE_TEST_DPKG_QUERY_CMD']
+if (os.environ.get('CFENGINE_TEST_APT_GET_CMD')):
+    apt_get_cmd = os.environ['CFENGINE_TEST_APT_GET_CMD']
+
+def get_package_data():
+    pkg_string = ""
+    for line in sys.stdin:
+        result = re.match("^File=(.*)", line)
+        if (result):
```
ediosyncratic: Spurious parentheses.  You're writing python, not C :-)

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+    dpkg_deb_cmd = os.environ['CFENGINE_TEST_DPKG_DEB_CMD']
+if (os.environ.get('CFENGINE_TEST_DPKG_QUERY_CMD')):
+    dpkg_query_cmd = os.environ['CFENGINE_TEST_DPKG_QUERY_CMD']
+if (os.environ.get('CFENGINE_TEST_APT_GET_CMD')):
+    apt_get_cmd = os.environ['CFENGINE_TEST_APT_GET_CMD']
+
+def get_package_data():
+    pkg_string = ""
+    for line in sys.stdin:
+        result = re.match("^File=(.*)", line)
```
ediosyncratic: Is `sys.stdin` liable to contain many lines ?
If so, it'd be worth pre-compiling the regex:
```
pkg_string, file_re = "", re.compile("^File=(.*)")
for line in sys.stdin:
    result = file_re.match(line)
```

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        if (result):
+            pkg_string = result.group(1)
+            break
+
+    if (pkg_string == ""):
```
ediosyncratic:
```
if not pkg_string:
```
Rather than using a comparison, simply use the value as a bool.
The empty string is false.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        return 1
+
+    repo = False
+    if (re.match("^/.*", pkg_string)):
```
ediosyncratic: Gratuitous clumsiness; no need for a regex match, that's going to enumerate every byte of the string checking it matches `.*`, when you could just test
```
if pkg_string.startswith('/'):
```

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        sys.stdout.write("PackageType=file\n")
+    elif (re.search("([:,]|-[0-9])", pkg_string)):
+        # Contains either a version number or an illegal symbol.
+        sys.stdout.write(line + "ErrorMessage: Package string with illegal format\n")
+        return 1
+    else:
+        repo = True
+        sys.stdout.write("PackageType=repo\n")
+
+    if (repo):
```
ediosyncratic: There is precisely one branch of the above that leaves repo true, one that leaves it false; so simply moving the code from each branch of this to the corresponding branch above would eliminate the need for the repo variable and simplify the code.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        return 0
+    else:
+        sys.stdout.flush()
+        return subprocess.call([dpkg_deb_cmd, "--showformat", dpkg_output_format, "-W", result.group(1)])
+
+
+def list_installed():
+    for line in sys.stdin:
+        # Ignore everything.
+        pass
```
ediosyncratic: You apparently have some Need to consume stdin ?
`sys.stdin.readlines() # read and discard`
will do it for you in a single line ;-)

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+    process = subprocess.Popen([dpkg_query_cmd, "--showformat", dpkg_status_format, "-W"], stdout=subprocess.PIPE)
+    installed_package = False
+    for line in process.stdout:
+        if line[-1] == "\n":
+            line = line[:-1]
+        # The space before "installed" is important, because it can be "not-installed".
+        if re.match("^Status=.* installed.*", line):
+            installed_package = True
+        elif re.match("^Status=.*", line):
+            installed_package = False
```
ediosyncratic: Rather than using two regexes (that I'll be tempted to suggest you compile) you can simply use `.startswith` again:
```
if line.startswith('Status='):
    offset = line.find('installed')
    installed_package = offset > 7 and not line[:offset].endswith('not-')
```

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+def list_installed():
+    for line in sys.stdin:
+        # Ignore everything.
+        pass
+
+    process = subprocess.Popen([dpkg_query_cmd, "--showformat", dpkg_status_format, "-W"], stdout=subprocess.PIPE)
+    installed_package = False
+    for line in process.stdout:
+        if line[-1] == "\n":
+            line = line[:-1]
```
ediosyncratic: You really don't need to do this.
You do in fact know that each line ends in a '\n', so can just `sys.stdout.write(line)` below, without having to add a '\n' - unless you're paranoid that the last might lack its trailing newline.
In that case, below, `sys.stdout.write(line.rstrip() + '\n')` will take care of canonicalising line-ending.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        elif installed_package:
+            sys.stdout.write(line + "\n")
+
+    return 0
+
+
+def list_updates(online):
+    for line in sys.stdin:
+        # Ignore everything.
+        pass
```
ediosyncratic: Again, not sure why you feel the need to do this, but `sys.stdin.readlines()` does it better.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+
+    if online:
+        result = subprocess.call([apt_get_cmd] + apt_get_options + ["update"], stdout=sys.stderr)
+        if (result != 0):
```
ediosyncratic: More spurious parentheses.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        #           V              V                    V                        V                 V
+        #   Inst php5-cli [5.3.10-1ubuntu3.17] (5.3.10-1ubuntu3.18 Ubuntu:12.04/precise-updates [amd64]) []
+        #
+        # Note architecture included in the name on this one:
+        #   Inst php5-cli:i386 [5.3.10-1ubuntu3.17] (5.3.10-1ubuntu3.18 Ubuntu:12.04/precise-updates [i386]) []
+        #
+        # Note multiple repositories in this one:
+        #   Inst linux-libc-dev [2.6.32-48squeeze4] (2.6.32-48squeeze6 Debian:6.0.10/oldstable, Debian-Security:6.0/oldoldstable [amd64])
+        #
+        match = re.match("^Inst\s+([^\s:]+)(?::\S+)?\s+\[[^]\s]+\]\s+\((\S+)\s+[^,\s]+(?:,\s+[^,\s]+)*\s+\[([^]\s]+)\]\).*", line)
```
ediosyncratic: Now, quite apart from how hard this is to review (because I have to page down six times to find the scroll-bar I need in order to see the whole thing, then page back up again), this is illegible gibberish (as regexes so often are), which means that someone reading the code can't really tell if it does what your (excellent) comment claims.  It's wonderfully terse but anti-debuggable.
```
words = line.split()
if not words or words.pop(0) != 'Inst' or not words: continue
name = words.pop(0)
if words[0].startswith('['): words = words.pop(0) # discard old version, if present
if not words[0].startswith('('): continue
version = words.pop(0)[1:]
while not words[0].endswith(')'): words.pop(0) # discard repositories
arch = words.pop(0)[:-1]
if arch.startswith('[') and arch.endswith(']'): arch = arch[1:-1]
else: continue
sys.stdout.write('Name=%s\nVersion=%s\nArch=%s\n' % (name, version, arch))
```
I don't think the reader is in any doubt that this parses what you have above; it's possible to add comments where they belong; and someone looking for a bug can probably see where it is.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+            sys.stdout.write("Version=" + match.group(2) + "\n")
+            sys.stdout.write("Architecture=" + match.group(3) + "\n")
+
+    return 0
+
+
+def one_package_argument(name, arch, version, is_apt_install):
+    args = []
+    archs = []
+    if arch != "":
```
ediosyncratic: Simpler to just use arch as a boolean, `if arch:`, rather than doing an explicit comparison.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        # Here we have no specified architecture and we are installing.
+        # If we have existing versions, operate on those, instead of the
+        # platform default.
+        process = subprocess.Popen([dpkg_query_cmd, "--showformat", "${Architecture}=${Status}\n",
+                                    "-W", name + ":*"], stdout=subprocess.PIPE)
+        for line in process.stdout:
+            # The space before "installed" is important, because it can be "not-installed".
+            match = re.match("^([^=]*)=.* installed.*", line)
+            if match:
+                archs.append(match.group(1))
```
ediosyncratic: Again, regex is a heavy way to do this.
```
if '=' in line:
    arch, stat = line.split('=', 1)
    if stat.find(' installed'): archs.append(arch)
```

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+
+    version_suffix = ""
+    if version != "" and is_apt_install:
+        version_suffix = "=" + version
+
+    if len(archs) > 0:
+        for arch in archs:
+            args.append(name + ":" + arch + version_suffix)
```
ediosyncratic: args was previously empty, so we can just use a *list comprehension* in place of this loop:
```
args = [name + ':' + arch + version_suffix for arch in archs]
```
(had args not been ampty, using `args +=` would do the job equally).

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+            # The space before "installed" is important, because it can be "not-installed".
+            match = re.match("^([^=]*)=.* installed.*", line)
+            if match:
+                archs.append(match.group(1))
+
+    version_suffix = ""
+    if version != "" and is_apt_install:
+        version_suffix = "=" + version
+
+    if len(archs) > 0:
```
ediosyncratic: Simply use archs as a boolean; the semantics of sequences as booleans is that empty is false, anything else is true.
(Note that this means the list `[ '' ]` is true, albeit the sole entry is false; the list has an entry, so is true.)

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        args.append(name + ":*" + version_suffix)
+
+    return args
+
+
+def package_arguments_builder(is_apt_install):
+    name = ""
+    version = ""
+    arch = ""
+    args = []
```
ediosyncratic: It's a style choice, but when you get a lot of things to initialise, doing so in one line is an option (and I like it because it saves vertical space):
```
name, version, arch, args = '', '', '', []
```

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+    for line in sys.stdin:
+        result = re.match("^Name=(.*)", line)
```
ediosyncratic: Again with the re where `.startswith('Name=')` would do just fine.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        result = re.match("^Version=(.*)", line)
+        if (result):
+            version = result.group(1)
+            continue
+
+        result = re.match("^Architecture=(.*)", line)
+        if (result):
+            arch = result.group(1)
+            continue
+
```
ediosyncratic:
```
for line in sys.stdin:
    if line.startswith('Name='):
         if name:
            args += one_package_argument(name, arch, version, is_apt_install)
            version = arch = ''
         name = line[5:]
elif line.startswith('Version='): version = line[8:]
elif line.startswith('Architecture='): arch = line[13:]
```
Because you don't need to do the matching outside the flow of the `if-elif` chain, you save yourself a stack of bureaucracy and the need to continue.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+    return args
+
+
+def repo_install():
+    cmd_line = [apt_get_cmd] + apt_get_options + ["install"]
+
+    args = package_arguments_builder(True)
+
+    if (not args):
```
ediosyncratic: Spurious parentheses.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        if (result):
+            version = result.group(1)
+            continue
+
+        result = re.match("^Architecture=(.*)", line)
+        if (result):
+            arch = result.group(1)
+            continue
+
+    if name != "":
```
ediosyncratic: Use `name` as a boolean.

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        return 0
+
+    cmd_line.extend(args)
+
+    return subprocess.call(cmd_line, stdout=sys.stderr)
+
+
+def remove():
+    cmd_line = [dpkg_cmd]
+    cmd_line.append("--remove")
```
ediosyncratic:
```
cmd_line = [dpkg_cmd, '--remove']
```

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+
+    args = package_arguments_builder(False)
+
+    if (not args):
+        return 0
+
+    cmd_line.extend(args)
+
+    return subprocess.call(cmd_line, stdout=sys.stderr)
```
ediosyncratic:
```
if args:
    return subprocess.call(cmd_line + args, stdout=sys.stderr)
return 0

File: package_modules/apt_get [hunk 28]
```diff
@@ -0,0 +1,259 @@
...
+
+
+def file_install():
+    cmd_line = [dpkg_cmd]
+    cmd_line.append("-i")
```
ediosyncratic:
```
cmd_line = [dpkg_cmd, '-i']
```

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+        result = re.match("^File=(.*)", line)
+        if (result):
+            found = True
+            cmd_line.append(result.group(1))
+
+    if (not found):
+        return 0
+
+    return subprocess.call(cmd_line, stdout=sys.stderr)
+
```
ediosyncratic:
```
args = [line[5:] for line in sys.stdin if line.startswith('File=')]
if args:
    return subprocess.call(cmd_line + args, stdout=sys.stderr)
return 0
```
Writing the pythonic code rather than transcribing the perl code into python works so much better ...

package_modules/apt_get
```diff
@@ -0,0 +1,259 @@
...
+elif sys.argv[1] == "remove":
+    ret = remove()
+
+elif sys.argv[1] == "file-install":
+    ret = file_install()
+
+else:
+    sys.stderr.write("Invalid operation\n")
+    exit(2)
```
ediosyncratic: I'd be inclined to put this whole `if-elif` chain into a function main(), returning `ret` or the 2 on bad command-lines, and end with (below) `exit(main())`.  You could also turn the whole chain into a dictionary look-up, mapping `sys.argv[1]` to a tuple of function to call and sequence of parameters:
```

def main(args):
    if not args:
        sys.stderr.write("Need to provide argument\n") 
        return 2

    try:
        use = {
            "supports-api-version": (sys.stdout.write, "1\n"),
            "get-package-data": (get_package_data),
            "list-installed": (list_installed),
            "list-updates": (list_updates, True),
            "list-updates-local": (list_updates, False),
            "repo-install": (repo_install),
            "remove": (remove),
            "file-install": (file_install)
            }[args[0]]
    except KeyError:
        sys.stderr.write("Invalid operation: %s\n" % args[0])
        return 2
    return use[0](*use[1:])

exit(main(sys.argv[1:]))
```

package_modules/yum
```diff
@@ -0,0 +1,216 @@
...
+rpm_cmd = "/bin/rpm"
+rpm_output_format = "Name=%{name}\nVersion=%{version}\nArchitecture=%{arch}\n"
+
+yum_cmd = "/usr/bin/yum"
+yum_options = "--quiet"
+
+if (os.environ.get('CFENGINE_TEST_RPM_CMD')):
+    rpm_cmd = os.environ['CFENGINE_TEST_RPM_CMD']
+if (os.environ.get('CFENGINE_TEST_YUM_CMD')):
+    yum_cmd = os.environ['CFENGINE_TEST_YUM_CMD']
```
ediosyncratic: As for apt, this can all be simplified by using `.get()` properly.

package_modules/yum
```diff
@@ -0,0 +1,216 @@
...
+    else:
+        repo = True
+        sys.stdout.write("PackageType=repo\n")
+
+    if (repo):
+        sys.stdout.write("Name=" + pkg_string + "\n")
+        return 0
+    else:
+        sys.stdout.flush()
+        return subprocess.call([rpm_cmd, "--qf", rpm_output_format, "-qp", result.group(1)])
```
ediosyncratic: As for apt, one branch above sets repo true, one leaves it false; so folding the two branches of the condition on it into those can simplify this and eliminate the variable repo.

package_modules/yum
```diff
@@ -0,0 +1,216 @@
...
+
+def list_updates(online):
+    for line in sys.stdin:
+        # Ignore everything.
+        pass
+
+    if online:
+        online_flag = []
+    else:
+        online_flag = ["-C"]
```
ediosyncratic: There is a pythonic equivalent of C's ?: operator:
```
online_flag = [] if online else ['-C']
```

package_modules/yum
```diff
@@ -0,0 +1,216 @@
...
+
+    process = subprocess.Popen([yum_cmd, yum_options] + online_flag + ["check-update"], stdout=subprocess.PIPE)
+    lastline = ""
+    for line in process.stdout:
+        # Combine multiline entries into one line. A line without at least three
+        # space separated fields gets combined with the next line, if that line
+        # starts with a space.
+        if len(lastline) > 0 and line[0] != " ":
```
ediosyncratic: Do you really mean `line[0] != ' '` (only the simple space character) or would `not line[0].isspace()` be more apt ?

Also, `len(lastline) > 0` is better phrased as simply `lastline`, using it as a boolean with less effort and the same effect.

package_modules/yum
```diff
@@ -0,0 +1,216 @@
...
+            # Keep line but strip trailing newline.
+            lastline = line[:-1]
+            continue
+
+        lastline = ""
+        match = re.match("^(\S+)\.([^.\s]+)\s+(\S+)\s+\S+\s*$", line)
+        if match != None:
+            sys.stdout.write("Name=" + match.group(1) + "\n")
+            sys.stdout.write("Version=" + match.group(3) + "\n")
+            sys.stdout.write("Architecture=" + match.group(2) + "\n")
```
ediosyncratic: From `lastline = ""` onwards:
```
words = []
for line in process.stdout:
    if not line[0].isspace(): words = []
    words += line.split()
    if len(words) > 3: words = [] # discard over-long
    elif len(words) == 3:
        dot = words[0].split('.', 1)
        if '.' not in dot[1]: # else discard malformed
           sys.stdout.write('Name=%s\nVersion=%s\nArchitecture=%s\n' % (dot[0], words[1], dot[1]))
        words = []
```

package_modules/apt_get
teroka: [re.VERBOSE](https://docs.python.org/2/library/re.html#re.VERBOSE) is useful when regex's are getting unwieldy.

ediosyncratic:
Instead of repeating myself for the yum script, I just sent you a pull request to your pull request, that does all the things I was suggesting the the apt_get script and does similar to yum :-)

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
```diff
@@ -0,0 +1,540 @@
...
+
+# Given the transitions between the two version, determine if the change of one
+# architecture package will trigger a change in the other. Only one of the two
+# inputs can be different, but the returned value may cause both to be
+# different. This is only relevant for "repo" style promises, since "file"
+# promises will never touch more than one package at a time.
+def resolve_arch_conflicts(from_64, from_32, to_64, to_32):
+    if from_64 != to_64:
+        if from_32 != "0":
+            if to_64 >= from_32:
```
ediosyncratic: Are you comparing strings ?
Should you be comparing numbers instead ?
Remember that: '1000000' < '2'

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
```diff
@@ -0,0 +1,540 @@
...
+
+    return (to_64, to_32)
+
+
+# Simulate, with the given promise, what the state of the system would be,
+# if we started from the versions in from_64 and from_32.
+def simulate_promise(promise, from_64, from_32):
+    to_64 = from_64
+    to_32 = from_32
+
```
ediosyncratic: I think this would all be *hugely* easier to read if you started with
```
arch, version = promise.get('arch'), promise.get('version')
```
and similar for anything else you `.get()` repeatedly otherwise.  It would incidentally also run faster.

ediosyncratic: I'll do that on my pull-request ;-)

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
```diff
@@ -0,0 +1,540 @@
...
+    promise_candidates = get_possible_promises()
+    valid_promises = []
+
+    for promise in promise_candidates:
+        result = simulate_promise(promise, from_64, from_32)
+        if result[0] == to_64 and result[1] == to_32:
+            valid_promises.append(promise)
+
+    return valid_promises
```
ediosyncratic: This looks suspiciously like a case where a generator would be preferrable.
Throw away valid_promises and replace its `.append(promise)`s with `yield promise`; delete the return.
The result of calling this function is then an iterable that iterates over the things you would have appended to your list, only you don't have to collect them all up in memory at the same time.
Do the same to get_possible_promises() and you may well save a lot of hassle.

I'll do this in my pull request ...

ediosyncratic: hmm ... no, you want to iterate each of these twice, once for 'do' and once for 'count', and the iterable would be used up the first time round.  If you run into memory problems with running this script, it'd still be worth doing (just move the `calc_all_transitions()` call inside the 'count', 'do' loop) but I'm guessing memory isn't a problem for you.

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
```diff
@@ -0,0 +1,540 @@
...
+        return "latest"
+    else:
+        return "$(p.version[" + version + "])"
+
+
+# Write one test case, using one promise to go from one state to another (or the
+# same) state.
+# This function is quite hard to read because of all the quoting going on, it
+# might be easier to look at the output it produces.
+def make_test(current_count, total_test_count, from_state, to_state, transition, test_handle, warn_only):
```
ediosyncratic: The interleaving of python and cf-policy language makes this horrific to read !

tests/acceptance/17_packages/12_new/timed/test_module
```diff
@@ -0,0 +1,40 @@
...
+
+for line in sys.stdin:
+    match = re.match("^File=(.*)", line)
+    if match:
+        pkg_name = match.group(1)
+
+if sys.argv[1] == "list-installed":
+    sys.stdout.write("Name=cfe-present-package\nVersion=1\nArchitecture=amd64\n")
+elif sys.argv[1] == "list-updates" or sys.argv[1] == "list-updates-local":
+    log = open(module_dir + "/test_module.log", "r")
```
ediosyncratic: You leaked this.
Fixed on my branch, along with tidy-up of this script.

tests/unit/test_package_module_apt_get
```diff
@@ -0,0 +1,175 @@
...
+    try:
+        os.unlink(mock_log)
+    except OSError:
+        pass
+
+    return True
+
+apt_get_options = "-o Dpkg::Options::=--force-confold -o Dpkg::Options::=--force-confdef -y --force-yes"
+
+assert(check("supports-api-version", [], 1, ["1"], 0, []))
```
ediosyncratic: Note that assert isn't a function in python; it's a keyword.  So your parentheses are spurious; and you have the option of providing a subsequent message, output on failure, characterising what just failed, e.g.
```
assert sanity_check(), 'we went crazy'
```

Fixed, along with a few other petty details, in my PR to your PR.

ediosyncratic:
OK, I've reviewed the python (not the *.cf) and send you a PR for most of what I found !

kacfengine:
Thanks, I'll take a look at this soon, I just need to handle https://dev.cfengine.com/issues/7161 first.

package_modules/apt_get
basvandervlies: If you want to use regex, you also can give logical names (group names), eg:
``
blog_regex = '''
    (?P<blog>blog(?P<blog_params>[?][^:]*)?:(?P<blog_id>\S*))
'''
blog_re = re.compile(blog_regex, re.VERBOSE)
result = blog_re.search(<str>)
```
Now you can use:
'''
                if result.group('blog'):
                    self.system = 'blog'
                    self.blog(m, subject[result.end('blog'):], result.group('blog_id'), result.group('blog_params'))
'''

File: configure.ac (hunk 3 above)
Comment (kacfengine) [inline configure.ac]: No, it's not a Makefile, it's m4 and shell script. While some shell magic might be possible here, not all of these occurrences are as trivial as this one (check out core for example), so I'd rather keep the standard.

File: package_modules/apt_get (hunk 4 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 5 above)
Comment (kacfengine) [inline package_modules/apt_get]: Aha! Another gem I didn't know about Python. Fixed!

File: package_modules/apt_get (hunk 7 above)
Comment (kacfengine) [inline package_modules/apt_get]: Python documentation states that you don't have to, it will pre-compile for you for "small" programs. It makes no mention of what a small program is, but I doubt any of these will exceed the limit.

File: package_modules/apt_get (hunk 6 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 8 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 9 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 10 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 11 above)
Comment (kacfengine) [inline package_modules/apt_get]: Indeed, fixed!

File: package_modules/apt_get (hunk 12 above)
Comment (kacfengine) [inline package_modules/apt_get]: I think handling offsets is both more error prone and less clear than letting the regex handle it. Speed is not an issue here, since the package manager will have execution time orders of magnitude higher than the script.

The second one can use ``startswith()`` though, since it is 100% equivalent. Fixed that.

**Comment by kacfengine** [comment]
I haven't gone through everything yet, so the branch is not updated. Will continue next week.

File: package_modules/apt_get (hunk 6 above)
Comment (tzz) [inline package_modules/apt_get]: This fix and others haven't been pushed.

**Comment by tzz** [comment]
For the APT interface, please consider the https://apt.alioth.debian.org/python-apt-doc/library/index.html or https://pypi.python.org/pypi/apt-wrapper/1.08 instead of rolling your own regexes etc.

For Yum, take a look at http://yum.baseurl.org/wiki/5MinuteExamples http://www.plugged.in/linux/using-the-python-api-for-yum.html and http://yum.baseurl.org/api/yum-3.2.27/ for the full API.

I think those modules will simplify the package managers greatly and requiring them is not unreasonable (you could even package them in the distribution as a fallback).

**Comment by basvandervlies** [comment]
I had the same remark as @tzz there are good python libraries available for this purpose

**Comment by kacfengine** [comment]
@tzz, @basvandervlies: The YUM Python API was considered, but it is very low level, with limited documentation. We decided that calling the commands directly was easier to achieve, and has a couple of advantages, like much higher reusability of yum code in implementing apt, and reusing knowledge we already have in masterfiles.

That doesn't mean it can't be improved later though, some of the more complicated parsing could certainly be avoided by using the yum library.

File: package_modules/apt_get (hunk 13 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed by replacing offset by ``rstrip``.

File: package_modules/apt_get (hunk 15 above)
Comment (kacfengine) [inline package_modules/apt_get]: The parenthesis reflex is deeply embedded! Fixed!

File: package_modules/apt_get (hunk 16 above)
Comment (kacfengine) [inline package_modules/apt_get]: I guess it is a matter of opinion regarding reading regexes, but I think the coded approach is even harder to read than the regex, and more error prone as well. I've kept the regex, but I've added markers where each section begins and ends, for better readability, and split the expression on two lines. I also added names for the sub groups that are extracted.

File: package_modules/apt_get (hunk 17 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 18 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 19 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 20 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 21 above)
Comment (kacfengine) [inline package_modules/apt_get]: I think I prefer the existing style. It's a bit more lines, but easier to read, and much easier to edit later.

File: package_modules/apt_get (hunk 23 above)
Comment (kacfengine) [inline package_modules/apt_get]: I'm allergic to offsets though, they are begging to introduce bugs later. Regexes are simply safer.

File: package_modules/apt_get (hunk 24 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 25 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 26 above)
Comment (kacfengine) [inline package_modules/apt_get]: Doh! Fixed!

File: package_modules/apt_get (hunk 27 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 28 above)
Comment (kacfengine) [inline package_modules/apt_get]: Fixed!

File: package_modules/apt_get (hunk 29 above)
Comment (kacfengine) [inline package_modules/apt_get]: Nah, offsets again.

File: package_modules/apt_get (hunk 30 above)
Comment (kacfengine) [inline package_modules/apt_get]: Turned it into a ``main()`` function, but kept the structure.

File: package_modules/yum (hunk 32 above)
Comment (kacfengine) [inline package_modules/yum]: Fixed!

File: package_modules/yum (hunk 33 above)
Comment (kacfengine) [inline package_modules/yum]: Aha, fixed!

File: package_modules/yum (hunk 34 above)
Comment (kacfengine) [inline package_modules/yum]: You're right, should probably be using ``isspace()``.

File: package_modules/yum (hunk 35 above)
Comment (kacfengine) [inline package_modules/yum]: Kept the regex, but assigned names to the sub groups.

File: package_modules/apt_get (hunk 23 above)
Comment (ediosyncratic) [inline package_modules/apt_get]: If you look at my PR, you'll see I used an alternative pattern that avoids the offsets: e.g.
```
elif line.startswith('Version='): version = line.split('=', 1)[1]
```
which is at least as robust and safe as a regex, without having to traverse the rest of the line making sure each byte matches /./.

File: package_modules/yum (hunk 34 above)
Comment (ediosyncratic) [inline package_modules/yum]: One caveat there: the line might be empty, in which case `line == '\n'` and `line[0].isspace()` is true.

Comment (ediosyncratic) [inline package_modules/yum]: ... so you might want `line[0].isspace() and line[0] != '\n'`` :-(

Comment (kacfengine) [inline package_modules/yum]: That's ok, since adding an empty line will once again not match the regex, and then continue with the next line.

File: package_modules/apt_get (hunk 23 above)
Comment (kacfengine) [inline package_modules/apt_get]: Ok, that's better, I have replaced all the simple cases with that.

Comment (kacfengine) [inline package_modules/apt_get]: FYI, I had to add ``rstrip()`` to all of those, because the regex doesn't match newlines, but split will include it.

File: package_modules/yum (hunk 34 above)
Comment (ediosyncratic) [inline package_modules/yum]: OK, as long as a blank line with no space on it counts as a continuation line (i.e. the rule for continuation doesn't requires *at least something* on the line that starts with space to make it valid.

File: package_modules/apt_get (hunk 23 above)
Comment (ediosyncratic) [inline package_modules/apt_get]: Ah, yes - good point.
Then again, you're also probably better off getting rid of any trailing hspace, so this is good ;-)

**Comment by kacfengine** [comment]
Branch is updated.

File: Makefile.am (hunk 1 above)
Comment (ediosyncratic) [inline Makefile.am]: Fixed :-)

File: package_modules/apt_get [hunk 42]
```diff
@@ -0,0 +1,245 @@
...
+
+dpkg_cmd = os.environ.get('CFENGINE_TEST_DPKG_CMD', "/usr/bin/dpkg")
+dpkg_deb_cmd = os.environ.get('CFENGINE_TEST_DPKG_DEB_CMD', "/usr/bin/dpkg-deb")
+dpkg_query_cmd = os.environ.get('CFENGINE_TEST_DPKG_QUERY_CMD', "/usr/bin/dpkg-query")
+
+dpkg_output_format = "Name=${Package}\nVersion=${Version}\nArchitecture=${Architecture}\n"
+dpkg_status_format = "Status=${Status}\n" + dpkg_output_format
+
+apt_get_options = ["-o",
+                   "Dpkg::Options::=--force-confold",
```
ediosyncratic: I would combine these two into a single line, likewise the next to; the option is consumed by the -o, it's not independent of it.

package_modules/apt_get
```diff
@@ -0,0 +1,245 @@
...
+    process = subprocess.Popen([dpkg_query_cmd, "--showformat", dpkg_status_format, "-W"], stdout=subprocess.PIPE)
+    installed_package = False
+    for line in process.stdout:
+        line = line.rstrip("\n")
+        # The space before "installed" is important, because it can be "not-installed".
+        if line.startswith("Status=") and line.split("=", 1)[1].find(" installed") >= 0:
+            installed_package = True
+        elif line.startswith("Status="):
+            installed_package = False
+        elif installed_package:
```
ediosyncratic: Should this be a fresh `if` rather than an `elif` ?
i.e. do we want to also write the Status line to `stdout` ?
It seems to me we should, since you seem to be allowing several distinct statuses that count as installed.

kacfengine: No, the package module API doesn't support the ``Status`` keyword, that's just for our consumption here. The rest of the output will be ``Name``/``Version``/``Architecture``, which is what CFEngine expects.

package_modules/apt_get
```diff
@@ -0,0 +1,245 @@
...
+        # If we have existing versions, operate on those, instead of the
+        # platform default.
+        process = subprocess.Popen([dpkg_query_cmd, "--showformat", "${Architecture}=${Status}\n",
+                                    "-W", name + ":*"], stdout=subprocess.PIPE)
+        for line in process.stdout:
+            # The space before "installed" is important, because it can be "not-installed".
+            if "=" in line:
+                arch, stat = line.split("=", 1)
+                if stat.find(" installed") >= 0:
+                    archs.append(arch)
```
ediosyncratic: For reference, rewriting this as a list comprehension would look like:
```
archs = [arch for arch, stat in (line.split('=', 1) for line in process.stdout if '=' in line) if stat.find('installed) >= 0]
```
Notice that the inner generator (iterating over line) is in simple parentheses; this is just a generator, so we don't gather up its results into a list (or a tuple, as you might have expected (...) to give you), you'll just iterate them in the course of building your list of archs.

I'm used to generators, so find the above reads quite plainly; but I can quite understand that you might prefer the iterative form if you find that clearer ...

package_modules/apt_get
```diff
@@ -0,0 +1,245 @@
...
+
+    version_suffix = ""
+    if version != "" and is_apt_install:
```
ediosyncratic: Use version as a bool instead of comparing to empty.

package_modules/apt_get
ediosyncratic: Fair enough.

package_modules/apt_get
ediosyncratic: Not visibly !  Did you get mixed up between this one and one of the others like it ?

package_modules/apt_get
ediosyncratic:
```
args = [ line.split("=", 1)[1].rstrip() for line in sys.stdin if line.startswith('File=')]
if args:
    return subprocess.call(cmd_line + args, stdout=sys.stderr)
return 0
```
then ...

package_modules/yum
```diff
@@ -0,0 +1,190 @@
...
+import os
+import subprocess
+import re
+
+
+rpm_cmd = os.environ.get('CFENGINE_TEST_RPM_CMD', "/bin/rpm")
+rpm_output_format = "Name=%{name}\nVersion=%{version}\nArchitecture=%{arch}\n"
+
+yum_cmd = os.environ.get('CFENGINE_TEST_YUM_CMD', "/usr/bin/yum")
+yum_options = "--quiet"
```
ediosyncratic: Shouldn't this be a list, `[ '--quiet' ]` for ease of adding to other lists ?
It's called option**s**, plural, after all ...

package_modules/yum
```diff
@@ -0,0 +1,190 @@
...
+
+def get_package_data():
+    pkg_string = ""
+    for line in sys.stdin:
+        if line.startswith("File="):
+            pkg_string = line.split("=", 1)[1].rstrip()
+            break
+
+    if not pkg_string:
+        return 1
```
ediosyncratic:
```
files = (line.split('=', 1)[1].rstrip() for line in sys.stdin if line.startswith('File='))
try: pkg_string = files.next()
except StopIteration: return 1
```
does all of the above.  Note that files is a raw generator (and you'd get a SyntaxError if you missed off the parentheses round it).  When we call its `.next()` we do the first iteration of it, which may iterate many lines of sys.stdin, but it'll stop on the first that starts with 'File=' and give you the rest of the line.  It won't bother with the rest of the iteration and it won't consume the rest of sys.stdin; no need for a break, just don't ask for any remaining values.

The iterator protocol, as used by `for` loops and other things that iterate over a container (or generator) goes: `seq = iter(container)` where `iter` is a builtin and any generator's `iter()` is itself; then it uses `seq.next()` as the value for the loop variable at each iteration, until calling `seq.next()` raises `StopIteration` at which point it gracefully exits the loop.  Any other exception from `.next()` is delivered up the stack as usual.

package_modules/yum
```diff
@@ -0,0 +1,190 @@
...
+
+    repo = False
```
ediosyncratic: The `repo` variable serves no purpose - discard the two places that set it; nothing tests it !

package_modules/yum
ediosyncratic: Note that all your other top-level commands carefully consume all of sys.stdin.
If that isn't necessary, kill their calls to .readlines().
If it necessary, then it is here, too; and you can replace the parentheses for files with square brackets, then use files[0] and catch IndexError instead of using the generator I wrote above.  You'll then be sure to consume all of sys.stdin and you'll quietly discard any excess 'File=' lines.  You could even check for those and grumble about them, if len(files) > 1.

package_modules/yum
```diff
@@ -0,0 +1,190 @@
...
+            # Keep line but strip trailing newline.
+            lastline = line[:-1]
+            continue
+
+        lastline = ""
+        match = re.match("^(?P<name>\S+)\.(?P<arch>[^.\s]+)\s+(?P<version>\S+)\s+\S+\s*$", line)
+        if match != None:
+            sys.stdout.write("Name=" + match.group("name") + "\n")
+            sys.stdout.write("Version=" + match.group("version") + "\n")
+            sys.stdout.write("Architecture=" + match.group("arch") + "\n")
```
ediosyncratic: Concatenating many lines into one, while repeatedly matching a "three words" regex, is more expensive than accumulating the results of line.split() in a words list until its length is three (or more, in which case discard it).

Another way to write this code, assuming you prefer to join the continuation lines, would be to define a generator first,
```
def joinedlines(src):
    src = iter(src) # get the iterator over it, in case it's not already a generator
    joined = src.next()
    for line in src:
        if line[0].isspace(): joined += line
        else:
            yield joined
            joined = line

    if joined: yield joined
```
Because this "function" uses yield, it's automagically a generator and raises StopIteration when it falls off the end.
You can then just iterate `joinedlines(process.stdout)` applying your second regex to each line, without bothering with the first regex check at all.
I'm not sure what the exact semantics of your continuation lines is, though: I've assumed that if a fifty-word line is split up with a space at the start of each continuation, it should be counted as one line.  If it isn't, your code groups up its first three words (and uses them) then groups up the remaining lines into a single long line (the first of which starts with a space so doesn't match either regex) until you hit the next line to start with a non-space, at which point you'll discard this long line and start building a new one.  This leads to whether or not you match depending on the pattern of continuation lines; if I give you
```
one
    two
    three
    four
    ...
    fifty
```
you'll match the initial `one two three` and discard the rest; but if I give you
```
one
    two
    three four
    ...
    fifty
```
you won't match anything; you'll initially build up a line until it has the first four words, then clear lastline and fail to match the line (too many words); after that, you'll accumulate the rest in lastline (starting with a space, so matching neither pattern) until you hit the next line that doesn't start with a blank, when you'll throw away the whole lot and start over.

package_modules/yum
```diff
@@ -0,0 +1,190 @@
...
+        line = lastline + line
+        match = re.match("^\S+\s+\S+\s+\S+", line)
+        if match == None:
+            # Keep line but strip trailing newline.
+            lastline = line[:-1]
+            continue
+
+        lastline = ""
+        match = re.match("^(?P<name>\S+)\.(?P<arch>[^.\s]+)\s+(?P<version>\S+)\s+\S+\s*$", line)
+        if match != None:
```
ediosyncratic: For comparisons to `None`, it's usual to use an `is` test (`None` is a singleton, so every value that is `None` is the same object), `match is None` above and `match is not None` here.  This avoids complications in some situations (where the value being compared to None might have weird comparison semantics, so be equal to it despite not being it) and is more efficient (checking object identity is cheaper than looking for the value's `.__eq__()` or `.__ne__()`  method in order to call it, quite apart from all the fall-back weirdness that then happens if its class doesn't have that method).

package_modules/yum
```diff
@@ -0,0 +1,190 @@
...
+
+            name = line.split("=", 1)[1].rstrip()
+
+        elif line.startswith("Version="):
+            version = line.split("=", 1)[1].rstrip()
+
+        elif line.startswith("Architecture="):
+            arch = line.split("=", 1)[1].rstrip()
+
+    if name:
```
ediosyncratic: Did you mean `if **not** name` here ?
The following code seems to be assuming name is non-empty ... but your present test ensures it's empty !

package_modules/yum
```diff
@@ -0,0 +1,190 @@
...
+    found = False
+    for line in sys.stdin:
+        if line.startswith("File="):
+            found = True
+            cmd_line.append(line.split("=", 1)[1].rstrip())
+
+    if not found:
+        return 0
+
+    return subprocess.call(cmd_line, stdout=sys.stderr)
```
ediosyncratic:
```
args = [line.split("=", 1)[1].rstrip() for line in sys.stdin if line.startswith('File=')]
if args:
    return ubprocess.call(cmd_line + args, stdout=sys.stderr)
return 0
```

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
```diff
@@ -0,0 +1,542 @@
...
+
+from __future__ import print_function
+
+import pprint
+import sys
+
+TEST_COUNT = 0
+PLATFORM_ARCH_32 = False
+
+states = {
```
ediosyncratic: I'd be inclined to generate this programmatically:
```
def states(top=2, archs=("64_bit", "32_bit")):
    """Iterate the possible states.

    Here, state 0 means absent, which won't be included in the state description.
    The other two states are different versions (1 and 2) of a package.
    Each arch can be absent or at either version.
    If both are present, they are constrained to have the same version.
    We iterate pairs of a state-name and a state-description dictionary.
    The dictionary's keys are (arch, version) pairs, mapped to True if
    the given arch is present at the specified version.
    """
    # Empty state - pair each arch with each version (not 0) and map the pair to False:
    blank = dict(((arch, str(ver)), False) for arch in archs for ver in range(1, top + 1))
    # Each conceivable state is expressed as a value for this list:
    vers = [top for v in archs] # as many top as entries in archs
    while True:
        if len(filter(None, set(vers))) < 2: # we have at most one non-zero version
            pairs = [(arch, str(ver)) for arch, ver in zip(archs, vers) if ver]
            yield ('_'.join('_'.joiin(p) for p in pairs) or 'absent',
                   dict((pair, pair in pairs) for pair in blank.items()))

        # Now step vers to the next conceivable (not necessarily valid) candidate:
        ind = len(vers)
        while ind > 0:
            ind -= 1
            if vers[ind] > 0:
                vers[ind] -= 1
                break
            else:
                vers[ind] = top
        else: break # we fell off the bottom, ind < 0 so we have finished.

# Now collect those up in a dictionary mapping state name to its details:
states = dict(states())
```
(Note that the `else` block of a `while` or `for` loop is run precisely if we don't `break` out of the loop.
The function zip() is a transposition: zip(('a', 'b', 'c'), (1, 2, 3)) is (('a', 1), ('b', 2), ('c', 3)) and similar)
Unfortunately, I don't think that's very clear to the reader !

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
```diff
@@ -0,0 +1,542 @@
...
+    # D     = policy => "absent"
+    # F     = File install
+    # R     = Repo install
+    # F64_1 = File to install is 64 bit and version 1
+    # A64   = architecture = "64_bit"
+    # V1    = version => "1"
+    #
+    # This is used to generate the promise structure. It does not include
+    # impossible promises, such as installing a 64-bit file and mentioning
+    # architecture => "32_bit".
```
ediosyncratic: Can this, in its turn, be generated programmatically ?
Doesn't look obviously like it can; but that may just be because I have no idea what it says !

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
ediosyncratic: Fixed - thanks.

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
```diff
@@ -0,0 +1,542 @@
...
+                if version == from_32:
+                    to_32 = "0"
+            else:
+                to_64 = "0"
+                to_32 = "0"
+
+    return to_64, to_32
+
+
+# Calculate all possible transitions from from_state, to to_state.
```
ediosyncratic: Doubling 'from' and 'to' is ugly.  Might be clearer as
`# Calculate all possible transitions from_state -> to_state`

package_modules/yum
kacfengine: > Note that all your other top-level commands carefully consume all of sys.stdin.
> If it necessary, then it is here, too

Indeed it is, good catch!

kacfengine: Fixed by removing ``break``. I kept the existing structure, I think it's simpler.

package_modules/yum
kacfengine: Fixed!

package_modules/yum
kacfengine: I don't really care (much) about execution speed, only readability. *One* call to apt-get will already increase running time by orders of magnitude compared to what we do here.

package_modules/yum
kacfengine: Following my RedHat fixes (which I shall push soon), this section is completely rearranged.

tests/acceptance/17_packages/10_new/unsafe/the_great_package_test_generator.py
kacfengine: Both this and the other case you mention probably can, but this is just test code, and I'd rather keep it simple for the reader: Everyone can see what's going on this way. With luck we'll never have to run this script again.

kacfengine:
Superseded by #472.

PR #466, opened 2015-05-21
Change: Local software dir definition

nickanderson:
HPUX $(sys.flavour) includes a version that does not align with other software
update directories. Switch hpux to use $(sys.class)_$(sys.arch) and refactor
all platforms to use a common variable for the package dir for good hygiene.

### DIFF
--- a/cfe_internal/update/update_bins.cf
+++ b/cfe_internal/update/update_bins.cf
@@ -31,7 +31,21 @@ bundle agent cfe_internal_update_bins
       comment => "The Cfengine binary updates directory on the policy host",
       handle => "cfe_internal_update_bins_vars_master_software_location";
 
-      "local_software_dir"        string => translatepath("$(sys.workdir)/software_updates/$(sys.flavour)_$(sys.arch)"),
+    !hpux::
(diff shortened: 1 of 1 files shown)


cfe_internal/update/update_bins.cf
```diff
@@ -68,18 +82,6 @@ bundle agent cfe_internal_update_bins
...
-      "stop_signal"              string => "kill",
-      comment => "Define a stop signal for Windows",
-      handle => "cfe_internal_update_bins_vars_stop_signal_windows";
-
-    !windows.enterprise::
-
-      "stop_signal"              string => "term",
-      comment => "Define a stop signal for *NIX system",
-      handle => "cfe_internal_update_bins_vars_stop_signal_unix";
-
```
kacfengine: This looks like it doesn't belong here, but I take it that ``stop_signal`` is not used anywhere. This should be made into a separate commit in case we need to revert either fix later.

kacfengine:
Looks good otherwise!

estenberg:
Agree that stop_signal is unrelated, but will let that pass.

PR #469, opened 2015-05-22
Remove log rotation for cf_notkept.log and cf_repaierd.log from 3.7 l…

maciejmrowiec:
…ib as their generation is removed.

### DIFF
--- a/controls/3.7/def.cf
+++ b/controls/3.7/def.cf
@@ -180,8 +180,6 @@ bundle common def
 
       "enterprise_log_files" slist =>
       {
-        "$(sys.workdir)/cf_notkept.log",
-        "$(sys.workdir)/cf_repair.log",
         "$(sys.workdir)/state/cf_value.log",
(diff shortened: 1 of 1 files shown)


nickanderson:
They are not generated at all any longer? I thought it was tunable.

Should we still tidy up those old log files though? They will be there after an upgrade form 3.6 to 3.7.

maciejmrowiec:
There are PR's in core/ent to remove those.
RIght, forgot about the leftovers :/:/ If we ignore this, will be get notkept every time it want's to do some cleanup? (how often?)

nickanderson:
Is it really not kept if the file isn't found?

Its perfectly fine to leave leftovers, but I think we should document the fact they are left overs. Can you add this in the CHANGELOG rotation for these files was removed, it is safe to purge the left over files if desired.

maciejmrowiec:
tested, it's: kept
I'd rather do not keep unmanaged leftovers so I'll keep that cleanup for now.

PR #471, opened 2015-05-27
Add new package promise to be default method for software inventory g…

pasinskim:
…eneration.

From 3.7 new package promise way of generating software inventory
should be default one.

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -555,13 +555,17 @@ bundle agent cfe_autorun_inventory_packages
                                  "0");
 
   packages:
-    debian::
+    # From 3.7 onwards there is a new package promise implementation using package
+    # modules, in which case installing non_existing_package is not needed any more.
+    (cfengine_3_5|cfengine_3_6).debian::
       "cfe_internal_non_existing_package"
       package_policy => "add",
       package_method => inventory_apt_get($(refresh)),
       action => if_elapsed_day;
 
-    redhat::
+    # From 3.7 onwards there is a new package promise implementation using package
+    # modules, in which case installing non_existing_package is not needed any more.
+    (cfengine_3_5|cfengine_3_6).redhat::
       "cfe_internal_non_existing_package"
       package_policy => "add",
       package_method => inventory_yum_rpm($(refresh)),
--- a/promises.cf
+++ b/promises.cf
@@ -57,14 +57,22 @@ body common control
 
       version => "CFEngine Promises.cf 3.7.0";
 
+      # From 3.7 onwards there is a new package promise implementation using package
+      # modules in which you MUST provide package modules used to generate
+      # software inventory reports. You can also provide global default package module 
+      # instead of specifying it in all package promises.
+      # Please also note that software inventory reports is CFEngine enterprise only
+      # feature and core users can remove package_inventory line.
+      #debian|redhat::
+      #    package_inventory => { $(package_module_knowledge.platform_default) };
+      #    package_module => $(package_module_knowledge.platform_default);
+
       # Uncomment to connect to the hub using latest protocol.
       #protocol_version => "latest";
 
       #   goal_categories => { "goals", "targets", "milestones" };
       #   goal_patterns   => { "goal_.*", "target.*","milestone.*" };
 }
-
-      #
 
 bundle common inventory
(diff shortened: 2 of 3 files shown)


kacfengine:
Depends on both https://github.com/cfengine/core/pull/2215 and #465.

inventory/any.cf
```diff
@@ -555,13 +555,17 @@ bundle agent cfe_autorun_inventory_packages
                                  "0");
 
   packages:
-    debian::
+    # In 3.7 we are usining package_module specific way of generating software inventory.
+    # In this case installing non_existing_package is not needed any more.
+    !cfengine_3_7.debian::
```
kacfengine: This should be inclusive rather than exclusive, in order to be future proof. I would also rephrase the sentence a bit. Maybe:

```
# From 3.7 onwards there is a new package promise implementation using package
# modules, in which case installing non_existing_package is not needed any more.
(cfengine_3_5|cfengine_3_6).debian::
```

And similar for RedHat.

pasinskim: Makes a lot of sense. Fixed!

promises.cf
```diff
@@ -57,6 +57,10 @@ body common control
 
       version => "CFEngine Promises.cf 3.7.0";
 
+      cfengine_3_7.(debian||redhat)::
```
estenberg: The most common way to do OR is | and not ||

estenberg:
Cool!

promises.cf
kacfengine: Indeed, I think ``||`` won't work actually, it will think you are ORing with a class called literally ``|redhat``.

promises.cf
```diff
@@ -57,6 +57,10 @@ body common control
...
+          package_inventory => { $(package_module_knowledge.platform_default) };
+          package_module => $(package_module_knowledge.platform_default);
```
kacfengine: Hmm, here's a problem: This will break 3.5 and 3.6. I thought we had handled this in the spec, but this apparently got left over. I don't immediately see how this can be solved, since splitting ``promises.cf`` is infeasible, and quite undesirable, regardless. Suggestions?

nickanderson: Do these have sensible defaults inside of the C code? They could be hash commented out with a note saying the syntax is incompatible prior to 3.7. 

Splitting promises.cf is possible, but really only from execd where you could choose a different promsies.cf file. You would still end up out of luck for the actual promsies.cf file. I guess promises.cf would be the 3.6 compat version, and you would have to have a different one for 3.7. Very nasty. I'd rather see a commented block.

kacfengine: Agree, that's very nasty, and it is the first file everyone who tries CFEngine will see, I really don't think we should split it. The problem is that even if the section is commented, you cannot enable it as long as you have any 3.6 hosts. In the transition phase I assume this will be quite common.

kacfengine: > Do these have sensible defaults inside of the C code?

Not ATM, the default is nothing. We could add it, but it's still not ideal. It means that no one can add a package inventory module besides the two standard ones as long as they have 3.6 hosts.

nickanderson: The only way in 3.7 that I can think to support setting them is having promises.cf and promsie_3.7.cf where the correct version is chosen in cf-execd.

In 3.9 we could use the IF cf_version thing. But that's not available in 3.6 or 3.5 (we try to maintain 2 version back compatibility).

I believe we had a similar situation with protocol_version in 3.6. You could not enable the setting until all your hosts were on 3.6 because that control option was not valid in 3.5. So probably a commented out block is the best and least complex thing to do.

kacfengine: Yes, I suppose you're right. I've updated the task https://dev.cfengine.com/issues/7210.

promises.cf
kacfengine: Additionally, the CFEngine version selection should be exclusive rather than inclusive here as well, just like in `` inventory/any.cf``.

pasinskim:
@nick @kacfengine @estenberg thanks for reviews. I have changed this stuff accordingly.

kacfengine:
Merged manually after adding the same change for the ``lib/3.8`` directory, and cherry-picked to 3.7.x.

PR #472, opened 2015-05-29
Redmine #7121 and #7122: Add apt_get and yum package modules.

kacfengine:
Tests will be in a separate commit.

This is a subset of https://github.com/cfengine/masterfiles/pull/465 which contains only the modules, not the tests. Work on making the tests pass will continue, but this will allow us to start building packages, and at least the Debian package module is already in a quite good state.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,8 +1,10 @@
-if HAVE_CORE
-    TESTS_SUBDIRS = tests/acceptance
-endif
+SUBDIRS = \
+	modules/packages
+	tests/unit
 
-SUBDIRS = $(TESTS_SUBDIRS)
+if HAVE_CORE
+    SUBDIRS += tests/acceptance
+endif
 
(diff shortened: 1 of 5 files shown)


cdituri:
@kacfengine - First off, +1 to all involved to the various facets of the new package promise implementation. Awesome.

Second, what's your thought on abstracting some of the python away with classes and decorators? Seems a good portion of these modules will be quite similar.

kacfengine:
> First off, +1 to all involved to the various facets of the new package promise implementation. Awesome.

Thank you, it was due!

> what's your thought on abstracting some of the python away with classes and decorators? Seems a good portion of these modules will be quite similar.

Possibly, but although the modules may appear superficially similar, there are a lot of small differences; the devil's in the details, so it may not be as easy as it sounds. But I'm happy to be proven wrong on this!

PR #474, opened 2015-06-03
Fix broken or.

pasinskim:
Use | instead of ||.


pasinskim:
@kacfengine I assume following up https://github.com/cfengine/masterfiles/pull/471 this is also wrong?!?

kacfengine:
Indeed.

kacfengine:
Cherry-picked to 3.7.x.

PR #476, opened 2015-06-09
changes reporting maintenance policies

maciejmrowiec:


maciejmrowiec:
this also needs cherry-pick into 3.7.x

lib/3.7/cfe_internal.cf
```diff
@@ -211,26 +223,72 @@ bundle agent cfe_internal_cleanup_agent_reports
...
+{
+  vars:
+    "promise_outcome"
+    slist => {"REPAIRED", "NOTKEPT"};
+
+    "query_create_promise_log_$(promise_outcome)"
+    string => "SELECT promise_log_partition_create(NOW() - INTERVAL '7 day', 7 + 3, '$(promise_outcome)');";
+
+  commands:
+      # "$(sys.bindir)/psql cfdb -c \"$(query_create_promise_log_$(promise_outcome))\""
```
estenberg: I believe this comment should replace the line below? We need to use full path to the binary in case they have another postgres installation.

lib/3.7/cfe_internal.cf
```diff
@@ -211,26 +223,72 @@ bundle agent cfe_internal_cleanup_agent_reports
...
+      "psql cfdb -c \"$(query_create_promise_log_$(promise_outcome))\""
+      contain => in_shell_and_silent,
```
estenberg: Do we have to run it in shell? If not, please use contain => silent instead.

lib/3.7/cfe_internal.cf
```diff
@@ -211,26 +223,72 @@ bundle agent cfe_internal_cleanup_agent_reports
...
+  vars:
+    "cleanup_query_repaired"
+    string => "SELECT promise_log_partition_cleanup('REPAIRED', '$(history_length_days) day');";
+
+    "cleanup_query_notkept"
+    string => "SELECT promise_log_partition_cleanup('NOTKEPT', '1 day');";
+
+  commands:
+      "$(sys.bindir)/psql cfdb -c \"$(cleanup_query_repaired)\""
+      contain => in_shell_and_silent,
```
estenberg: Do we have to run it in shell? If not, please use contain => silent instead.

lib/3.7/cfe_internal.cf
```diff
@@ -211,26 +223,72 @@ bundle agent cfe_internal_cleanup_agent_reports
...
+      classes => kept_successful_command,
+      handle => "cf_database_maintain_promise_log_repaired";
+
+      "$(sys.bindir)/psql cfdb -c \"$(cleanup_query_notkept)\""
+      contain => in_shell_and_silent,
```
estenberg: Do we have to run it in shell? If not, please use contain => silent instead.

estenberg:
I think we should remove usage of the kept_successful_command body. Isn't this just run once a day?

It's nice to be able to see that the job actually ran, and this is a change to the system after all..

maciejmrowiec:
thanks for the fast response,
updated

estenberg:
Thanks for updating.

It is still not 100% clear to me how frequently we will get promise repairs out of this.
A repair or two daily would be great for logging, but if it generates several repairs every run on the hub we should adjust.

Something we will easily see/adjust when we do scale testing and watch changes reporting. :)

estenberg:
Cherry-picked to 3.7.x.

PR #483, opened 2015-06-15
Add support for Debian 6 and Ubuntu 10

pasinskim:
Some apt_get modifications supporting Ubuntu 10 and Debain 6.


kacfengine:
Cherry-picked to 3.7.x as well.

PR #485, opened 2015-06-18
Add Debian 7 full support for new package promise.

pasinskim:


pasinskim:
@kacfengine this should fix all deb based distros we are supporting.

modules/packages/apt_get
```diff
@@ -238,8 +244,13 @@ def repo_install():
 
     args = package_arguments_builder(True)
 
-    if args:
-        return subprocess_call(cmd_line + args, stdout=sys.stderr)
+    if (not args):
+        return 0
+
+    a1, a2 =  map(list, zip(*args))
```
kacfengine: I think this could use a comment to explain what it does.

modules/packages/apt_get
```diff
@@ -198,9 +197,16 @@ def one_package_argument(name, arch, version, is_apt_install):
...
     if archs:
-        args = [name + ':' + arch + version_suffix for arch in archs]
+        for cur_arch in archs:
+            if cur_arch == platform_arch:
+                if is_apt_install:
+                    args.append((name + version_suffix, name + version_suffix))
+                else:
+                    args.append((name + version_suffix, name + ':' + cur_arch + version_suffix))
+            else:
+                args.append((name + ':' + cur_arch + version_suffix, name + ':' + cur_arch + version_suffix))
```
kacfengine: A comment would be useful here as well, to explain why we add the two different variants.

kacfengine:
Very nice way of solving it though!

pasinskim:
@kacfengine I have added some comments to explain what is happening there...

kacfengine:
Cherry-picked to 3.7.x as well.

PR #486, opened 2015-06-19
Adapt yum package module to RedHat 6.

kacfengine:


kacfengine:
@pasinskim: FYI, my latest changes for the yum package module.

PR #488, opened 2015-06-19
Correctly detect if systemd is used in Debian

brontolinux:
CFEngine 3.6 tries to understand if a Linux is using systemd as init system
by looking at the contents of /proc/1/cmdline, that happens in
bundle common inventory_linux. That's indeed a smart thing to do but unfortunately
fails on Debian Jessie, where you have:

root@cf-test-v10:~# ls -l /sbin/init
lrwxrwxrwx 1 root root 20 May 26 06:07 /sbin/init -> /lib/systemd/systemd

the pseudo-file in /proc will still report /sbin/init and as a result the systemd
class won't be set. This affects services promises negatively

This patch checks if /sbin/init is a link; if it is, and it points to systemd,
the systemd class is set

### DIFF
--- a/inventory/debian.cf
+++ b/inventory/debian.cf
@@ -23,6 +23,11 @@ bundle common inventory_debian
       "mint_release" string  => "$(mint_info[RELEASE][1])" ;
       "mint_codename" string => "$(mint_info[CODENAME][1])" ;
 
+    init_is_link::
+      "init_link_destination"
+          string => filestat("/sbin/init","linktarget") ;
+
+
   classes:
     any::
       "debian_derived_evaluated"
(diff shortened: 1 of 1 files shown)


inventory/debian.cf
```diff
@@ -82,6 +87,12 @@ bundle common inventory_debian
       comment => "derived from Debian",
       meta => { "inventory", "attribute_name=none" };
 
+    init_is_link::
+      "systemd"
+          expression => regcmp("/lib/systemd/systemd",
```
kacfengine: Perhaps this should be ``.*/systemd$`` instead, to catch all possible paths where systemd might be installed?

kacfengine:
+1!

Note that the same thing exists in C code (https://github.com/cfengine/core/blob/master/libenv/sysinfo.c#L993-L1012).

Ticket: https://dev.cfengine.com/issues/7297

kacfengine:
See also https://github.com/cfengine/core/pull/2249.

brontolinux:
No problem in matching the final part of the path. Currently it's /lib/systemd/systemd in jessie but a pattern matching is probably more future-proof.

kacfengine:
In fact I don't think this is necessary anymore after https://github.com/cfengine/core/pull/2252. Closing this to avoid having the detection in two places. Thanks for the Proof of Concept though! :-)

PR #492, opened 2015-06-24
Add: Negative class guard protecting agent email settings

nickanderson:
This allows agent email to be easily disabled by augmenting def.json.

Ref: https://dev.cfengine.com/issues/7312

### DIFF
--- a/controls/3.7/def.cf
+++ b/controls/3.7/def.cf
@@ -236,8 +236,9 @@ bundle common def
       # Internal CFEngine log files rotation
       "cfengine_internal_rotate_logs" expression => "any";
 
-      # Enable agent email output (also see mailto, mailfrom and smtpserver)
+      # Enable or disable agent email output (also see mailto, mailfrom and smtpserver)
       "cfengine_internal_agent_email" expression => "any";
+      "cfengine_internal_disable_agent_email" expression => "!any";
 
       # Transfer policies and binaries with encryption
       # you can also request it from the command line with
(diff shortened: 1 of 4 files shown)


controls/3.7/def.cf
```diff
@@ -236,8 +236,9 @@ bundle common def
       # Internal CFEngine log files rotation
       "cfengine_internal_rotate_logs" expression => "any";
 
-      # Enable agent email output (also see mailto, mailfrom and smtpserver)
+      # Enable or disable agent email output (also see mailto, mailfrom and smtpserver)
       "cfengine_internal_agent_email" expression => "any";
+      "cfengine_internal_disable_agent_email" expression => "!any";
```
kacfengine: If people are not expected to edit ``def.cf``, can't we just kill ``cfengine_internal_agent_email`` altogether? Or do we need it to be around for people with existing ``def.cf``?

kacfengine:
3.7 version: https://github.com/cfengine/masterfiles/pull/493

controls/3.7/def.cf
nickanderson: I actually prefer to have both a positive and a negative class to control things. 

We should keep it at least for people with existing def.cf. If we really want to remove it then we can add some reports based on the usage of the class to warn people that it will be deprecated.

tzz: It really complicates things for beginners to have multiple ways to break things...

nickanderson: Ya, but if we have a class guard that is defaulted on, its not easy to override that in something like def.json. If things are guarded by enable.!disable we can turn things on by default if we want, and users can still easily disable them with an override class.

nickanderson:
trigger build

nickanderson:
trigger build

PR #499, opened 2015-06-24
Change: Inform guards to DEBUG

nickanderson:
Inform mode is supposed to inform you of changes to the system, debug is more
appropriate for these informational messages. Also not using the same class as
defined by the log level options to the agent allows us to not be overwhelemed
with output when doing policy debugging.

Ref: https://dev.cfengine.com/issues/7318
(cherry picked from commit e4d4b9daf3ea26c32e9dafca1bf49527a2a2738f)


nickanderson:
trigger build

PR #500, opened 2015-06-26
Add: Policy to configure external watchdog for cf-execd

nickanderson:
Ref: https://dev.cfengine.com/issues/7324

### DIFF
--- a/cfe_internal/core/watchdog/watchdog.cf
+++ b/cfe_internal/core/watchdog/watchdog.cf
@@ -1,0 +1,66 @@
+bundle agent cfe_internal_core_watchdog(state)
+# @breif Configure external watchdog processes to keep cf-execd running
+# @param state (enabled|disabled) The state to keep the watchdog configuration in
+{
+  meta:
+    "description"
+      string => "Configure external watchdog processes (like cron, or monit) to
+                 make sure that cf-execd is always running";
+
+  classes:
+    any::
+
+      "have_cron_d"
+        expression => isdir("/etc/cron.d");
+
+      "invalid_state"
+        not => regcmp("(enabled|disabled)", "$(state)");
+
+    !invalid_state::
+
+      # Define a class for whatever the desired state is so long as the state is
+      # valid
+
+      "$(state)"
+        expression => "any";
+
+  vars:
+    any::
+
+      "cf_execd"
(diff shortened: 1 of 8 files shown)


nickanderson:
@estenberg @tzz @kacfengine thoughts?

kacfengine:
A couple of things:

* I think you should make this dependent on the existence of pgrep as well, since Linux is the only one that has it.
* This is new functionality, I think this should go into 3.8 only.

cfe_internal/core/watchdog/watchdog.cf
```diff
@@ -0,0 +1,66 @@
...
+
+      "$(state)"
+        expression => "any";
+
+  vars:
+    any::
+
+      "cf_execd"
+        string => ifelse("windows", translatepath('"$(sys.bindir)/cf-execd.exe)"'),
+                         "$(sys.bindir)/cf-execd");
```
kacfengine: There is the ``$(sys.cf_execd)`` variable, which saves you from doing this.

cfe_internal/core/watchdog/watchdog.cf
```diff
@@ -0,0 +1,66 @@
...
+
+  files:
+
+    enabled.have_cron_d::
+
+      "$(cron_d_watchdog)"
+        create => "true";
+
+      "$(cron_d_watchdog)"
+        edit_defaults => empty,
```
kacfengine: Not needed for ``edit_template``, all templates start with an empty file. Only ``edit_line`` uses ``edit_defaults``.

kacfengine:
Looks useful though!

tzz:
Looks reasonable.  It should complain loudly if `/etc/cron.d` doesn't exist.

cfe_internal/core/watchdog/watchdog.cf
nickanderson: thanks

cfe_internal/core/watchdog/watchdog.cf
nickanderson: ack

kacfengine:
New one with slightly tweaked Changelog in #516.

kacfengine:
Merged #516.

PR #506, opened 2015-06-30
Reorganize postgres maintenance policies.

maciejmrowiec:
Scope monitoring maintenance only to 3.6.0 and 3.6.1 (postgresql)
Daily vacuum job (3.6 3.7 3.8)
Cleanups.
Add measure_promise_time action body.
Enable timing of maintenance promises.


maciejmrowiec:
I need to test first so don't merge yet ;)
I need comments about structure and overall.

CHANGELOG.md
```diff
@@ -31,6 +33,8 @@ Notable changes to the framework should be documented here
    bundle being activated by default if there is no bundlesequence specified.
  - Reduce benchmarks sample history to 1 day.
  - Update policy no longer generates a keypair if one is not found. (Redmine: #7167)
+ - Relocate cfe_internal_postgresql_maintenance bundle to lib/VER/
+ - Set postgresql_monitoring_maintenance only for versions 3.6.0 and 3.6.1
```
nickanderson: Thanks for the changelog updates :)

lib/3.6/cfe_internal.cf
```diff
@@ -234,3 +234,63 @@ bundle agent cfe_internal_cleanup_agent_reports
         delete => tidy,
         handle => "cf_cleanup_agent_reports_$(report_files)";
 }
+
```
nickanderson: Can we put this into a new file so that we can load it only for the hub?

cfe_internal_enterprise_hub.cf perhaps?
This way we can load it only for the enterprise hub. (will have to be separate work to do that but moving this there would make that a bit easier).
You can simply add the new file to `inputs` in `bundle common stdlib` which is found in `lib/{3.6,3.7}/stdlib.cf`. For 3.5 you will need to add the file to `inputs` under the `cfengine_3_5` class guard in `bundle common cfengine_stdlib` in `promises.cf`

nickanderson:
Other than moving that one bundle to an enterprise_hub specific file the structure seems good

controls/3.6/def.cf
```diff
@@ -278,11 +278,15 @@ bundle common def
...
       # pre-defined to every Sunday at 2 a.m.
       # This can be changed later on.
-
       "postgresql_full_maintenance" expression => "postgresql_maintenance_supported.Sunday.Hr02.Min00_05";
 
-      # Vacuum monitoring_mg table every day (except Sunday, when vacuum is run on entire db)
-      "postgresql_monitoring_maintenance" expression => "postgresql_maintenance_supported.!Sunday.Hr02.Min00_05";
+      # Vacuum full monitoring_mg table every day (except Sunday, when vacuum full is run on entire db)
+      # Required only by CFEngine Enterprise Hub 3.6.0 and 3.6.1
+      "postgresql_monitoring_maintenance" expression => "postgresql_maintenance_supported.!Sunday.Hr02.Min00_05.(nova_3_6_0|nova_3_6_1)";
```
nickanderson: (enterprise_edition.(cfengine_3_6_0|cfengine_3_6_1)) instead of nova? If the semantics of nova/community change I think enterprise_edition will be an easier thing to replace.

lib/3.6/cfe_internal.cf
maciejmrowiec: most of the file (except one bundle) is hub specific

maciejmrowiec:
I moved hub specific bundles into cfe_internal_hub.cf but I still need to fix a bug:

getting error: Undefined bundle cfe_internal_hub_maintain with type usebundle
from the clients

maciejmrowiec:
I think that this would only worked if I would reorganise cfe_insternal/enterprise/main.cf also in some kind of conditional loading but nut sure if this is worth ...

nickanderson:
Right, just include the hub only one on all enterprise clients that should fix the remote agents. Later I can refactor and fix it so that its only loaded by the hub without causing errors on the clients

maciejmrowiec:
Ok, I left only the split between hub and common into separate files.

maciejmrowiec:
this needs back port to 3.7 and 3.6 should I do it ?

nickanderson:
yes, and we need to make sure that 3.5 doesn't blow up with it as well.

nickanderson:
trigger build

maciejmrowiec:
3.5 passes cf-promises

nickanderson:
Ready to merge?

controls/3.6/def.cf
```diff
@@ -278,11 +278,15 @@ bundle common def
...
       # pre-defined to every Sunday at 2 a.m.
       # This can be changed later on.
-
       "postgresql_full_maintenance" expression => "postgresql_maintenance_supported.Sunday.Hr02.Min00_05";
 
-      # Vacuum monitoring_mg table every day (except Sunday, when vacuum is run on entire db)
-      "postgresql_monitoring_maintenance" expression => "postgresql_maintenance_supported.!Sunday.Hr02.Min00_05";
+      # Vacuum full monitoring_mg table every day (except Sunday, when vacuum full is run on entire db)
+      # Required only by CFEngine Enterprise Hub 3.6.0 and 3.6.1
+      "postgresql_monitoring_maintenance" expression => "postgresql_maintenance_supported.!Sunday.Hr02.Min00_05.(cfengine_3_6_0|cfengine_3_6_1)";
```
nickanderson: I just noticed these class names changed. Can you include that in the changelog please?

maciejmrowiec:
Nothing more from my side :)

controls/3.6/def.cf
maciejmrowiec: It's mentioned in the changelog: scoping down monitoring maintenance

PR #511, opened 2015-07-02
Add: bundle agent sysctl

atsaloli:
This is a bundle to set and load sysctl settings (a common sysadmin task).

It's a wrapper around set_variable_values.  It edits /etc/sysctl.conf and then runs sysctl -p to load the values.

### DIFF
--- a/lib/3.6/bundles.cf
+++ b/lib/3.6/bundles.cf
@@ -416,3 +416,51 @@ bundle agent run_ifdefined(namespace, mybundle)
     verbose_mode::
       "$(this.bundle): found matching bundles $(bundlesfound) for namespace '$(namespace)' and bundle '$(mybundle)'";
 }
+bundle agent sysctl(settings_array)
+# @brief bundle to set and load sysctl settings
+# @param name of array containing sysctl settings
+#
+# This bundle is a wrapper around set_variable_values 
+# targetting /etc/sysctl.conf. It runs sysctl -p after
+# editing sysctl.conf, to load the new values.
+# Contributed by Aleksey Tsalolikhin, 2 July 2015
+#
+# **Example:**
+#
+# ```cf3
+# bundle agent example
+# {
+#   vars:
+#       "sysctl_settings[vm.swappiness]"         string => "4";
+#       "sysctl_settings[net.core.rmem_max]"     string => "4194304";
+#       "sysctl_settings[net.core.rmem_default]" string => "4194304";
+#
+#   methods:
+#       "Edit and reload /etc/sysctl.conf" usebundle => sysctl("$(this.bundle).sysctl_settings");
+# }
+#
+# ```
+{
+  vars:
+      "sysctl_conf" string => "/etc/sysctl.conf";
+
+    DEBUG|verbose_mode::
+      "keys" slist => getindices("$(settings_array)");
+
+  files:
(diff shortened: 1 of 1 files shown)


lib/3.6/bundles.cf
```diff
@@ -416,3 +416,51 @@ bundle agent run_ifdefined(namespace, mybundle)
...
+        edit_defaults => backup_timestamp,
+        classes       => if_repaired("reinit_sysctl");
+
+  commands:
+    reinit_sysctl::
+      "$(paths.path[sysctl]) -p $(sysctl_conf)";
+
+  reports:
+    DEBUG|verbose_mode::
+      "array name = $(settings_array)";
```
nickanderson: Can you please use DEBUG|DEBUG_sysctl:: and can you prefix the reports with "DEBUG $(this.bundle):" please

nickanderson:
Does this also work with data containers?

atsaloli:
I've added DEBUG_sysctl class, DEBUG reporting prefix complete with $(this.bundle) name, and verified that this works with data containers.  Thanks for keeping quality high, Nick.

lib/3.6/bundles.cf
```diff
@@ -416,3 +416,54 @@ bundle agent run_ifdefined(namespace, mybundle)
     verbose_mode::
       "$(this.bundle): found matching bundles $(bundlesfound) for namespace '$(namespace)' and bundle '$(mybundle)'";
 }
+bundle agent sysctl(settings_array)
+# @brief bundle to set and load sysctl settings
+# @param name of array containing sysctl settings
```
nickanderson: after @param should be "settings_array" without the quotes.

nickanderson:
I also noticed this is only against lib/3.6/bundles
We need it in 3.7 and 3.8 as well.

lib/3.6/bundles.cf
```diff
@@ -416,3 +416,54 @@ bundle agent run_ifdefined(namespace, mybundle)
...
+      "sysctl_conf" string => "/etc/sysctl.conf";
+
+    DEBUG|DEBUG_sysctl|verbose_mode::
+      "keys" slist => getindices("$(settings_array)");
+
+  files:
+      "$(sysctl_conf)"
+        edit_line     => set_variable_values("$(settings_array)"),
+        edit_defaults => backup_timestamp,
+        classes       => if_repaired("reinit_sysctl");
```
nickanderson: if_repaired produces a globaly scoped class. Could you please use a classes body that defines a bundle scoped class? I don't see any reason to signal outside this bundle.

Perphas: 

classes => scoped_classes_generic("bundle", "sysctl_conf")

And then to guard the reload command:

sysctl_conf_repaired::

lib/3.6/bundles.cf
```diff
@@ -416,3 +416,54 @@ bundle agent run_ifdefined(namespace, mybundle)
     verbose_mode::
       "$(this.bundle): found matching bundles $(bundlesfound) for namespace '$(namespace)' and bundle '$(mybundle)'";
 }
+bundle agent sysctl(settings_array)
```
nickanderson: I think we might be very likely to conflict with someone elses bundle here. Perhaps call the bundle cfe_stdlib_sysctl

PR #516, opened 2015-07-06
Add: Policy to configure external watchdog for cf-execd

kacfengine:
Ref: https://dev.cfengine.com/issues/7324


kacfengine:
trigger build

PR #519, opened 2015-07-14
fix process_result logic to match the purpose of days_older_than(). s…

soumyadipdm:
Fix process_result logic to match the purpose of days_older_than()
see: https://dev.cfengine.com/issues/3009


nickanderson:
Thanks @soumyadipdm I have confirmed its incorrectly selecting processes. In a few hours after I confirm it properly selects with your patch I will merge it.

PR #524, opened 2015-07-16
Fix: Multiple execd processes after upgrade

nickanderson:
cf-execd does not like being killed. Up through cfengine 3.6.5 it launches
itself again if it is killed. This addresses the situation for cf-upgrade
initiated upgrades by spinning a tight loop killing cf-agent and cf-execd
processes before finally restarting the service.

### DIFF
--- a/cfe_internal/update/update_bins.cf
+++ b/cfe_internal/update/update_bins.cf
@@ -145,6 +145,14 @@ bundle agent cfe_internal_update_bins
 
       "have_software_dir" expression => fileexists($(local_software_dir));
 
+
+      "found_cf_upgrade_process_running"
+        expression => regline(".*cf-upgrade.*", "$(sys.workdir)/state/cf_procs");
+
+  reports:
+    found_cf_upgrade_process_running::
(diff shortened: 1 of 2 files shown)


cfe_internal/update/update_bins.cf
```diff
@@ -391,9 +414,14 @@ bundle edit_line u_install_script
...
 /usr/sbin/pkgadd -n -a $(cfe_internal_update_bins.admin_file) -d $(const.dollar)1 all
-$(sys.workdir)/bin/cf-execd || true
-exit 0";
-
+while /usr/bin/pgrep cf-agent || /usr/bin/pgrep cf-execd; do
+echo Maniacally trying to kill all cf-execd, and cf-agent processes after package upgrade to ensure only one copy of the new binaries is running.
+/usr/bin/pkill cf-agent
+/usr/bin/pkill cf-execd
+/etc/init.d/cfengine3 stop
+done
```
jimis: I'd add "sleep 1" before "done" to give some time to the process to actually die.

jimis: Maybe also a counter up to 30 to ensure that we don't loop infinitely

jimis:
+1 for looking into the details! :-)

PR #525, opened 2015-07-16
Change: Enable limit_robot_agents maintenance policy

nickanderson:
This policy was originally disabled because it killed execds running inside of
containers. However it was keeping multiple cf-execds in check especially
after agent upgrades.

This needs picked to 3.7 and 3.6


estenberg:
Picked to 3.7.x, could you make a new pull for 3.6.x? It's not a simple backport there.

PR #527, opened 2015-07-16
Change: Manually backport enablement of limit_robot_agents

nickanderson:
This policy was originally disabled because it killed execds running inside of 
containers. However it was keeping multiple cf-execds in check especially 
after agent upgrades.


estenberg:
Thanks!

PR #529, opened 2015-07-22
rework cfe_autorun_inventory_ipv4_addresses.ipv4 and make the value the interface

tzz:
Note comments, especially that by using `sys.interfaces` we exclude the loopback automatically.

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -83,20 +83,20 @@ bundle agent cfe_autorun_inventory_ipv4_addresses
 # This will filter the loopback address (127.0.0.1, as it is likely not very interesting)
 {
   vars:
-    "filter_reg"
-      string => "127\.0\.0\.1",
-      comment => "Addresses that match this regular expression will be filtered
+      "filter_reg"
+        string => "127\.0\.0\.1",
+        comment => "Addresses that match this regular expression will be filtered
                   from the inventory for ipv4 addresses";
 
-    # Strings are displayed more beautifully in Mission Portal than lists, so
-    # we first generate the list of addresses to be inventoried and then do
-    # inventory using an array.
-    "ipv4_addresses"
-      slist => filter( $(filter_reg), "sys.ip_addresses", "true", "true", 999999999);
+      # Strings are displayed more beautifully in Mission Portal than lists, so
+      # we first generate the list of addresses to be inventoried and then do
+      # inventory using an array.
+      "ipv4_addresses"
+        slist => filter( $(filter_reg), "sys.ip_addresses", "true", "true", 999999999);
 
-    "ipv4[$(ipv4_addresses)]"
-      string => "$(ipv4_addresses)",
-      meta => { "inventory", "attribute_name=IPv4 addresses" };
+      "ipv4[$(ipv4_addresses)]"
+        string => "$(ipv4_addresses)",
+        meta => { "inventory", "attribute_name=IPv4 addresses" };
 
   reports:
     DEBUG|DEBUG_cfe_autorun_inventory_ipv4_addresses::
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -79,29 +79,26 @@ bundle agent cfe_autorun_inventory_listening_ports
 }
 
 bundle agent cfe_autorun_inventory_ipv4_addresses
-# @brief Inventory ipv4 addresses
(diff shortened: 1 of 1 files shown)


inventory/any.cf
```diff
@@ -79,29 +79,27 @@ bundle agent cfe_autorun_inventory_listening_ports
...
-      string => "$(ipv4_addresses)",
-      meta => { "inventory", "attribute_name=IPv4 addresses" };
+      "ipv4[$(ip[$(if)])]" string => "$(if)",
+        meta => { "inventory", "attribute_name=map of IPv4 addresses to interface names" };
 
   reports:
     DEBUG|DEBUG_cfe_autorun_inventory_ipv4_addresses::
       "DEBUG $(this.bundle)";
-      "$(const.t)Inventorying: '$(ipv4_addresses)'";
+      "$(const.t)Inventorying: ipv4[$(ip[$(if)])] = '$(if)'";
```
nickanderson: Can you make these single line "DEBUG $(this.bundle): Inventorying: ipv4[$(ip[$(if)])] = '$(if)'";

nickanderson:
I had used a regex to exclude addresses (with the thought that variable could be defined in def.cf, and overriden with def.json). But I have not heard anyone wanting to exclude addresses other than 127.0.0.1 yet so I see no issue with this change, unless you can think of a reason why someone would want to exclude other ips from inventory.

tzz:
The report is fixed as requested.  I agree about the regex.  I would move this to the C code eventually (open a ticket, I'll do it) but it's a good improvement for now regardless.

nickanderson:
You would move it to C in what way?

tzz:
I'd put it in `sys.if2ip` or something like that, with the same contents as the array in this PR, at the same time that the map `sys.ip` is being generated.  It should be a one-line change.

tzz:
For the C implementation, see https://github.com/cfengine/core/pull/2290

inventory/any.cf
```diff
@@ -79,29 +79,26 @@ bundle agent cfe_autorun_inventory_listening_ports
...
+      # because the parser doesn't like them in array keys, even if
+      # they are being expanded.
+      "if" slist => { @(sys.interfaces) }; # note this excludes loopback
+      "ip[$(if)]" string => "$(sys.ipv4[$(if)])";
 
-    "ipv4[$(ipv4_addresses)]"
-      string => "$(ipv4_addresses)",
-      meta => { "inventory", "attribute_name=IPv4 addresses" };
+      "ipv4[$(ip[$(if)])]" string => "$(if)",
+        meta => { "inventory", "attribute_name=map of IPv4 addresses to interface names" };
```
nickanderson: Why update the attribute name? The value is what we are interested in here, which with your refactor should still be "IPv4 addresses"

tzz: I thought it was better to say what it is; it's definitely not just the `IPv4 addresses` as the current attribute says.  But it's not a big deal to leave it as it is.

nickanderson: OK, we use that inventory item OOTB in Enterprise so we should keep the name the same. At least for enterprise attribute names ought to describe the value of a variable that you are inventorying.

tzz: OK; fixed.

inventory/any.cf
```diff
@@ -79,29 +79,26 @@ bundle agent cfe_autorun_inventory_listening_ports
...
+      # because the parser doesn't like them in array keys, even if
+      # they are being expanded.
+      "if" slist => { @(sys.interfaces) }; # note this excludes loopback
+      "ip[$(if)]" string => "$(sys.ipv4[$(if)])";
 
-    "ipv4[$(ipv4_addresses)]"
-      string => "$(ipv4_addresses)",
-      meta => { "inventory", "attribute_name=IPv4 addresses" };
+      "ipv4[$(ip[$(if)])]" string => "$(if)",
+        meta => { "inventory", "attribute_name=IPv4 addresses" };
```
nickanderson: Ah, I am sorry Ted. I must have misread this previously. This completely changes the behaviour.

Now we are inventorying the interfaces instead of the ipv4 addresses (since $(if) is the name of the interface, and its the value of the variable). I am fine if we add a map, but the ip address it self needs to be in the value of the variable that's tagged for IPv4 addresses.

tzz:
Oh, let's forget it then.  https://github.com/cfengine/core/pull/2290 does it better anyhow.

PR #530, opened 2015-07-22
def.cf: separate inputs into a secondary def_inputs.cf file

tzz:
Because of our friend the 3-pass convergent fairy, the `def.cf` inputs are not evaluated in time.  This commit separates the loading of the inputs from JSON in `def.cf` from using them in `def_inputs.cf`.

This should probably have an acceptance test.

Ref: https://dev.cfengine.com/issues/7420

### DIFF
--- a/controls/3.7/def.cf
+++ b/controls/3.7/def.cf
@@ -4,11 +4,6 @@
 #  - common/global variables and classes here
 #
 ###############################################################################
-body file control
-{
-      inputs => { @(def.augments_inputs) };
-}
-
 bundle common def
 {
   classes:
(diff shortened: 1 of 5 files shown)


nickanderson:
I was working on that last night, and didnt quite make it as far as you.

Thanks! I will try this out soon.

tzz:
This PR is a great argument for loading the inputs and other things from def.json at the C level, **before** evaluation.

nickanderson:
Brilliant. With this I can add inputs to def.json, and then I can still use policies in services/autorun to activate the bundles that are in those additional inputs as specificed in def.json

This is really nice.

nickanderson:
trigger build

nickanderson:
Ted can you add a changelog entry for this?

PR #531, opened 2015-07-23
inventory: rework LLDP inventory to produce a data container and use -f keyvalue

tzz:
The `lldpctl` location was incorrect and the module script was broken.

This revision adds a full LLDP inventory into a data container.

### DIFF
--- a/controls/3.6/def.cf
+++ b/controls/3.6/def.cf
@@ -322,7 +322,7 @@ bundle common inventory_control
 # are disabled.
 {
   vars:
-      "lldpctl_exec" string => "/usr/bin/lldpctl";
+      "lldpctl_exec" string => "/usr/sbin/lldpctl";
       "lsb_exec" string => "/usr/bin/lsb_release";
       "mtab" string => "/etc/mtab";
       "proc" string => "/proc";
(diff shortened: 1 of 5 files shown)


nickanderson:
@estenberg 

thoughts?

estenberg:
What is this to be used for (I didn't even know we had LLDP inventory already)?

Reporting? Policy?

If reporting, then data container won't work that well.

estenberg:
Also not a big fan of throwing in scripts into masterfiles in general...

tzz:
This is useful for policy right now.  I'd rather generate a data container because the data is complex, and maybe reporting will catch up with data containers at some point.  Users can always generate inventory arrays from the data container meanwhile if they need it.

nickanderson:
Ted I am going to close this one out. I would take it if you could use `-f json`.

Maybe the command could be specified as a variable in `bundle common def` which would allow for a user to override and give it the path to a module

tzz:
The mainsteam LLDP distro (`lldpd` on Ubuntu 15.10) supports `-f json` now so I can resubmit this in a one-liner.

What should the variable name be? `cfe_autorun_inventory_LLDP.something`?

nickanderson:
cfe_autorun_inventory_LLDP.lldpctl_result ?

PR #540, opened 2015-08-11
Fix syntax error in HA policy.

pasinskim:


pasinskim:
@nickanderson some fix for the last HA related commit.

PR #543, opened 2015-08-18
HA changes

pasinskim:


pasinskim:
@nickanderson if you could take a look at this that would be great. Especially this might be interested for you: https://github.com/pasinskim/masterfiles/commit/19e3a810e05719a17432fdf698bc87e2bdf2e92d.

I think we can set some "common pattern" for using dynamic inputs. Everything should work as long as the classes are set in the same file and are not depending on other classes.

nickanderson:
I think those classes need to be setable from def.json so that users are not required to modify the shipped policy in order to enable or disable a feature.

pasinskim:
Agree, but moving this stuff back to update_def will cause that dynamic inputs wont work correctly. Then we will have to load all HA related policy all the time.
One solution might be to guard HA stuff inside enterprise/ha directory with the same class, but in this case at least policy will be parsed.

cfe_internal/enterprise/ha/ha_update.cf
```diff
@@ -0,0 +1,129 @@
...
+bundle agent ha_share_hub_keys
+{
+ files:
+   "$(ha_def.ppkeys_hubs)"
+      copy_from => no_backup_cp("$(sys.workdir)/ppkeys"),
+      file_select => hubs_keys_select,
+      comment => "Copy all hubs keys to directory accessible by clients",
+      handle => "ha_copy_hubs_keys",
+      depth_search => recurse("1"),
+      classes => if_repaired("hubs_keys_staged");
```
nickanderson: Is this class used somewhere?

cfe_internal/enterprise/ha/ha_update.cf
```diff
@@ -0,0 +1,129 @@
...
+
+bundle agent ha_hub_sync
+{
+ vars:
+  "exclude_files" slist => {"localhost.priv", "localhost.pub", @(ha_def.hub_shas)};
+ files:
+  "$(ha_update.ppkeys_staging)"
+      copy_from => ha_no_backup_scp("$(sys.workdir)/ppkeys", @(update_def.standby_servers)),
+      file_select => ex_list(@(exclude_files)),
+      comment => "Distribute all client keys between replica set servers",
```
nickanderson: Can we expand this comment to describe why this is necessary. What will happen if these keys do not get replicated between hubs.

cfe_internal/enterprise/ha/ha_update.cf
```diff
@@ -0,0 +1,129 @@
...
+      "ppkeys_staging" string => "/var/cfengine/ppkeys_staging";
+ methods:
+  enable_cfengine_enterprise_hub_ha.am_policy_hub::
+      "share_keys" usebundle => ha_share_hub_keys;
+      "sync_hub_data" usebundle => ha_hub_sync;
+
+  enable_cfengine_enterprise_hub_ha::
+      "sync_agent_data" usebundle => ha_agent_sync;
+
+  enable_cfengine_enterprise_hub_ha.hub_data_synced::
```
nickanderson: This manage_hub_synced_data bundle looks like it would be activated  on all agents, not just the policy_server. Is that supposed to happen?

cfe_internal/enterprise/ha/ha_update.cf
```diff
@@ -0,0 +1,129 @@
...
+      classes => if_repaired("hub_data_synced");
+}
+
+bundle agent manage_hub_synced_data
+{
+ files:
+   "$(sys.workdir)/ppkeys"
+      copy_from => no_backup_cp("$(ha_update.ppkeys_staging)"),
+      file_select => plain,
+      comment => "Copy staged client keys to ppkeys",
```
nickanderson: Can we expand the comment about why this is important. What might happen if these staged keys are not copied into ppkeys.

cfe_internal/enterprise/ha/ha_update.cf
```diff
@@ -0,0 +1,129 @@
...
+      handle => "ha_copy_hub_keys_to_nodes";
+}
+
+bundle agent ha_share_hub_keys
+{
+ files:
+   "$(ha_def.ppkeys_hubs)"
+      copy_from => no_backup_cp("$(sys.workdir)/ppkeys"),
+      file_select => hubs_keys_select,
+      comment => "Copy all hubs keys to directory accessible by clients",
```
nickanderson: Perhaps add something like "clients need to be able to download other hubs keys so that in the event of a fail over the client will be able to trust alternative hubs." to the comment?

cfe_internal/enterprise/ha/ha_update.cf
```diff
@@ -0,0 +1,129 @@
...
+      "manage_keys" usebundle => manage_hub_synced_data;
+}
+
+bundle agent ha_agent_sync
+{
+ files:
+  !am_policy_hub::
+    "$(sys.workdir)/policy_server.dat"
+      copy_from => ha_no_backup_scp("$(ha_update.master_hub_location)", @(update_def.policy_servers)),
+      comment => "Update master hub IP on CFEngine node",
```
nickanderson: This ensures that remote agents are looking at the active policy hub right?

Maybe the comment can be improved ....
So that remote agents  continue to get policy we need to make sure that they look to the currently active hub as their policy server.

cfe_internal/enterprise/ha/ha_update.cf
pasinskim: Fixed!

cfe_internal/enterprise/ha/ha_update.cf
pasinskim: Done.

cfe_internal/enterprise/ha/ha_update.cf
pasinskim: This will be activated only if 'hub_data_synced' is set. It can be set only in 'bundle agent ha_hub_sync' which is hub specific.

cfe_internal/enterprise/ha/ha_update.cf
```diff
@@ -0,0 +1,141 @@
+bundle agent ha_update
+{
+ vars:
+      "master_hub_location" string => "/var/cfengine/state/master_hub.dat";
+      "hubs_keys_location" string => "/var/cfengine/ppkeys_hubs";
+      "ppkeys_staging" string => "/var/cfengine/ppkeys_staging";
```
nickanderson: These paths are also shared by the hub and we have access promises for them right?
If these variables are defined in a common bundle, then the both the access promises in the server bundle and the promises in the agent bundle would be able to reference them. If there was ever a change, we would only have to update one location.

pasinskim: Makes sense, but this stuff is used in server bodies. I can try to experiment with common and server bodies to make sure it will work as expected.

pasinskim:
@nickanderson now all definitions are in one place.

cfe_internal/enterprise/ha/ha_def.cf
```diff
@@ -2,9 +2,12 @@ bundle common ha_def
 {
  vars:
   enable_cfengine_enterprise_hub_ha::
+   "master_hub_location" string => "/var/cfengine/state/master_hub.dat";
+   "hubs_keys_location" string => "/var/cfengine/ppkeys_hubs";
+   "ppkeys_staging" string => "/var/cfengine/ppkeys_staging";
+
+   #The location of the config file can not be changed!
    "config_file" string => "$(this.promise_dirname)/ha_info.json";
```
nickanderson: Nice

cfe_internal/enterprise/ha/ha_info.json
```diff
@@ -1,15 +1,15 @@
...
+  "is_in_cluster" : true,
  },
  "192.168.100.11":
  {
-  "sha": "37baa6f2f18a907131dadbf3e255d4ed0eda2f403e2123bab2c5f82f7ac8cd79",
+  "sha": "PLACE KEY HERE",
   "internal_ip": "192.168.100.11",
-  "tags": ["node2"]
+  "is_in_cluster" : true,
  }
```
nickanderson: Do you think this should be example_ha_info.json or something like what we did with def.json in the root of the repository?

nickanderson:
Looks pretty good

nickanderson:
trigger build

cfe_internal/enterprise/ha/ha_info.json
pasinskim: I don't know being honest. This is something which is not used if you don't have HA configured. Once you are starting using HA you need to fill this out and don't have too much choice - it MUST be formatted as shown here.

pasinskim:
@nickanderson build is green :)

PR #548, opened 2015-08-24
Add support for failover to 3rd HA node located outside cluster.

pasinskim:
In order to support environments with 3 HA nodes we need to
provide functionality allowing manual failover to 3rd HA node.
Idea is that 3rd node is used only as disaster-recovery node
and failover to that node will be manual.
CFEngine policy will limit manual steps to absolute minimum.

### DIFF
--- a/cfe_internal/enterprise/ha/ha_def.cf
+++ b/cfe_internal/enterprise/ha/ha_def.cf
@@ -22,6 +22,18 @@ bundle common ha_def
      slist => getvalues("hub_sha"),
      comment => "We use the list of hub key files for restricting clients access only to those";
 
+
+   "replication_only_node[$(ips)]"
+      string => "dummy", ifvarclass => strcmp("$(config[$(ips)][is_in_cluster])", "false");
+
+   "replication_only_nodes"
+      slist => getindices("replication_only_node");
+
+
+  classes:
+      "ha_replication_only_node" expression => some($(sys.policy_hub), replication_only_nodes);
+
+
   reports:
     verbose_mode::
       "HA hub $(ips) $(config[$(ips)][sha])";
--- a/cfe_internal/update/update_processes.cf
+++ b/cfe_internal/update/update_processes.cf
@@ -122,6 +122,14 @@ bundle agent maintain_cfe_hub_process
       comment => "Check for $(sys.workdir)/inputs/promises.cf",
       handle => "cfe_internal_maintain_cfe_hub_process_classes_files_ok";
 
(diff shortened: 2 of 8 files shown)


cfe_internal/enterprise/ha/ha_def.cf
```diff
@@ -22,6 +22,18 @@ bundle common ha_def
...
+
+   "replication_only_node[$(ips)]"
+      string => "dummy", ifvarclass => strcmp("$(config[$(ips)][is_in_cluster])", "false");
+
+   "replication_only_nodes"
+      slist => getindices("replication_only_node");
+
+
+  classes:
+      "ha_replication_only_node" expression => some($(sys.policy_hub), replication_only_nodes);
```
nickanderson: Nice, easy to read.

cfe_internal/update/update_processes.cf
```diff
@@ -154,12 +162,17 @@ bundle agent maintain_cfe_hub_process
       handle => "cfe_internal_maintain_cfe_hub_process_processes_postgres",
       ifvarclass => "nova|enterprise";
 
-    am_policy_hub.files_ok.!windows::
+    am_policy_hub.!enable_cfengine_enterprise_hub_ha.files_ok.!windows|ha_run_hub_process::
```
nickanderson: We could probably use an improved comment here. Under what conditions is it important to make sure the cf-hub process is running and why?

cfe_internal/update/update_processes.cf
```diff
@@ -154,12 +162,17 @@ bundle agent maintain_cfe_hub_process
...
       "cf-hub"      restart_class => "start_hub",
       comment => "Monitor cf-hub process",
       handle => "cfe_internal_maintain_cfe_hub_process_processes_cf_hub",
       ifvarclass => "(nova|enterprise).no_vacuumdb";
 
+    am_policy_hub.ha_kill_hub_process::
+      "cf-hub"      signals => { "term" },
+      comment => "Terminate cf-hub on backup HA node outside cluster",
```
nickanderson: Why is it important to make sure that cf-hub isn't running on replication nodes?

pasinskim: In order to not try contacting agents and try to collect reports. If we will have more than one hub collecting reports we will surely end up with mess.

cfe_internal/update/update_processes.cf
pasinskim: Makes sense. This looks pretty complicated already.

pasinskim:
trigger build!

PR #562, opened 2015-09-16
update paths.cf to include solaris svcprop

asil75:


nickanderson:
Can you also add it to the re-unified paths bundle so that when we release 3.8 3.7 and 3.8 clients will be able to take advantage of it? https://github.com/cfengine/masterfiles/blob/master/lib/paths.cf

nickanderson:
Closing in favour of https://github.com/cfengine/masterfiles/issues/563 which includes your patch.

Thanks for the contribution!

PR #567, opened 2015-09-24
Fix: Path to update def controls

nickanderson:
This one slipped past my testing when re-unifying the version split policies.


tzz:
YES

PR #568, opened 2015-09-25
Change: Move executor control mail settings

nickanderson:
Def is not fully resolved when evaluated by execd which makes it
currently not possible to set mail settings from the augments_file or
from the override vars defined in def. For now, these settings need to
be set directly for the best reliability.

Ref: https://dev.cfengine.com/issues/7607

This should be picked to 3.7.x as well if a better solution is not found.


nickanderson:
This is likely to be superseded by the def.json parsing from C

PR #572, opened 2015-10-10
Add path to FreeBSD's service command.

madpilot78:
This makes it possible for service promises to use the standard_services bundle on stock FreeBSD too.


nickanderson:
looks good to me.

PR #573, opened 2015-10-13
Ensure failsafe.cf is always present at clients.

eysteins:
ChangeLog: failsafe.cf is created in inputs/ during bootstrap, and
if the update.cf policy gets broken later, cf-agent will try to
run it. This change includes failsafe.cf in masterfiles in order
to ensure it is recreated if it gets removed after bootstrapping.

** Please note **
cfe_internal/failsafe/version MUST BE SYNCED from core/libpromises/failsafe.cf each release!
Do not merge until we figure out how to do this.


eysteins:
Will create another shortly.

eysteins:
Please see #574

PR #574, opened 2015-10-13
Ensure failsafe.cf is always present at clients.

estenberg:
ChangeLog: failsafe.cf is created in inputs/ during bootstrap, and if the update.cf policy gets broken later, cf-agent will try to run it. This change includes failsafe.cf in masterfiles in order to ensure it is recreated if it gets removed after bootstrapping.

** Please note **
cfe_internal/failsafe/version MUST BE SYNCED from core/libpromises/failsafe.cf each release!
Do not merge until we figure out how to do this.


nickanderson:
Failsafe.cf should be generated if it does not exist, is it ONLY generated during bootstrap now? Seems like it should be generated upon failsafe condition if it does not exist.

cfe_internal/failsafe/failsafe_copy.cf
```diff
@@ -0,0 +1,26 @@
...
+# by CFEngine and its original state is required for internal operations of
+# CFEngine. If the file has been modified CFEngine’s upgrades may require
+# manual intervention. Contact CFEngine support if additional information
+# and/or recommendation is needed.
+#
+##################################################################
+##################################################################
+#
+# cfe_internal_failsafe_copy
+#  - ensure $(sys.inputsdir)/failsafe.cf exists
```
jimis: s/inputsdir/inputdir/

cfe_internal/failsafe/failsafe_copy.cf
```diff
@@ -0,0 +1,26 @@
...
+#  - this file is fallen back to if update.cf has syntax errors
```
jimis: > this file is fallen back to if update.cf has syntax errors
I don't understand.

jimis:
`failsafe.cf` has no place in masterfiles. But I believe I'm missing context here, I don't see any redmine ticket mentioned in order to get some background info.

nickanderson:
We generate the failsafe in case there is not one. If there is one in your custom policy set it should be used. Shipping a failsafe in masterfiles would complicate upgrades for those who need a custom failsafe.

eysteins:
In my tests failsafe.cf was not generated except during bootstrap. However, hasn't that always been the behaviour?

But always generating it when cf-agent runs if it does not exist sounds like a much better solution than this.

eysteins:
No Redmine, just https://cfengine.zendesk.com/agent/tickets/2539

estenberg:
In my tests failsafe.cf was not generated except during bootstrap. However, hasn't that always been the behaviour?

But always generating it when cf-agent runs if it does not exist sounds like a much better solution than this.

estenberg:
No Redmine, just https://cfengine.zendesk.com/agent/tickets/2539

nickanderson:
I actually don't know all the failsafe specifics. I thought it was supposed to be generated if it didn't exist and the failsafe condition triggers.

If purging was enabled (very common) ,  this file would be regularly refreshed, at least during policy updates.



Just a note from my brain 
There was once talk of having a failsafe/failsafe.cf as a failsafe to the failsafe. IIRC it was discussed that file would always be generated from c when necessary as the assumption is if failsafe fails it's bad custom policy.  I don't think that ever happened. And if we looked to do it I would consider a place outside inputs.

estenberg:
Yes, the fact that purging could be enabled is a case for keeping it in masterfiles.
Custom failsafe is a bit exotic, I think, as it's supposed to be this very simple and rock-solid thing.

This might be the best solution then:
* Look for failsafe outside of outputs, e.g. $(sys.workdir)/failsafe.cf
* Auto-generate failsafe.cf from C if it doesn't exist
* Fall back to the failsafe outside of outputs (although now we're breaking custom failsafe cases, ugh)

jimis:
Currently `failsafe.cf` is auto-generated only during `--bootstrap`, I thought it was quite clear. Furthermore, I don't see how `failsafe.cf` has anything to do with masterfiles.

The proper solution IMHO is to also generate it **when it needs to run** (e.g. in case of syntax error), and only if the file doesn't exist.

basvandervlies:
I agree @jimis. Keep it simple

nickanderson:
I also agree. failsafe.cf should be generated if it does not exist, and only when failsafe condition is triggered.

estenberg:
Created ticket here: https://dev.cfengine.com/issues/7634

PR #575, opened 2015-10-13
Add FreeBSD ports package module and tests

zofrex:
This is my initial stab at adding support for the FreeBSD ports system via a package module.

Nice things:
* It exists and it works – yay FreeBSD support!
* It has no dependencies outside of the FreeBSD base system (other than assuming a bootstrapped pkg, which I think is reasonable)
* It's tested

Areas that (might) need work that I know of:
* The Architecture is faked at the moment
* It can't install more than one package at a time – it wasn't clear to me from the docs if this is necessary
* My m4 knowledge is non-existent so the way I've crammed my tests in is probably not ideal

### DIFF
--- a/lib/packages.cf
+++ b/lib/packages.cf
@@ -94,6 +94,29 @@ body package_module yum
     #default_options => {};
 }
 
+body package_module freebsd_ports
+# @brief Define details used when interfacing with the freebsd ports package
+# module.
+#
+# **Note:** Ports are expected to be setup prior to trying to use the packages
+# promise. You may need to ensure that `portsnap extract` has been run, e.g.
+# fileexists("/usr/ports/Mk/bsd.port.mk")
+#
+# **Example:**
+# ```cf3
+# bundle agent main
+# {
+#   packages:
+#     freebsd::
+#       "vim"
+#         policy => "present",
+#         package_module => freebsd_ports;
+# }
+# ```
+{
+    query_installed_ifelapsed => "60";
+    query_updates_ifelapsed => "1440";
+}
 
 bundle common packages_common
 # @ignore
--- a/modules/packages/freebsd_ports
+++ b/modules/packages/freebsd_ports
@@ -1,0 +1,108 @@
+#!/bin/sh -e
+
+command=$1
+
+while read -r line; do
+  export INPUT_$line
+done
+
+get_package_data() {
+  name="${INPUT_File?File must be given to get-package-data}"
+  echo PackageType=repo
+  echo Name=$name
+}
+
+list_installed() {
+  # Example pkg output:
+  # sudo-1.8.14p3
+  # Name           : sudo
+  # Version        : 1.8.14p3
+  # Installed on   : Sun Aug 16 05:36:05 UTC 2015
+  # Origin         : security/sudo
+  # Architecture   : freebsd:10:x86:64
+  #
+  # After rewrite:
+  # Name=sudo
+  # Version=1.8.14p3
+  # Architecture=none
+  pkg info -f -a | egrep '^(Name|Version|Architecture)' | sed -e 's/[ ]*:[ ]*/=/' -e 's/^Architecture=.*/Architecture=none/'
+}
+
+repo_install() {
+  name="${INPUT_Name?Name must be given to repo-install}"
+  version="${INPUT_Version}"
+
+  export BATCH=1
+  PORT_PATH=$(whereis -sq "$name")
+
+  if [ -z "$PORT_PATH" ]
+  then
+    echo "ErrorMessage=Could not install $name, port does not exist"
+    exit 0
+  fi
+
+  cd "$PORT_PATH"
+
+  if [ -n "$version" ]
+  then
+    available=$(make -V PKGVERSION)
+    if [ "$available" != "$version" ]
+    then
+      echo "ErrorMessage=Could not install $name $version, available version was $available"
+      exit 0
+    fi
+  fi
+
+  make deinstall reinstall >&2
+}
+
+update_ports_tree() {
+  portsnap --interactive fetch update >&2
+}
+
+list_updates_local() {
+  # Example pkg output:
+  # ca_root_nss-3.19.3                 <   needs updating (index has 3.20)
+  #
+  # After sed:
+  # Name=ca_root_nss
+  # Version=3.20
+  # Architecture=none
+  pkg version -v -l "<" | sed -e 's/\([^ ]*\)-[^-]* .* \(.*\))/Name=\1\
+Version=\2\
+Architecture=none/'
+}
+
+remove() {
+  name="${INPUT_Name?Name must be given to remove}"
+  BATCH=1
+  cd $(whereis -sq "$name")
+  make deinstall >&2
+}
+
+case $command in
+  supports-api-version)
+    echo 1
+  ;;
+  get-package-data)
+    get_package_data
+  ;;
+  list-installed)
+    list_installed
+  ;;
+  repo-install)
+    repo_install
+  ;;
+  list-updates)
+    update_ports_tree
+    list_updates_local
+  ;;
+  list-updates-local)
+    list_updates_local
+  ;;
+  remove)
+    remove
+  ;;
+  *)
+    echo "ErrorMessage=Invalid operation"
+esac
--- a/tests/unit/Makefile.am
+++ b/tests/unit/Makefile.am
@@ -7,8 +7,9 @@ check_SCRIPTS += \
 	test_package_module_yum
 endif
 
-if HAVE_PKG_INSTALL
 if HAVE_SHUNIT2
+check_SCRIPTS += test_package_module_freebsd_ports
+if HAVE_PKG_INSTALL
 check_SCRIPTS += test_package_module_pkgsrc
 endif
 endif
@@ -16,12 +17,14 @@ endif
 TESTS = $(check_SCRIPTS)
 
 EXTRA_DIST = \
+	harness_freebsd_ports \
 	mock_apt_get \
 	mock_dpkg \
 	mock_dpkg_deb \
 	mock_dpkg_query \
 	mock_pkgin \
 	mock_pkg_info \
+	mock_freebsd_pkg \
 	mock_rpm \
 	mock_yum
 
--- a/tests/unit/harness_freebsd_ports
+++ b/tests/unit/harness_freebsd_ports
@@ -1,0 +1,97 @@
+#!/bin/sh -e
+
+whereis() {
+  case "$*" in
+  "-sq nano")
+    echo "/usr/ports/editors/nano"
+  ;;
+  "-sq ca_root_nss")
+    echo "/usr/ports/security/ca_root_nss"
+  ;;
+  "-sq madeup")
+  ;;
+  *)
+    >&2 echo "ERROR: whereis incorrect args: $*"
+    exit 1
+  ;;
+  esac
+}
+
+cd() {
+  CD_ARGS="$*"
+}
+
+make() {
+  if [ "$BATCH" != "1" ]
+  then
+    >&2 echo "ERROR: Make: BATCH != 1"
+    exit 1
+  fi
+
+  if [ "$CD_ARGS" != "$ASSERT_CWD" ]
+  then
+    >&2 echo "ASSERT:make expected to be run in directory:<$ASSERT_CWD> but ran in:<$CD_ARGS>"
+    exit 1
+  fi
+
+  if [ "$*" = "-V PKGVERSION" ]
+  then
+    case $CD_ARGS in
+    /usr/ports/editors/nano)
+      echo "2.4.2"
+    ;;
+    /usr/ports/security/ca_root_nss)
+      echo "3.20_1"
+    ;;
+    esac
+    return 0
+  fi
+
+  echo "Make output"
+
+  MAKE_RAN="$*"
+}
+
+portsnap() {
+  PORTSNAP_RAN=1
+  echo "Portsnap output"
+  expected="--interactive fetch update"
+  if [ "$*" != "$expected" ]
+  then
+    >&2 echo "ASSERT:portsnap expected:<$expected> but was:<$*>"
+    exit 1
+  fi
+}
+
+finish_and_assert() {
+  if [ "$ASSERT_MAKE_RUNS" ]
+  then
+    if [ ! "$MAKE_RAN" ]
+    then
+      >&2 echo "ASSERT:make was expected to run:<$ASSERT_MAKE_RUNS> but did not run"
+      exit 1
+    elif [ "$MAKE_RAN" != "$ASSERT_MAKE_RUNS" ]
+    then
+      >&2 echo "ASSERT:make task expected:<$ASSERT_MAKE_RUNS> but was:<$MAKE_RAN>"
+      exit 1
+    fi
+  elif [ "$MAKE_RAN" ]
+  then
+    >&2 echo "ASSERT:make was not expected to run but ran:<$MAKE_RAN>"
+    exit 1
+  fi
+
+  if [ "$ASSERT_PORTS_UPDATED" -a ! "$PORTSNAP_RAN" ]
+  then
+    >&2 echo "ASSERT:portsnap was expected to be run but it was not"
+    exit 1
+  fi
+}
+
+pkg() {
+  ./mock_freebsd_pkg "$@"
+}
+
+trap finish_and_assert EXIT
+
+. ../../modules/packages/freebsd_ports
--- a/tests/unit/test_package_module_freebsd_ports
+++ b/tests/unit/test_package_module_freebsd_ports
@@ -1,0 +1,183 @@
+#! /bin/sh
+
+FREEBSD_PORTS=./harness_freebsd_ports
+
+setUp()
+{
+  unset ASSERT_MAKE_RUNS
+  unset ASSERT_CWD
+  unset ASSERT_PORTS_UPDATED
+}
+
+testApiVersion()
+{
+  actual=`echo | $FREEBSD_PORTS supports-api-version`
+
+  assertEquals "exit code" 0 $?
+  assertEquals 1 "$actual"
+}
+
+testInvalidComand()
+{
+  actual=`echo | $FREEBSD_PORTS oogly-boogly-moogly`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "ErrorMessage=Invalid operation" "$actual"
+}
+
+testGetData()
+{
+  input="File=zip
+Version=3.0-4
+Architecture=none"
+  expected="PackageType=repo
+Name=zip"
+
+  actual=`echo "$input" | $FREEBSD_PORTS get-package-data`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "$expected" "$actual"
+}
+
+testListPackages()
+{
+  expected="Name=autoconf
+Version=2.69
+Architecture=none
+Name=autoconf-wrapper
+Version=20131203
+Architecture=none
+Name=cfengine
+Version=3.7.0_1
+Architecture=none
+Name=gettext-tools
+Version=0.19.5.1
+Architecture=none
+Name=sudo
+Version=1.8.14p3
+Architecture=none"
+
+  actual=`echo | $FREEBSD_PORTS list-installed`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "$expected" "$actual"
+}
+
+testInstallPackage()
+{
+  input="Name=nano
+Version=2.4.2
+Architecture=none"
+  ASSERT_MAKE_RUNS="deinstall reinstall"
+  export ASSERT_MAKE_RUNS
+  ASSERT_CWD=/usr/ports/editors/nano
+  export ASSERT_CWD
+
+  actual=`echo "$input" | $FREEBSD_PORTS repo-install`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "" "$actual"
+}
+
+testListUpdates()
+{
+  expected="Name=ca_root_nss
+Version=3.20_1
+Architecture=none
+Name=portmaster
+Version=3.17.8
+Architecture=none"
+  ASSERT_PORTS_UPDATED=1
+  export ASSERT_PORTS_UPDATED
+
+  actual=`echo | $FREEBSD_PORTS list-updates`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "$expected" "$actual"
+}
+
+
+testListUpdatesLocal()
+{
+  expected="Name=ca_root_nss
+Version=3.20_1
+Architecture=none
+Name=portmaster
+Version=3.17.8
+Architecture=none"
+
+  actual=`echo | $FREEBSD_PORTS list-updates-local`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "$expected" "$actual"
+}
+
+testUpdatePackage()
+{
+  input="Name=ca_root_nss
+Version=3.20_1"
+  ASSERT_MAKE_RUNS="deinstall reinstall"
+  export ASSERT_MAKE_RUNS
+  ASSERT_CWD=/usr/ports/security/ca_root_nss
+  export ASSERT_CWD
+
+  actual=`echo "$input" | $FREEBSD_PORTS repo-install`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "" "$actual"
+}
+
+testInstallWrongVersion()
+{
+  input="Name=nano
+Version=2.4.3
+Architecture=none"
+  ASSERT_CWD=/usr/ports/editors/nano
+  export ASSERT_CWD
+
+  actual=`echo "$input" | $FREEBSD_PORTS repo-install`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "ErrorMessage=Could not install nano 2.4.3, available version was 2.4.2" "$actual"
+}
+
+testInstallAnyVersion()
+{
+  input="Name=nano
+Architecture=none"
+  ASSERT_MAKE_RUNS="deinstall reinstall"
+  export ASSERT_MAKE_RUNS
+  ASSERT_CWD=/usr/ports/editors/nano
+  export ASSERT_CWD
+
+  actual=`echo "$input" | $FREEBSD_PORTS repo-install`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "" "$actual"
+}
+
+testRemove()
+{
+  input="Name=nano"
+  ASSERT_MAKE_RUNS=deinstall
+  export ASSERT_MAKE_RUNS
+  ASSERT_CWD=/usr/ports/editors/nano
+  export ASSERT_CWD
+
+  actual=`echo "$input" | $FREEBSD_PORTS remove`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "" "$actual"
+}
+
+testTryInstallNonexistentPackage()
+{
+  input="Name=madeup"
+
+  actual=`echo "$input" | $FREEBSD_PORTS repo-install`
+
+  assertEquals "exit code" 0 $?
+  assertEquals "ErrorMessage=Could not install madeup, port does not exist" "$actual"
+}
+
+. `which shunit2`
--- a/modules/packages/Makefile.am
+++ b/modules/packages/Makefile.am
@@ -1,3 +1,3 @@
 package_modulesdir = $(prefix)/modules/packages
 
-dist_package_modules_SCRIPTS = apt_get yum pkgsrc
+dist_package_modules_SCRIPTS = apt_get yum pkgsrc freebsd_ports
--- a/tests/unit/mock_freebsd_pkg
+++ b/tests/unit/mock_freebsd_pkg
@@ -1,0 +1,177 @@
+#!/bin/sh -e
+
+case "$*" in
+"info -f -a")
+  cat <<EOF
+autoconf-2.69
+Name           : autoconf
+Version        : 2.69
+Installed on   : Tue Aug 18 01:11:54 UTC 2015
+Origin         : devel/autoconf
+Architecture   : freebsd:10:x86:64
+Prefix         : /usr/local
+Categories     : devel
+Licenses       :
+Maintainer     : tijl@FreeBSD.org
+WWW            : http://www.gnu.org/software/autoconf/
+Comment        : Automatically configure source code on many Un*x platforms
+Annotations    :
+Flat size      : 2.90MiB
+Description    :
+Autoconf is an extensible package of m4 macros that produce shell
+scripts to automatically configure software source code packages.
+These scripts can adapt the packages to many kinds of UNIX-like
+systems without manual user intervention.  Autoconf creates a
+configuration script for a package from a template file that lists the
+operating system features that the package can use, in the form of m4
+macro calls.
+
+WWW: http://www.gnu.org/software/autoconf/
+
+autoconf-wrapper-20131203
+Name           : autoconf-wrapper
+Version        : 20131203
+Installed on   : Tue Aug 18 01:11:53 UTC 2015
+Origin         : devel/autoconf-wrapper
+Architecture   : freebsd:10:x86:64
+Prefix         : /usr/local
+Categories     : devel
+Licenses       :
+Maintainer     : tijl@FreeBSD.org
+WWW            : UNKNOWN
+Comment        : Wrapper script for GNU autoconf
+Annotations    :
+Flat size      : 2.93KiB
+Description    :
+This port installs a wrapper script for autoconf, with symlinks to the
+unversioned name of each tool included with autoconf.  This allows the
+correct version to be selected depending on the user's requirements.
+
+cfengine-3.7.0_1
+Name           : cfengine
+Version        : 3.7.0_1
+Installed on   : Tue Sep 29 23:21:28 UTC 2015
+Origin         : sysutils/cfengine
+Architecture   : freebsd:10:x86:64
+Prefix         : /usr/local
+Categories     : sysutils
+Licenses       : GPLv3
+Maintainer     : cy@FreeBSD.org
+WWW            : http://www.cfengine.org/
+Comment        : Systems administration tool for networks
+Options        :
+	LIBVIRT        : off
+	MYSQL          : off
+	PGSQL          : off
+Shared Libs required:
+	libpcre.so.1
+	libssl.so.8
+	liblmdb.so
+	libcrypto.so.8
+Shared Libs provided:
+	libpromises.so.3
+Annotations    :
+	cpe            : cpe:2.3:a:gnu:cfengine:3.7.0:::::freebsd10:x64:1
+Flat size      : 2.38MiB
+Description    :
+Cfengine is an automated suite of programs for configuring and
+maintaining Unix-like computers. It has been used on computing arrays
+of between 1 and 20,000 computers since 1993 by a wide range of
+organizations. Cfengine is supported by active research and was the
+first autonomic, hands-free management system for Unix-like operating
+systems. Cfengine is an autonomic maintenance system not merely a
+change management roll-out tool. Cfengine has a history of security
+and adaptability.
+
+WWW: http://www.cfengine.org/
+
+gettext-tools-0.19.5.1
+Name           : gettext-tools
+Version        : 0.19.5.1
+Installed on   : Sun Aug 16 05:35:50 UTC 2015
+Origin         : devel/gettext-tools
+Architecture   : freebsd:10:x86:64
+Prefix         : /usr/local
+Categories     : devel
+Licenses       : GPLv3
+Maintainer     : tijl@FreeBSD.org
+WWW            : http://www.gnu.org/software/gettext/
+Comment        : GNU gettext development and translation tools
+Options        :
+	DOCS           : on
+	THREADS        : on
+Shared Libs required:
+	libintl.so.8
+	libexpat.so.1
+Shared Libs provided:
+	libgettextpo.so.0
+	libgettextlib-0.19.5.1.so
+	libgettextsrc-0.19.5.1.so
+Annotations    :
+	cpe            : cpe:2.3:a:gnu:gettext:0.19.5.1:::::freebsd10:x64
+Flat size      : 9.78MiB
+Description    :
+GNU gettext is a framework of libraries and tools for internationalisation
+and localisation of software.
+
+This package contains development and translation tools.
+
+WWW: http://www.gnu.org/software/gettext/
+sudo-1.8.14p3
+Name           : sudo
+Version        : 1.8.14p3
+Installed on   : Sun Aug 16 05:36:05 UTC 2015
+Origin         : security/sudo
+Architecture   : freebsd:10:x86:64
+Prefix         : /usr/local
+Categories     : security
+Licenses       : sudo
+Maintainer     : garga@FreeBSD.org
+WWW            : http://www.sudo.ws/
+Comment        : Allow others to run commands as root
+Options        :
+	AUDIT          : on
+	DISABLE_AUTH   : off
+	DISABLE_ROOT_SUDO: off
+	DOCS           : on
+	INSULTS        : on
+	LDAP           : off
+	NLS            : on
+	NOARGS_SHELL   : on
+	OPIE           : off
+	SSSD           : off
+Shared Libs required:
+	libintl.so.8
+Shared Libs provided:
+	libsudo_util.so.0
+Annotations    :
+	cpe            : cpe:2.3:a:todd_miller:sudo:1.8.14p3:::::freebsd10:x64
+Flat size      : 3.33MiB
+Description    :
+This is the CU version of sudo.
+
+Sudo is a program designed to allow a sysadmin to give limited root
+privileges to users and log root activity.  The basic philosophy is to
+give as few privileges as possible but still allow people to get their
+work done.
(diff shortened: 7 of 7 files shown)


nickanderson:
Oh nice.  I need to try this with my freenas.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh -e
+
+command=$1
+
+while read -r line; do
+  export INPUT_$line
+done
+
+list_updates() {
+  pkg version -v -l "<" | sed -Ee $'s/([^ ]+)-[^-]+ .* (.*)\\)/Name=\\1\\\nVersion=\\2\\\nArchitecture=freebsd:10:x86:64/'
```
zofrex: This is the place where the Architecture is faked. We can easily get the arch of each package that's installed from `pkg info` but when we get a list of available updates from `pkg version` that information isn't there.

This could be fixed but it might be tricky, either look up what each package is or combine the two lists somehow.

One thing I want to mention is that I don't think Architecture serves any purpose on this platform – from my understanding all packages installed will be the same arch. I tried installing some Linux packages on FreeBSD (which counts as a different arch in a sense) and they still showed up as freebsd:10:x86:64. Is Architecture strictly necessary? Leaving it out entirely caused issues, can it be blank? Could it be made optional within cfengine core?

tests/unit/harness_freebsd_ports
```diff
@@ -0,0 +1,95 @@
...
+
+  if [ "$ASSERT_PORTS_UPDATED" -a ! "$PORTSNAP_RAN" ]
+  then
+    >&2 echo "ASSERT:portsnap was expected to be run but it was not"
+    exit 1
+  fi
+}
+
+function pkg {
+  ./mock_pkg "$@"
```
zofrex: When I originally made this, I mocked pkg by adding to the PATH. When I copied my work into this repository I renamed pkg to mock_pkg and put it with the other files instead of in its own directory, to follow the conventions here. That's why this one 'mock' (stub really) is different to the others.

tests/unit/harness_freebsd_ports
```diff
@@ -0,0 +1,95 @@
+#!/bin/sh -e
```
zofrex: This wrapper exists because I need to override shell built-ins, but I can't source the script under test from the test file (it has to be a subprocess), and you cannot export functions in Bourne. It's a bit weird. Let me know if I need to make any parts clearer.

tests/unit/test_package_module_freebsd_ports
```diff
@@ -0,0 +1,165 @@
+#! /bin/sh -e
+
+export PATH=tests/stubs:$PATH
```
zofrex: Note to self: this line should be removed

kacfengine:
Awesome! And +1 for the testing! I'll take a closer look at this as soon as I can!

nickanderson:
vagrant init  freebsd/FreeBSD-10.2-RELEASE
vagrant up

lets see if i can get this thing to work

nickanderson:
One thing missing from this PR though is a freebsd_ports package module body.

should go to lib/packages
```
body package_module freebsd_ports
{
    query_installed_ifelapsed => "60";
    query_updates_ifelapsed => "1440";
    #default_options =>  {};
}
```

After I put that in place and added my package promise:

```
  packages:
    "vim"
      policy => "present",
      package_module => freebsd_ports;
```

I got this:

```
root@:/var/cfengine/masterfiles # cf-agent -KIf ./promises.cf
   error: Could not get list of installed packages
Looking up portsnap.FreeBSD.org mirrors... 7 mirrors found.
Fetching public key from your-org.portsnap.freebsd.org... done.
Fetching snapshot tag from your-org.portsnap.freebsd.org... done.
Fetching snapshot metadata... done.
Fetching snapshot generated at Fri Oct 16 00:07:09 UTC 2015:
3faa49c831e5b231931741466a097239893477f67d7bbf100% of   69 MB 1066 kBps 01m07s
Extracting snapshot... done.
Verifying snapshot integrity... done.
Fetching snapshot tag from your-org.portsnap.freebsd.org... done.
Fetching snapshot metadata... done.
Updating from Fri Oct 16 00:07:09 UTC 2015 to Fri Oct 16 21:25:17 UTC 2015.
Fetching 4 metadata patches... done.
Applying metadata patches... done.
Fetching 0 metadata files... done.
Fetching 102 patches. 
(102/102) 100.00%  done.                                    
done.
Applying patches... 
done.
Fetching 1 new ports or files... done.
/usr/ports was not created by portsnap.
You must run 'portsnap extract' before running 'portsnap update'.
    info: Some error occurred while updating available updates cache.
    info: Can not update packages cache.
make: don't know how to make install. Stop

make: stopped in /root
    info: Some error occurred while communicating with package module while installing package.
   error: Error installing package 'vim'
```

I had to manually run portsnap extract, but after I did that running my policy seemed to work ... its still building, will update when it ever finishes.

I suppose that it would make the most sense for that to go into a maintenance policy inside. Not sure if it makes sense for the package module itself to change the system other than as specified by the protocol. Does anyone else have thoughts on that?

Looking forward to one that works for pkg

nickanderson:
And, worky 
```
====> Compressing man pages (compress-man)
===>  Installing for vim-7.4.891
===>  Checking if vim already installed
===>   Registering installation for vim-7.4.891
Installing vim-7.4.891...
    info: Successfully installed package 'vim'
```

nickanderson:
@kacfengine should modules do anything to the system except the strict interface with the package manager?

Note above where I had to manually run `portsnap extract` before trying to use the package_module.
I lean toward the module NOT running portsnap extract and that being something you need to have policy for in order to setup ports for use. Because you might prefer to use a different method than portsnap.

nickanderson:
@zofrex I submitted a PR to your branch to include support in the stdlib.

https://github.com/zofrex/masterfiles/pull/1/files

zofrex:
I lean towards this script not installing the ports tree too. If you're using ports, you've presumably already done that to install CFengine itself.

nickanderson:
Yeah we can always introduce a bundle for it, or  augment the in-line example to also show setting up ports with policy first (maybe cfe got installed via  pkg and not ports)

So it works from my testing, @kacfengine anything else before this can go in?

modules/packages/freebsd_ports
rootwyrm: This is the wrong way to do it. Correct way is via pkg query to extract relevant data.

```
$ pkg query '%n %v %q %R'
perl5 5.18.4_17 freebsd:10:x86:64 DragonNorth
$ pkg shell 
sqlite> select * from packages where name = 'perl5';
17|lang/perl5.18|perl5|5.18.4_17|Practical Extraction and Report Language|Perl is a language that combines some of the features of C, sed, awk and shell.  See the manual page for more hype.  There are also many books published by O'Reilly & Assoc.  See pod/perlbook.pod for more
information.

WWW: http://www.perl.org/|||FreeBSD:10:amd64|perl@FreeBSD.org|http://www.perl.org/|/usr/local|51917202|1|0|124|1445255055|2$0$jpuj587kyjfez7bcceod8r8m49wrye6gbg3j7igpnczzytfndify||
```

It is vitally important to NOT fake the architecture because pkg is multiplatform - not just FreeBSD. As I mentioned on IRC, a lot of the trickery here is going to be pkg query or pkg shell. They are often dramatically faster (2-4x) and will only present relevant information. Note that architecture must be handled *case insensitive* if you are mixing query and shell (shell -c "SQL Command"). The architecture needs to be parsed out into distinct fields as well. Breaking it down:

freebsd - OS
10 - major; major-minor is no longer present in pkg
x86 - architecture may be x86, arm, ia64, ppc, etc.
64 - endianness

rootwyrm:
This is going to require a check for ports vs pkg before it's safe for commit; you cannot mix ports and pkg safely, or vice versa. That should be fairly easy though - do *not* key off /usr/ports, that way misery lies. Instead key off ```pkg query %R``` - ports will be "unknown-repository", quarterly is "FreeBSD", and custom is custom string (and should be signed.)

Other than that and arch fixup it looks like a good start to me. I'll try and find some time to test here ASAP and then take a look at output cleanup. I have a host of stuff already done for that which can be integrated fairly easily.

modules/packages/freebsd_ports
zofrex: Neither of those commands give me what we need here though - the latest version number available via ports:

```
$ pkg version -v -l "<" | grep weechat
weechat-1.0.1_1                    <   needs updating (index has 1.1.1_1)

$ pkg query '%n %v %q %R' | grep weechat
weechat 1.0.1_1 freebsd:9:x86:64 FreeBSD

$ pkg shell
SQLite version 3.8.8.2 2015-01-30 14:30:45
Enter ".help" for usage hints.
sqlite> select * from packages where name = 'weechat';

90|irc/weechat|weechat|1.0.1_1|Lightweight and user friendly ncurses based IRC client|WeeChat (Wee Enhanced Environment for Chat) is a fast and light IRC
client. Everything can be done with a keyboard, and it is customizable
and extensible with scripts.

WWW: http://weechat.org|||freebsd:9:x86:64|jase@FreeBSD.org|http://weechat.org|/usr/local|18505968|0|0|1||1417486125||2$0$tqhgmq53qqc6pzmd7f8oxx4f318zedk34y1idrtg4ycyyitniimb
```

kacfengine:
> @kacfengine should modules do anything to the system except the strict interface with the package manager?

No. Sounds sensible that the ports tree is not installed by the module. It should assume that this is available already.

Not knowing much about FreeBSD I can't comment on the implementation so much, but the package API seems correctly used, and there are nice tests here. I'm ok with merging this, but do you want to address @rootwyrm's latest comment first?

kacfengine:
trigger build

m4/cf3_platforms.m4
```diff
@@ -39,3 +39,4 @@ AM_CONDITIONAL([HPUX], [test -n "`(echo ${target_os} | egrep 'hpux|hp-ux')`"])
 AM_CONDITIONAL([FREEBSD], [test -n "`(echo ${target_os} | grep freebsd)`"])
 AM_CONDITIONAL([NETBSD], [test -n "`(echo ${target_os} | grep netbsd)`"])
 AM_CONDITIONAL([XNU], [test -n "`(echo ${target_os} | grep darwin)`"])
+AM_CONDITIONAL([SHUNIT], [test -n "`which shunit2`"])
```
kacfengine: Testing reveals that this test apparently evaluates to true on both Solaris and HP-UX, which is presumably wrong. Should probably base it on something more specific to FreeBSD.

kacfengine: Actually, there is a FreeBSD entry there already, why not use that? ``SHUNIT`` is foreign to me anyway.

zofrex: The tests actually work on any POSIX platform that has shunit available (I did most of the development for this script on OS X). If they don't I would consider that a bug, as I tried very hard to avoid bashisms and only use portable things :)

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,64 @@
...
+  ;;
+  list-installed)
+    pkg info -f -a | grep -E '^(Name|Version|Architecture)' | sed -Ee 's/[ ]+:[ ]+/=/' -e 's/^Architecture=.*/Architecture=none/'
+  ;;
+  repo-install)
+    name="${INPUT_Name?Name must be given to repo-install}"
+    version="${INPUT_Version}"
+
+    export BATCH=1
+    cd `whereis -sq "$name"`
```
zofrex: Bug here - we don't check that `whereis -sq` gives a result (in other words, whether or not the package requested actually exists). Is this something repo-install should handle gracefully, or should we be checking for this (and potentially returning something different) in get-package-data?

m4/cf3_platforms.m4
kacfengine: Here is a snippet from the failing test on Solaris 9:

```
11:11:30 /opt/csw/bin/gmake   test_package_module_freebsd_ports
11:11:30 gmake[2]: Entering directory `/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/masterfiles/tests/unit'
11:11:30 gmake[2]: Nothing to be done for `test_package_module_freebsd_ports'.
11:11:30 gmake[2]: Leaving directory `/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/masterfiles/tests/unit'
11:11:30 /opt/csw/bin/gmake  check-TESTS
11:11:30 gmake[2]: Entering directory `/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/masterfiles/tests/unit'
11:11:30 ./test_package_module_freebsd_ports: PATH=tests/stubs:/usr/bin:/opt/csw/bin:/opt/sfw/bin:/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/./buildscripts/build-scripts:/var/cfengine/httpd/php/bin:/usr/local/bin:/usr/sbin:/usr/sfw/bin:/usr/bin:/usr/local/sbin:/usr/contrib/bin:/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/./buildscripts/build-scripts:/var/cfengine/httpd/php/bin:/usr/local/bin:/usr/sbin:/usr/sfw/bin:/usr/bin:/usr/local/sbin:/usr/contrib/bin:/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/buildscripts/build-scripts:/var/cfengine/httpd/php/bin:/usr/local/bin:/usr/sbin:/usr/sfw/bin:/usr/bin:/usr/local/sbin:/usr/contrib/bin:/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/buildscripts/build-scripts:/var/cfengine/httpd/php/bin:/usr/local/bin:/usr/sbin:/usr/sfw/bin:/usr/bin:/usr/local/sbin:/usr/contrib/bin: is not an identifier
11:11:30 FAIL: test_package_module_freebsd_ports
11:11:30 ==================
11:11:30 1 of 1 test failed
11:11:30 ==================
11:11:30 gmake[2]: *** [check-TESTS] Error 1
11:11:30 gmake[2]: Leaving directory `/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/masterfiles/tests/unit'
11:11:30 gmake[1]: *** [check-am] Error 2
11:11:30 gmake[1]: Leaving directory `/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/masterfiles/tests/unit'
11:11:30 gmake: *** [check-recursive] Error 1
11:11:30 gmake: Leaving directory `/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ULTRASPARC_SOLARIS_9/masterfiles'
```

And from HP-UX 11.23:

```
11:49:30 /usr/local/bin/gmake   test_package_module_freebsd_ports
11:49:30 gmake[2]: Entering directory '/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ia64_hpux_11.23/masterfiles/tests/unit'
11:49:30 gmake[2]: Nothing to be done for 'test_package_module_freebsd_ports'.
11:49:30 gmake[2]: Leaving directory '/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ia64_hpux_11.23/masterfiles/tests/unit'
11:49:30 /usr/local/bin/gmake  check-TESTS
11:49:30 gmake[2]: Entering directory '/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ia64_hpux_11.23/masterfiles/tests/unit'
11:49:30 ./test_package_module_freebsd_ports[165]: no:  not found.
11:49:30 FAIL: test_package_module_freebsd_ports
11:49:30 ==================
11:49:30 1 of 1 test failed
11:49:30 ==================
11:49:30 Makefile:201: recipe for target 'check-TESTS' failed
11:49:30 gmake[2]: *** [check-TESTS] Error 1
11:49:30 gmake[2]: Leaving directory '/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ia64_hpux_11.23/masterfiles/tests/unit'
11:49:30 Makefile:324: recipe for target 'check-am' failed
11:49:30 gmake[1]: *** [check-am] Error 2
11:49:30 gmake[1]: Leaving directory '/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ia64_hpux_11.23/masterfiles/tests/unit'
11:49:30 Makefile:311: recipe for target 'check-recursive' failed
11:49:30 gmake: *** [check-recursive] Error 1
11:49:30 gmake: Leaving directory '/home/jenkins/workspace/testing-enterprise-pr_masterfiles/label/PACKAGES_ia64_hpux_11.23/masterfiles'
```

I would really recommend making this specific to FreeBSD, unless you have the time and equipment to investigate and solve this test failure yourself. Since the package module is for a platform we don't officially support, I'm afraid I cannot invest time in this myself.

modules/packages/freebsd_ports
kacfengine: It should handle it gracefully. ``get-package-data`` should be cheap, and should do as little checking as possible, so it's better to check it in ``repo-install``.

zofrex: Done!

m4/cf3_platforms.m4
zofrex: I fixed the shunit2 detection to work on any POSIX platform which is the immediate problem you were running into, but if someone ran these tests on Solaris with shunit2 on their path, they would fail. I made the test script compatible with just about everything, but the package module itself won't work on Solaris without tweaking, and it really doesn't make sense to write that with anything other than FreeBSD in mind.

I'll rewrite the checks to be FreeBSD && shunit, I guess? (Slightly worried no-one will actually run them because of that requirement)

kacfengine: Well, it depends on whether it's possible to run the test on a different platform. If it is, then surely we can run it in our test farm on a different platform, but since we don't have a FreeBSD build machine it won't run there.

kacfengine:
trigger build

m4/cf3_platforms.m4
zofrex: Outcome from IRC conversation: Worth attempting to get this working cross-platform if it isn't too tricky, so I will take a stab at that using Solaris and then will find someone to check it on HP UX for me.

Failing that, the tests should be limited to platforms they work on (FreeBSD, OS X, Linux?) if shunit2 is installed. Failing _that_ we should only run them on platform=FreeBSD + shunit2 available.

kacfengine: Sounds good!

kacfengine:
@zofrex: The ``SHUNIT`` detection here should work in the same way as in #603, take a look at the commit e531d63bb8e3792c00153a0e939974b07d480129. If you can make sure that in addition to ``shunit2``, the tool ``pkg`` is also detected. Just add it in the same way as was done in the mentioned commit.

nickanderson:
ping @zofrex

BasementTrix:
@nickanderson referred me to this project.  It looks like you're building support for Ports.  I'm trying to work on a module to support pkgng, but have zero knowledge about how to use the CFEngine testing framework.

Maybe we could work together?

nickanderson:
@BasementTrix Also check out https://github.com/cfengine/masterfiles/pull/603 where support for pkgsrc was added.

zofrex:
@BasementTrix I would love to work together! CFEngine would definitely benefit from supporting pkgng.

Are you ever on IRC? That's a much better way of getting hold of me. I should be around tomorrow (I've set aside time to polish off this PR and hopefully get it into a mergeable state), and I'll be online on and off over the coming weeks.

BasementTrix:
On Wed, Dec 09, 2015 at 06:25:31AM -0800, zofrex wrote:
> @BasementTrix I would love to work together!

Glad to hear it!  Hopefully you can help me get to grips with git.
I'm more of an Subversion guy.

> Are you ever on IRC? That's a much better way of getting hold of
> me.

Almost never on IRC, but maybe I should get back to it.  It's filtered
at $DAYJOB, unfortunately, but maybe I can get on in the evenings.

Nick Anderson pointed me at the pull where pkgsrc was submitted.  I've
been using that to work up the unit test script for
$(sys.workdir)/modules/packages/pkg. 

-- 

m4/cf3_platforms.m4
zofrex: Got this working cross-platform without toooo much hassle, but I concede it got a bit uglier. `egrep` is well supported (better than `grep -E`, which requires PATH fiddling in Solaris). `sed -E` is not so I rewrote those statements to not need it. I checked docs for AIX and HP-UX, and tested on Solaris, and I'm pretty confident this is portable. Need help testing though.

If there are strong opinions that the previous form was nicer, I think another elegant option would be to limit the platforms we run the tests on by testing sed / grep's capabilities in the autoconf stuff.

zofrex:
I did a rebase to pull in the shunit detection from #603. I didn't add detection for `pkg` because the tests work fine without it :)

#603 also added mocks for some of the same pkg commands my tests use. I'll merge the two sets of mocks, it'll be cleaner that way.

zofrex:
Tested and working on OS X, Ubuntu, Solaris, and FreeBSD :)

lib/packages.cf
```diff
@@ -94,6 +94,28 @@ body package_module yum
     #default_options => {};
 }
 
+body package_module freebsd_ports
+# @brief Define details used when interfacing with the freebsd ports package
+# module.
+#
+# **Note:** Ports are expected to be setup prior to trying to use the packages
+# promise. You may need to ensure that `portsnap extract` has been run.
```
tzz: Maybe there's a way to check that in `packages.cf` or in the package script itself?

lib/packages.cf
```diff
@@ -94,6 +94,28 @@ body package_module yum
...
+#
+# **Example:**
+# ```cf3
+# bundle agent main
+# {
+#   packages:
+#     freebsd::
```
tzz: What happens if you try to use it on other systems? Does it fail gracefully?

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
+#!/bin/sh -e
+
+command=$1
+
+while read -r line; do
+  export INPUT_$line
+done
+
+list_updates() {
+  pkg version -v -l "<" | sed -e 's/\([^ ]*\)-[^-]* .* \(.*\))/Name=\1\
```
tzz: Maybe add a comment here with an example data line and how it looks after the rewrite, for clarity.

Also this doesn't validate the `pkg` return code.  Should it?

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+Version=\2\
+Architecture=none/'
+}
+
+case $command in
+  supports-api-version)
+    echo 1
```
tzz: I would `exit 0` right here for clarity.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+  ;;
+  get-package-data)
+    name="${INPUT_File?File must be given to get-package-data}"
```
tzz: The error handling may be better sent to `STDERR`.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+    echo PackageType=repo
+    echo Name=$name
+  ;;
+  list-installed)
+    pkg info -f -a | egrep '^(Name|Version|Architecture)' | sed -e 's/[ ]*:[ ]*/=/' -e 's/^Architecture=.*/Architecture=none/'
```
tzz: I would `exit 0` right here for clarity. Also I would make it a standalone function like `list_updates()` and also provide a representative source line and what it becomes, for clarity.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+  ;;
+  repo-install)
+    name="${INPUT_Name?Name must be given to repo-install}"
```
tzz: Also `STDERR` for errors might be more appropriate here. Maybe even `exit 1` because it's a required parameter.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+    version="${INPUT_Version}"
+
+    export BATCH=1
+    PORT_PATH="`whereis -sq \"$name\"`"
```
tzz: Maybe better as `$(whereis -sq "${name}")`

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+
+    if [ -z "$PORT_PATH" ]
+    then
+      echo "ErrorMessage=Could not install $name, port does not exist"
+      exit 0
+    fi
+
+    cd "$PORT_PATH"
```
tzz: I would check to make sure that this succeeded, based on `pwd`

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+
+    if [ "$version" ]
```
tzz: I would make the test explicit, not implied.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+      if [ "$available" != "$version" ]
+      then
+        echo "ErrorMessage=Could not install $name $version, available version was $available"
+        exit 0
+      fi
+    fi
+
+    if `pkg info -e "$name"`
+    then
+      make deinstall reinstall >&2
```
tzz: What happens if this fails?

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+    else
+      make install >&2
```
tzz: What happens if this fails?

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+    fi
+  ;;
+  list-updates)
+    portsnap --interactive fetch update >&2
```
tzz: What happens if this fails?

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+        exit 0
+      fi
+    fi
+
+    if `pkg info -e "$name"`
+    then
+      make deinstall reinstall >&2
+    else
+      make install >&2
+    fi
```
tzz: I'd maybe exit here, so it's clear that you're done.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+  ;;
+  list-updates)
+    portsnap --interactive fetch update >&2
+    list_updates
```
tzz: I'd maybe exit here, so it's clear that you're done.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+  ;;
+  list-updates-local)
+    list_updates
```
tzz: I'd maybe exit here, so it's clear that you're done.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+  ;;
+  remove)
+    name="${INPUT_Name?Name must be given to remove}"
```
tzz: Same error handling note as before, for `STDERR` and maybe `exit 1`

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+    BATCH=1
+    cd `whereis -sq "$name"`
```
tzz: Make sure this succeeds, check `pwd`

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+    make deinstall >&2
```
tzz: What happens if this fails?

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,74 @@
...
+  ;;
+  *)
+  echo "ErrorMessage=Invalid operation"
```
tzz: add `exit 1` maybe

tzz:
My comments are just advice, don't feel obligated to make those changes!

lib/packages.cf
rootwyrm: Yes - two means.
fileexist($PORTSDIR/.portsnap.INDEX) 
fileexist($PORTSDIR/.arcconfig)

modules/packages/freebsd_ports
rootwyrm: Concur - separate function would be better when considering possibility of pkg lock.

modules/packages/freebsd_ports
rootwyrm: There needs to be an explicit test and bailout here in failure scenarios. If make deinstall fails, it can ripple and break the whole system.

lib/packages.cf
zofrex: I doubt it. I'll test what happens with the existing package modules and follow suit.

modules/packages/freebsd_ports
zofrex: pkg return code implicitly checked by -e. Comment sounds like a good idea, will add.

modules/packages/freebsd_ports
zofrex: This will output to STDERR and exit 1 :) http://www.grymoire.com/Unix/Sh.html#uh-31 for reference

modules/packages/freebsd_ports
zofrex: Same as before, that's what happens :)

modules/packages/freebsd_ports
zofrex: Agreed!

modules/packages/freebsd_ports
zofrex: set -e has our back again :)

modules/packages/freebsd_ports
zofrex: Good catch!

modules/packages/freebsd_ports
zofrex: The script will error (assuming make returns non-zero)

modules/packages/freebsd_ports
zofrex: Likewise

modules/packages/freebsd_ports
zofrex: As long as it returns non-zero, the whole script will error out at this point

modules/packages/freebsd_ports
zofrex: Same answer as before :)

modules/packages/freebsd_ports
zofrex: Already covered by set -e, but I think the `$()` suggestion applies here too

modules/packages/freebsd_ports
zofrex: Same as anything else failing

lib/packages.cf
tzz: Maybe that should be noted in the example? I don't know how likely it is.

zofrex: I checked with the #freebsd channel and /usr/ports/Mk/bsd.port.mk is a file we can depend on existing if ports are extracted.

lib/packages.cf
zofrex: I tried out the apt_get package module on FreeBSD and it didn't fail very gracefully. That leaves me less inclined to handle that scenario... if someone tries to use FreeBSD ports on Debian, that's their lookout :P

lib/packages.cf
zofrex: Note added!

modules/packages/freebsd_ports
zofrex: Comment for this added

modules/packages/freebsd_ports
zofrex: Done!

zofrex:
I broke out all non-trivial functionality in the case statement into separate functions - I don't think there's any need for the explicit exit statements now, the structure of the case is a lot more obvious.

Also added example output to the comments and made a few other suggested tweaks.

modules/packages/freebsd_ports
```diff
@@ -0,0 +1,108 @@
...
+  then
+    available=$(make -V PORTVERSION)
+    if [ "$available" != "$version" ]
+    then
+      echo "ErrorMessage=Could not install $name $version, available version was $available"
+      exit 0
+    fi
+  fi
+
+  make deinstall reinstall >&2
```
zofrex: In testing on my own machines I found that "make install" failed sometimes (if a package is removed via "pkg del" instead of "make deinstall" then "make install" does nothing), but "make deinstall" doesn't error if there's nothing to be done, so this worked out the most reliable way of installing or upgrading a port.

lib/packages.cf
kacf: On that note, perhaps `freebsd_ports` should be renamed to simply `ports`. It's more in line with how the others are named and doesn't imply any particular platform. Unless you see it as very unlikely it will ever be used anywhere else.

zofrex: There are other systems also called "ports" (OpenBSD and MacPorts to name two), so I think it's useful to distinguish that this is only for FreeBSD. I don't think this will work anywhere else – as far as I know there's no other OS that uses ports and pkgng (OpenBSD uses the "old-style" pkg_* not pkgng).

tzz:
YES

lib/packages.cf
rootwyrm: `freebsd_ports` is most appropriate; as pointed out, OpenBSD and OSX (also NetBSD) use "ports". `freebsd` is inappropriate due to ports vs `pkgng` vs `portmaster` vs `portshaker`+`portmaster` vs `pkgng+poudriere`.

kacf: Alright, `freebsd_ports` it is!

zofrex:
I've checked the automated tests work on FreeBSD, OS X, Ubuntu, and Solaris.

I've manually tested installing, upgrading, and uninstalling ports on FreeBSD using this + packages promises.

kacf:
trigger build

tests/unit/Makefile.am
```diff
@@ -7,8 +7,9 @@ check_SCRIPTS += \
 	test_package_module_yum
 endif
 
-if HAVE_PKG_INSTALL
 if HAVE_SHUNIT2
+check_SCRIPTS += test_package_module_freebsd_ports
+if HAVE_PKG_INSTALL
```
kacf: Shouldn't this contain a check to make sure ports is available? Or is it enough to check for `shunit2`?

See also https://github.com/cfengine/masterfiles/blob/master/configure.ac#L122-L126 for how this is checked for `pkg_install`.

kacf:
Running the PR through the tests, just to see where we stand.

tests/unit/Makefile.am
zofrex: We only need `shunit2` available to run the tests - all the pkg / ports stuff is stubbed or mocked :)

tests/unit/Makefile.am
```diff
@@ -16,12 +17,14 @@ endif
...
 
 EXTRA_DIST = \
+	harness_freebsd_ports \
 	mock_apt_get \
 	mock_dpkg \
 	mock_dpkg_deb \
 	mock_dpkg_query \
 	mock_pkgin \
 	mock_pkg_info \
+	mock_pkg \
```
kacf: This should be `mock_freebsd_pkg` to match the file name. `make dist` fails without this.

zofrex: Oops, sorry about that! Updated now.

kacf:
trigger build

kacf:
Thanks!

A FreeBSD package module is a popular target apparently, there is both https://github.com/cfengine/masterfiles/pull/617 and https://github.com/cfengine/masterfiles/pull/620. :-)

kacf:
@zofrex, @fbettag, @BasementTrix: Including all relevant people in the discussion:

It seems that there is work of several people going on here that should be consolidated. If I understand correctly the "ports" package method used here is a way to install from source, whereas the "pkg" method used in #617 and #620, deals with binary packages. However, when listing packages they both use the same listing. Is this correct?

If so, we should consolidate these approaches, because the design of CFEngine is to use one, and only one, package module for each namespace of packages. This is important when using for example `packagesmatching()`, otherwise you will get each package twice. Different behavior when it comes to installation can be altered by passing in the `options` attribute in the promise, which will then be passed down to the package module in the variable with the same name. So for example:

```
packages:
  "mypackage"
    policy => "present",
    package_module => freebsd_pkg;
    #options => "" # For binary install (default).

  "mypackage"
    policy => "present",
    package_module => freebsd_pkg,
    options => "ports" # For source based install (will use ports).
```

I'm treating binary installs as default here, since it is typically more common, at least for Linux, but this can be discussed.

Does this make sense? I don't personally deal with FreeBSD, so apologies if I have misunderstood something.

fbettag:
@kacf i'm not really sure that this is the best way to handle it, since almost every function body is different depending on how you use it. @zofrex uses the "make" commands rather than the pkg commands (which is perfectly fine).

My suggestion would be to have 2 separate versions, one "freebsd_ports" (this #575) and one "freebsd_pkg" (combination of my #617 and @BasementTrix #620), since they are 2 different approaches.

This would also allow for a clean implementation of both and the user a fair choice.

BasementTrix:
I think @fbettag is correct.  Installing from packages is a very different from installing via ports.  While "make package" is valid for building a package from a directory in the Ports Tree, not all users will want to take that step when "make install" will suffice.

The pkg module that I am working on doesn't deal with the Ports Tree at all.  It is only for installing binary packages; either from a pre-configured repository or specified by path.

zofrex:
I concur, I think two separate versions is ideal here. In fact I want to throw a third into the mix: source-based package management on FreeBSD using portmaster ;)

@kacf: To your point that cfengine wants only one package module per namespace of packages† (incidentally, can you clarify the meaning of "namespace" here?), does it help the situation that only one of these modules should ever be in use at any given time?

FreeBSD users are strongly discouraged from mixing ports and packages‡ in a single installation, doing so tends to break things unless you are an expert, and even then requires a lot of manual fiddling – so someone doing this is unlikely to be managing their packages with cfengine.

Would it suffice if all these modules came with warnings to use one, or the other, but if you use more than one at once you're on your own?

In fact, the more I think about it, the more I think they _should_ be separate... making this an option, as per your proposal, positively encourages people to mix the two approaches.

† Incidentally, I'm not sure that the design of CFEngine itself is a compelling argument in these contexts. If CFEngine and any package manager we want it to support are not compatible, it seems like CFEngine is the only one of the two that is likely to be able to change.

‡ In this context, "packages" refers to installing binary packages through pkgng, even though installing a port via "make install" does _technically_ create and install a package. This technical subtlety is rarely accounted for in informal conversations about FreeBSD.

kacf:
> FreeBSD users are strongly discouraged from mixing ports and packages‡ in a single installation, doing so tends to break things unless you are an expert, and even then requires a lot of manual fiddling – so someone doing this is unlikely to be managing their packages with cfengine.

Thanks @zofrex, this is an important piece of information and clears things up a bit. Then it's not necessary for these modules to be merged; it sounds like having two versions `freebsd_ports` and `freebsd_pkg` is the best approach, as @fbettag suggests.

kacf:
This also passes all tests I see, so unless you have anything more to add, this is ready to be merged.

rootwyrm:
There's a working portmaster that can be pulled forward pretty easily from 3.5.x masterfiles, as I recall. (May still be there in 3.6, I didn't think to check.) Note that portmaster should be handled separate from ports, as it behaves differently.

Mixing ports and portmaster, a-okay. `ports` and `pkg`, *nope.* `ports` and `pkgng`, *much nope.* This breaks the solver frequently. I believe that long term, there should be a mutual exclusion mechanism that prevents mixing for safety. `pkgng` may take undesired actions during solver problems and can be triggered incidentally in rare scenarios. I don't think it's critical though - folks running CFE on FreeBSD know not to mix (or how to mix.)

Also remember that any instances of old `pkg` (inclusive `pkg_*`) need to be **removed** as unsupported by upstream. I would vastly prefer (as would upstream maintainers I'm sure) that `pkg` as referenced here should be referred to as `pkgng` in the *documentation* to prevent confusion with `pkg_*`. Otherwise looks pretty much perfect to me.

BasementTrix:
On Tue, Dec 22, 2015 at 06:51:08AM -0800, Phil Jaenke wrote:
> Also remember that any instances of old `pkg` (inclusive `pkg_*`)
> need to be **removed** as unsupported by upstream. I would vastly
> prefer (as would upstream maintainers I'm sure) that `pkg` as
> referenced here should be referred to as `pkgng` in the
> *documentation* to prevent confusion with `pkg_*`. Otherwise looks
> pretty much perfect to me.

The removal of the pkg_* tools is what lead me to create the pkg
module that I submitted with #620.  I agree with referring the the
module as "pkgng" in the documentation, but it is important for
non-FreeBSD users to know that there is no executable called "pkgng".

zofrex:
Rebased and squashed, I think this is good to go :)

kacf:
trigger build

kacf:
And merged! This will appear in CFEngine 3.9. Great work, @zofrex!

zofrex:
Thank you! And thanks to you and everyone else who gave help & feedback... couldn't have made my first CFEngine contribution without it :)

PR #579, opened 2015-10-23
Change: Improve git drop user support

nickanderson:
When using bundle agent git from the stdlib on repositories that are not
owned by root an error prevents action because cfengine does not inherit
the shell environment of the user it is executing as. We now set the
HOME variable when running git commands with dropped privledges.

@cdituri you might want to check this out.

### DIFF
--- a/lib/paths.cf
+++ b/lib/paths.cf
@@ -107,6 +107,7 @@ bundle common paths
       "path[lsattr]"        string => "/usr/bin/lsattr";
       "path[tar]"           string => "/bin/tar";
       "path[pgrep]"         string => "/usr/bin/pgrep";
+      "path[getent]"        string => "/usr/bin/getent";
 
     aix::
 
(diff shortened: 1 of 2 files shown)


nickanderson:
@tzz ping for thoughts

cdituri:
@nickanderson - nice looks good. Cool that you added this to the common ``git`` bundle so that the specific``git_.*`` bundles inherit the extra functionality, +1!

nickanderson:
@cdituri are able to test it easily? Mostly looking for your approval since you were the original contributor.

cdituri:
@nickanderson - I've got a pipeline that I can run this through but will need to wait til the weekend before I can begin.

nickanderson:
That's great!

nickanderson:
ping @cdituri just checking in to see if this change works for you as well

PR #580, opened 2015-10-23
Add: New `results` classes body [Redmine#7481]

nickanderson:
Ref: https://dev.cfengine.com/issues/7418

Changelog: Title


tzz:
YES

@peckpeck @jooooooon if the default bodies PR is merged with the "any" option to apply to any promise type, we can do something pretty amazing (connected to my suggestion of automatically defining results):

```
body classes any_classes_default
{
    inherit_from => results("bundle", $(this.promise_handle));
}
```

...and suddenly any promise can auto-define classes.

nickanderson:
;)

PR #581, opened 2015-10-28
Change: Seperate binary details from policy update

nickanderson:
Ref: https://dev.cfengine.com/issues/7662


nickanderson:
trigger_build

PR #596, opened 2015-11-03
Change: Make sure that 3.7 lib dir gets deployed

nickanderson:
Without this directory being deployed 3.7 agents running new policy will
error when running the policy.

Ref: https://dev.cfengine.com/issues/7680


kacfengine:
Shouldn't this be in master in general? Or do we want to get rid of it there?

nickanderson:
@kacfengine yep I just started in the wrong branch, i picked it back to master
https://github.com/cfengine/masterfiles/pull/597

PR #602, opened 2015-11-10
Disable faulty test, it's been failing randomly in Jenkins buildslaves.

jimis:


jimis:
Merging, @kacfengine should we cherry-pick?

PR #569, opened 2015-09-26
Add pkgsrc package module. See http://www.pkgsrc.org/

bahamat:
This module primarily supports NetBSD and SmartOS. By extension, it supports any platform that pkgsrc itself supports.

PR #603, opened 2015-11-12
Add pkgsrc package module. See http://www.pkgsrc.org/

kacfengine:
Supersedes https://github.com/cfengine/masterfiles/pull/569.


kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

bahamat:
Excellent! Will this also be merged into 3.8.x and 3.7.x branches?

kacfengine:
I'm afraid not. :-( We have a strict policy of not integrating new features in patch series. I know it sucks a bit, but we would rather not make exceptions to this rule.

So people will have to fetch it manually for now, but it's quite easy to drop into the modules directory, and 3.9 is not *that* far away, should be here somewhere in June.

Thanks for the work!

bahamat:
I find this pretty frustrating since I busted my ass to get it ready for 3.8 and this PR sat for weeks with no feedback and no activity. And after several attempts to move things along I finally pressed the issue and it was merged in a couple of days, but not before 3.8 was cut. And so now it will again sit for *nearly a year* before release.

kacfengine:
Yes, I know. Believe me, we are well aware of the capacity problems when it comes to pull requests, it is an unfortunate, but very real resource problem. We have discussed internally how we can handle pull requests if not faster, then at least more predictably, but it's still an ongoing discussion.

And to be clear, 3.9 is six months away, not a year.

bahamat:
June is 7 months away, plus the month that it has already sat.

PR #607, opened 2015-11-12
Let testall know about cf-runagent for future tests.

jimis:


jimis:
This has been waiting for too long, merging.

PR #608, opened 2015-11-23
Redmine #7737: Error while reading cf-consumer.pid file.

pasinskim:
Fix error messages while reading file during bootstrap.


pasinskim:
This was already verified and discussed with @nickanderson

PR #610, opened 2015-11-24
Fix broken acceptance tests.

jimis:


jimis:
I have no clue why, but this change fixes the broken acceptance test in my pull request https://github.com/cfengine/core/pull/2416, plus it looks more correct in general.

jimis:
trigger build

jimis:
trigger build

jimis:
This is mostly green, so merging.

Only one spurious failure happened on RHEL4, but this is a prexisting acceptance test glitch that will be fixed separately.

PR #612, opened 2015-11-25
blah

jimis:


jimis:
Dammit, travis rejected my valuable commit! :-)

PR #617, opened 2015-12-08
Added new generation package_module for FreeBSD.

fbettag:
This script adds support for the new package promise on FreeBSD. It also fixes https://dev.cfengine.com/issues/7421.

Thanks for your consideration.

### DIFF
--- a/lib/packages.cf
+++ b/lib/packages.cf
@@ -57,6 +57,9 @@ bundle common package_module_knowledge
     debian::
       "platform_default" string => "apt_get";
 
+    freebsd::
+      "platform_default" string => "pkgng";
+
     redhat::
       "platform_default" string => "yum";
 }
@@ -86,6 +89,14 @@ body package_module pkgsrc
     query_installed_ifelapsed => "60";
     query_updates_ifelapsed => "14400";
 }
+
+body package_module pkgng
+{
+    query_installed_ifelapsed => "60";
+    query_updates_ifelapsed => "1440";
+    #default_options => {};
+}
+
 
 body package_module yum
 {
--- a/modules/packages/pkgng
+++ b/modules/packages/pkgng
@@ -1,0 +1,118 @@
+#!/bin/sh
+
+## Licensed under:
+## MIT Public License
+## http://www.opensource.org/licenses/MIT
+
+## Copyright (c) 2015, Franz Bettag <franz@bett.ag>
+
+## pkgng package module for cfengine
+
+# Set up mock environment if necessary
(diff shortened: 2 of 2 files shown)


tzz:
nice!

kacf:
I have little experience with FreeBSD, so I may be wrong about this, but doesn't this module step on the toes of https://github.com/cfengine/masterfiles/pull/575? Both use the `pkg` command.

fbettag:
@kacf: Yes and no. :)
#575 apparently leverages [portsnap](https://github.com/zofrex/masterfiles/commit/e30c55f2bf84f058f569c5ec16885a4227671236#diff-64bdd1f24ebd3cd695826a79c30e78e3R50) which is the source variant of packaging. #575 takes care of downloading Makefiles and build packages using them and uses "pkg" to install/list/deinstall them. It **doesn't** download (for example) cfengine's Makefile, but a Makefile which describes to FreeBSD ports-system how to handle CFengine's configure/Makefile/etc. In short: It deals with Wrapper-Scripts that result in compiled packages.

My version takes care of the binary backend, where no "Makefiles" get downloaded, just plain and simple binary packages that are prebuilt (by either yourself or the official repo). This also uses "pkg" but a bit more extensive than #575.

The difference is also in naming. #575 uses "freebsd ports" (which is the name for the source form of 3rd party software), while mine is "pkgng" which only takes care of pre-built binaries.

kacf:
Ok, thanks for explaining. It sounds like it's natural that they live side by side.

But what about #620? That one looks very similar to this one.

fbettag:
@kacf Yes, they live side by side, giving the admin the option to either build packages specific to their needs on their boxes (like #575 offers) or prebuild their own/use stock packages (like we do).

Yes #620 is conflicting with this one, does the same as mine but i am not qualified enough to judge which one is better from the cfengine standpoint. (sure hope its mine lol)

fbettag:
Also to make it more clearer, they don't live "together". They are the same thing "pkg".
pkg is like the dpkg command. It takes care of the database of installed packages and all operations around it.

Portsnap (as used in #575) just has the extra advantage of using source-based-packages (from which all binary packages get built) as different source. Both ways end up installing their packages through "pkg". It's just about how you get from "i want package X" to the resulting binary package which gets installed. :)

kacf:
> Yes #620 is conflicting with this one, does the same as mine but i am not qualified enough to judge which one is better from the cfengine standpoint.

I will have a look at this later. But this point may become moot. See additional discussion in #575.

kacf:
Closing this in favor of #620.

PR #620, opened 2015-12-09
Initial commit for FreeBSD pkg support

BasementTrix:
New to git and shunit2.  Could probably use a hand with testing, but trying to follow "commit early and often".

### DIFF
--- a/modules/packages/pkg
+++ b/modules/packages/pkg
@@ -136,7 +136,18 @@ get_package_data() {
         if [ "${TYP}" = 'repo' ] ; then
             PKGNAM=`basename "${PKG}" | sed -r -e 's/^([A-Za-z0-9_-]+)-[0-9a-z.,_]+/\1/'`
             PKGVER=`basename "${PKG}" | sed -r -e 's/^[A-Za-z0-9_-]+-([0-9a-z.,_]+)/\1/' -e 's/\.(tgz|(tar\.)?gz|txz|zip)$//'`
+            if [ -z "`echo ${PKGVER} | grep -E '[0-9]'`" ] ; then 
+                if [ "${PKGNAM}" != "${PKGVER}" ] ;  then
+                    PKGNAM="${PKGNAM}-${PKGVER}"
+                fi
+                PKGVER=''
+            fi
             echo "Name=${PKGNAM}"
+            # Returning version for PackageType=repo is not advised
+            # https://docs.cfengine.com/latest/reference-standard-library-package_modules.html
+            # if [ -n "${PKGVER}" ] ; then 
+            #     echo "Version=${PKGVER}"
+            # fi
         else
             pkg query --file "${PKG}"  'Name=%n\nVersion=%v\nArchitecture=%q' | \
                                       sed -e 's/=freebsd:[0-9]*:/=:/' \
--- a/modules/packages/pkg
+++ b/modules/packages/pkg
@@ -173,6 +173,7 @@ case "${CMD}" in
         fi
         pkg version ${LOCALUPDATE} -vl'<' | \
             awk 'BEGIN { cmd="uname -m"; cmd | getline arch} 
+/ repositor(y|ies) / { next; }
 { 
   pkgname = $1; 
   pkgver  = $7; 
--- a/modules/packages/pkg
+++ b/modules/packages/pkg
@@ -112,7 +112,7 @@ process_inputs() {
 get_package_data() {
     if [ -n "${PKG}" ] ; then
         case "${PKG}" in
-            /*|./*|../) # File Path.  It's a file
+            /*) # File Path.  It's a file
                 TYP='file'
                 if [ ! -e "${PKG}" ] ; then
                     fatal "No such file: '${PKG}'."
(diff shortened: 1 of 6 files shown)


kacf:
I have little experience with FreeBSD, so I may be wrong about this, but doesn't this module step on the toes of #575? Both use the `pkg` command.

modules/packages/pkg
```diff
@@ -0,0 +1,220 @@
...
+                ;;
+            http://*|ftp://|file://) # URL, it's a file.
+                TYP='file'
+                ;;
+            */*)
+                if [ -e "${PKG}" ] ; then
+                    TYP='file'
+                else
+                    TYP='repo'
+                fi
```
kacf: I would not return `file` for a path which isn't absolute. Testing for existence is unpredictable, the appearance of a file might suddenly change CFEngine's idea of what a promise string is (file or repo). I think it's better to test only the "absolute-ness" of the path and that way you'll get a consistent result every time.

This applies to both this `case` entry (`*/*`), and the one above (`/*|./*|../`).

kacf:
This pull request has become messed up, it seems to be a mix of several branches, including a conflicting copy of the Ports package module, as well as @fbettag's https://github.com/cfengine/masterfiles/pull/617. It should be rebased so that there is a clean diff in the pull request.

Other than that, I have not reviewed it in depth since this is FreeBSD. Perhaps it's better that @fbettag and @BasementTrix review each other? There certainly couldn't be a better time, it's quite amazing that we managed to get two independent contributions for the same package module at the same time, the same exact day even!

One definite advantage with this pull request over #617 is that this one has tests, so in that sense I consider this one in the lead currently.

fbettag:
Since my #617 was put together in a few hours and this one seems more complete (due to tests), i think that this one here (#620) should be taken. One addition i'd like to request, which i haven't implemented, would be the feature to use options so that one could use repository=foobar and have pkg install -r foobar myPackage, that way one can specify excatly which repo (in case one has multiple).

modules/packages/pkg
BasementTrix: On Mon, Dec 28, 2015 at 01:37:36AM -0800, Kristian Amlie wrote:
> I would not return `file` for a path which isn't absolute. Testing
> for existence is unpredictable, the appearance of a file might
> suddenly change CFEngine's idea of what a promise string is (file or
> repo). I think it's better to test only the "absolute-ness" of the
> path and that way you'll get a consistent result every time.
> 
> This applies to both this `case` entry (`*/*`), and the one above
> (`/*|./*|../`).

You're right that "file" should never be returned.  I need to rethink
that logic.  A slash in the middle can still be valid, though, as a
"repo" entry.  In FreeBSD, both "bash" and "shells/bash" would be
valid names to install the package for the Bourne Again Shell.

I was trying to cover all possible cases and got a little confused, I
guess.  Will fix.

-- 

BasementTrix:
On Mon, Dec 28, 2015 at 01:45:13AM -0800, Kristian Amlie wrote:

> 
> This pull request has become messed up, it seems to be a mix of
> several branches, including a conflicting copy of the Ports package

My primary repository is SVN, but will get this cleaned up soonest;
hopefully this evening.

-- 

BasementTrix:
On Mon, Dec 28, 2015 at 02:20:33AM -0800, Franz Bettag wrote:
> One addition i'd like to request, which i haven't
> implemented, would be the feature to use options so that one could
> use repository=foobar and have pkg install -r foobar myPackage, that
> way one can specify excatly which repo (in case one has multiple).

This is in my tree already.  I'll see if I can add a test for it in
the test_package_module_pkg script, as well, when I'm cleaning up my
tree.  

-- 

BasementTrix:
On Mon, Dec 28, 2015 at 05:21:21PM -0600, Trix Farrar wrote:
> My primary repository is SVN, but will get this cleaned up soonest;
> hopefully this evening.

Or so I thought.  I could use some help/advice from someone, please.

I have cleaned up my SVN tree and committed it to github, but the site
says I still have conflicts.  I've gotten the difference between the
CFEngine tree and the BasementTrix tree down to the attached diff, but
I have no clue how to do the git magic to make the site happy.

I have the devel/git package installed (git-2.6.4) but have no idea
what to do with it.

-- 

kacf:
I tried to clean up your branch a bit, here: https://github.com/kacf/masterfiles/commits/trix_package_module. Try doing this:

```
git remote add kacf git@github.com:kacf/masterfiles.git
git fetch kacf
git reset --hard kacf/trix_package_module
```

Make sure you don't have any pending changes before doing this. If so, you should store them somewhere else first.

Note that the commit messages are still all messed up, this will need to be cleaned, but at least you should be able to work on it.

kacf:
And if you update this branch you'll need to force push: `git push -f`

BasementTrix:
On Wed, Dec 30, 2015 at 01:09:27AM -0800, Kristian Amlie wrote:
> And if you update this branch you'll need to force push: `git push -f`

Thank you very much for your help.  I could not have done it without
you.  The pull request now appears to be without conflicts.

-- 

kacf:
trigger build

kacf:
trigger build

kacf:
Alright, squashed and merged! Thanks for covering FreeBSD!

This will be in 3.9.0.

PR #624, opened 2015-12-16
services: allow explicit use of standard_services

teroka:
This will add a minimalist `standard_services` method.

I have multiple service libraries and wanted to be explicit about which one I use. Alas, I couldn't do that with the `standard_services`. Only way to use it is to omit the `service_method` entirely. Attempting to use it directly and you end up with:

* `error: Undefined body standard_services with type service_method`

Don't know if you wan't this in libs as it can be confusing for users. This would imply that you could change the method in libs and reap the benefits even if you don't use `service_method => standard_services` explicitly. Didn't test that tho, might even work like that ;)

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -97,9 +97,11 @@ body service_method standard_services
 # By default this service_method is *not* used. The call for standard_services
 # is within the core and not here. In case you use a promise like:
 #
-## services:
-##     "ssh"
-##       service_policy => "start";
+# ```cf3
+# services:
+#     "ssh"
+#       service_policy => "start";
+# ```
 #
 # Then this method is skipped and CFEngine calls standard_services bundle
 # directly. This is here as a helper in case you wan't to be explicit
@@ -118,7 +120,7 @@ body service_method standard_services
 #       service_method => standard_services;
 # ```
 {
-      service_bundle => standard_services("$(this.promiser)","$(this.service_policy)");
+        service_bundle => standard_services( $(this.promiser), $(this.service_policy) );
 }
 
 ##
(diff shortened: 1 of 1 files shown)


tzz:
I think this is OK, but it should have some docs explaining why it's not normally necessary, and an example of usage like the other bodies.

teroka:
some docs added and squashed.

lib/services.cf
```diff
@@ -91,6 +91,36 @@ body service_method force_deps
...
+# **Example:**
+#
+# ```cf3
+# services:
+#     "ssh"
+#       service_policy => "start",
+#       service_method => standard_services;
+# ```
+{
+      service_bundle => standard_services("$(this.promiser)","$(this.service_policy)");
```
tzz: Don't quote these, it's not necessary.

tzz:
Looks great with one minor comment.

lib/services.cf
teroka: I don't mind, but they are quoted in the reference example.
https://docs.cfengine.com/lts/reference-promise-types-services.html

```
body service_method service_test
    {
      service_bundle => non_standard_services("$(this.promiser)","$(this.service_policy)");
    }
```

tzz: Yup. A lot of older CFE code quotes a lot of variables unnecessarily. It's just a matter of style because it ends up in a string regardless.

nickanderson:
It does look good, please see: https://github.com/teroka/masterfiles/pull/1

nickanderson:
lgtm, thanks!

PR #625, opened 2015-12-16
Add guard for binary upgrade during bootstrap

nickanderson:


nickanderson:
This really needs https://github.com/cfengine/core/pull/2441/files in order to be useful.

PR #629, opened 2015-12-22
cfruncommand parameters are no longer hardcoded in server_common.c.

jimis:


jimis:
Accompanies Core pull request https://github.com/cfengine/core/pull/2450.

controls/cf_serverd.cf
```diff
@@ -38,19 +38,19 @@ body server control
...
     aix::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)";
+      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent) -I -D cfruncommand";
 
     solaris::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)"; 
+      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent) -I -D cfruncommand";
 
     !(windows|hpux|aix|solaris)::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)";
```
jimis: I'm wondering if these escaped quotes are wrong all this time, shouldn't they be around both commands? What do you think @kacf @estenberg ?

estenberg: To be honest I think that this was not thought through, but rather grown organically.

I do think it makes sense to run both update and promises with -I because when you run cf-runagent you do it because you want to trigger a change *now* and you want to see if that change happened or not.

The -Dcfruncommand can be kept for backward compatibility where it is, but I doubt that people rely on this (have not seen it used at least).

estenberg: I would also remove "$(def.cf_runagent_shell)". If we are running cf-agent by default, why do we need to run it in a shell? I can't see how this could break compatibility (?).

If you run a custom command, then perhaps it makes sense, but then you would modify this line anyway.

nickanderson: If I recall correctly, cf-runagent did not used to allow the running of the update policy in order to avoid potential DOS situation. But from user feedback it's always been noted that they would want the update policy to run first, so they can be running the most recently available policy.

jimis: shell is important because running `cmd blah ; cmd blue` is only possible under shell, since the semicolon is a shell interpreted character.

But as I mentioned I think this line has been misquoted for some time. I'm manually testing now several cfruncommands and honestly I think something has been broken for quite some time.

BTW I'm all for removing completely "-Dcfruncommand", it is indeed unused from us at present, and it hasn't been there for long either.

And "-I" parameter was

nickanderson:
-D cfruncommand is indeed useful, I recently proposed a similar class to be defined when running update policy from bootstrap. It would be better in my opinion if the class defined were baked into the C and just documented. But it can be useful to know the provenance of the policy run. There may be parts of policy that you want to skip, or only activate when the agent is invoked via runagent.

controls/cf_serverd.cf
jimis: Heh, my tests show that it needs the following change to run the second command, else it is skipped:
```
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent) -I -D cfruncommand";
+      cfruncommand => "$(def.cf_runagent_shell) -c \'$(sys.cf_agent) -f $(sys.update_policy_path) ; $(sys.cf_agent) -I -D cfruncommand\'";
```

This is another side-effect of having `$(sys.cf_agent)` always enclosed in double quotes.

I'll be adding this change in separate commit.

jimis: BTW do you have any clue about why we are using semicolon `;` instead of double ampersands `&&` as command separators? IMHO a failure in the first command should skip the second.

nickanderson: No, we should be using the semicolon. If we fail to update the policy, we still expect to run the policy that is currently cached.

jimis:
Thanks for the feedback, it was valuable. Can you please review the last change?

jimis:
BTW it's manually tested on linux and seems OK.

controls/cf_serverd.cf
```diff
@@ -38,19 +38,10 @@ body server control
       denybadclocks         => "false";
 
     windows::
-      cfruncommand => "$(sys.cf_agent) -f \"$(sys.update_policy_path)\" & $(sys.cf_agent)";
+      cfruncommand => "$(sys.cf_agent) -I -f \"$(sys.update_policy_path)\"  &  $(sys.cf_agent) -I -D cfruncommand";
```
nickanderson: `&` is different on windows I guess, always trips me up.

controls/cf_serverd.cf
```diff
@@ -38,19 +38,10 @@ body server control
...
-    aix::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)";
-
-    solaris::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)"; 
-
-    !(windows|hpux|aix|solaris)::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)";
+    !windows::
+      cfruncommand => "$(def.cf_runagent_shell) -c \'$(sys.cf_agent) -I -f $(sys.update_policy_path)  ;  $(sys.cf_agent) -I -D cfruncommand\'";
```
nickanderson: Oh lets be sure to define `cfruncommand` class during the update policy as well (windows too)

I would probably switch to using the long options for readability. Other than that seems OK to me (though I use runagent almost never, so I might be overlooking things).

jimis:
Thanks @nickanderson, `cfruncommand`is now set on both `update.cf` and regular invocations, and it has been beautified a bit. Tested on Linux again.

controls/cf_serverd.cf
```diff
@@ -38,19 +38,14 @@ body server control
       denybadclocks         => "false";
 
     windows::
-      cfruncommand => "$(sys.cf_agent) -f \"$(sys.update_policy_path)\" & $(sys.cf_agent)";
+      cfruncommand => "$(sys.cf_agent) -I -D cfruncommand -f \"$(sys.update_policy_path)\"  &
+                       $(sys.cf_agent) -I -D cfruncommand";
```
nickanderson: Looks nice

controls/cf_serverd.cf
```diff
@@ -38,19 +38,14 @@ body server control
...
-    solaris::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)"; 
-
-    !(windows|hpux|aix|solaris)::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)";
+    !windows::
+      cfruncommand => "$(def.cf_runagent_shell) -c \'
+                           $(sys.cf_agent) -I -D cfruncommand -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I -D cfruncommand\'
+                      ";
```
nickanderson: Looks good

kacf: Yep, definitely better, the old quotation was just bad.

jimis:
Argh! Not so easy. This is definitely the correct quotation (the previous one wasn't working) but extra arguments are not passed as they should. For example:
```
# cf-runagent -H 192.168.122.119 --remote-bundles blah
192.168.122.119> cf-serverd executing cfruncommand: /bin/sh -c '
                           "/var/cfengine/bin/cf-agent" -KI -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                           "/var/cfengine/bin/cf-agent" -KI -D cfruncommand'
                       --bundlesequence blah
```
As you can see, the appended arguments (in this example "--bundlesequence") are outside the `sh -c` argument.

jimis:
After lots of testing, it seems there is a good reason to leave the closing quote open... Added a comment so that we remember.

Will merge unless somebody objects, this seems to work right in my tests.

controls/cf_serverd.cf
```diff
@@ -38,19 +38,15 @@ body server control
...
-
-    solaris::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)"; 
-
-    !(windows|hpux|aix|solaris)::
-      cfruncommand => "$(def.cf_runagent_shell) -c \"$(sys.cf_agent) -f $(sys.update_policy_path)\" ; $(sys.cf_agent)";
+    !windows::
+      # last single quote in cfruncommand is left open, so that
+      # arguments (like -K and --remote-bundles) are properly appended.
+      cfruncommand => "$(def.cf_runagent_shell) -c \'
```
nickanderson: Good call on the comment!

PR #632, opened 2015-12-23
lib/bundles.cf: add collect bundle similar to cmerge

tzz:
@nickanderson pretty simple and clean, saves a step over `cmerge()`.

### DIFF
--- a/lib/bundles.cf
+++ b/lib/bundles.cf
@@ -382,6 +382,80 @@ bundle agent cmerge(name, varlist)
       "$(name)_str" string => format("%S", $(name)),    policy => "free";
 }
 
+bundle agent collect_vars(name, tag, flatten)
+# @brief bundle to collect tagged variables into a data container
+# @param name the variable name to create inside `collect_vars`
+# @param tag the tag regex string to match e.g. "beta,gamma=.*"
+# @param flatten to flatten variable values, set to "any" or "true" or "1"
+#
+# The result will be a map in `collect.$(name)`. You can also use
+# `cmerge.$(name)_str` for a string version of the merged containers
+# (if it fits in a CFEngine string).
+#
+# The name is variable so you can call this bundle for more than one
+# collection.
+#
+# Every found variable will be a key in the map, unless you specify
+# `flatten`, in which case they'll be flattened into a top-level array
+# of data.
+#
+# The `flatten` parameter can be "any" or "true" or "1" to be true.
+#
+# **Example:**
+#
+# ```cf3
+# body common control
+# {
+#       inputs => { "$(sys.libdir)/stdlib.cf" };
+# }
+#
+# bundle agent main
+# {
+#   vars:
+#       # the "mymerge" tag is user-defined
+#       "a" data  => parsejson('{ "mark": "burgess" }'), meta => { "mymerge" };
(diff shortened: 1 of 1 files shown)


lib/bundles.cf
```diff
@@ -382,6 +382,47 @@ bundle agent cmerge(name, varlist)
...
+#
+# **Example:**
+#
+# ```cf3
+# bundle agent run
+# {
+#   vars:
+#       # the "mymerge" tag is user-defined
+#       "a"       data  => parsejson('{ "mark": "b" }'), meta => { "mymerge" };
+#       "b"       data  => parsejson('{ "volker": "h" }'), meta => { "mymerge" };
```
nickanderson: Can we come up with a better example that is more sysadmin oriented? Maybe we ports that should be open?
If not, then lets make the names more specific: for example instead of b and h use the full last names, and instead of collect.all maybe collect.last_names?

lib/bundles.cf
```diff
@@ -382,6 +382,47 @@ bundle agent cmerge(name, varlist)
       "$(name)_str" string => format("%S", $(name)),    policy => "free";
 }
 
+bundle agent collect(name, tag)
```
nickanderson: Since we are collecting values, do you think it would make more sense to call the bundle collect_values ?

lib/bundles.cf
tzz: OK; done.

lib/bundles.cf
tzz: Yes, but I went further... take a look.

tzz:
* renamed to `collect_vars`
* can either merge each variable separately OR flatten them all
* docs show two examples of usage (merge vs. flatten)

lib/bundles.cf
```diff
@@ -382,6 +382,72 @@ bundle agent cmerge(name, varlist)
...
+#
+# bundle agent main
+# {
+#   vars:
+#       # the "mymerge" tag is user-defined
+#       "a"       data  => parsejson('{ "mark": "burgess" }'), meta => { "mymerge" };
+#       "b"       data  => parsejson('{ "volker": "hilsheimer" }'), meta => { "mymerge" };
+#
+#   methods:
+#       "go" usebundle => collect_vars("all", "mymerge", ""); # merge a, b into container collect_vars.all
```
nickanderson: nitpick, can you break the long line to fit inside 80 chars?

lib/bundles.cf
```diff
@@ -382,6 +382,72 @@ bundle agent cmerge(name, varlist)
...
+      "flatten" expression => strcmp($(flatten), "any");
+      "flatten" expression => strcmp($(flatten), "1");
+      "flatten" expression => strcmp($(flatten), "true");
+
+  vars:
+      "todo_$(name)"    slist => variablesmatching(".*", $(tag));
+
+    !flatten::
+      "$(name)" data => parsejson('{}'),            policy => "free";
+      "$(name)" data => mergedata($(name), '{ "$(todo_$(name))": $(todo_$(name)) }'), policy => "free"; # iterates!
```
nickanderson: another long line nitpick

tzz:
OK, long lines broken. One is at 84 but I think it's a LOT clearer on a single line so I left it.

nickanderson:
lgtm

nickanderson:
THANKS!

PR #633, opened 2015-12-28
Change: Guard policy parsing of augments_file

nickanderson:
So that variables defined from the augments_file wins, specific
variables that are defaulted in bundle agent def.cf have been guarded
with isvariable. Additionally if the agent is capable of parsing the
augments file directly, skip doing the work.

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -7,6 +7,9 @@
 bundle common def
 {
   classes:
+    # If the augments_file is parsed from C then we do not need to do this work
+    # from policy.
+    !feature_augments::
       # all override classes are defined true
       "$(override_classes)" expression => "any", meta => { "override" };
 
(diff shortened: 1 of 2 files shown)


tzz:
You've mixed changing the defaults with changing the operational mode. Maybe do it in two separate commits?

tzz:
Looks good to me. Maybe add some reports to tell the user what's up.

nickanderson:
@tzz now with more commits

nickanderson:
ping @kacf

nickanderson:
additional notes: https://dev.cfengine.com/issues/7453#note-61

controls/def.cf
```diff
@@ -315,6 +333,10 @@ bundle common def
       ifvarclass => "$(augments_classes_data_keys)";
 
       "$(const.t) $(defvars) = $($(defvars))";
+      "DEBUG $(this.bundle): Agent parsed augments_file"
+        ifvarclass => "have_augments_file.feature_augments";
+      "DEBUG $(this.bundle): Policy parsed augments_file"
```
tzz: Maybe add a report for the case where we do nothing because `feature_augments` is defined? Not sure...

nickanderson: I think this is enough, it would take a fair amount of policy to do all the exact reports about each variable skipping definition because it was already defined and had an augments_file tag.

tzz:
Can you please use `feature_def_json_preparse` instead of `feature_augments`? I'm removing the latter.

nickanderson:
Really?  It was the one I liked the most. But I don't have a strong opinion,  and I can appreciate the specificness of it, so I will switch of you still want.

tzz:
The name change was requested in https://github.com/cfengine/core/pull/2297 so let's be consistent. I know it's verbose but users will generally not have to use it...

nickanderson:
ping @tzz as requested

tzz:
YES

kacf:
Cherry-picked to 3.8.x as well.

PR #637, opened 2015-12-30
Change: Tidy some long lines and add changelog

nickanderson:


lib/bundles.cf
```diff
@@ -439,21 +439,29 @@ bundle agent collect_vars(name, tag, flatten)
...
     !flatten::
-      "$(name)" data => parsejson('{}'),            policy => "free";
+      "$(name)"
+        data => parsejson('{}'),
+        policy => "free";
 
       # this iterates!
-      "$(name)" data => mergedata($(name), '{ "$(todo_$(name))": $(todo_$(name)) }'),
+      "$(name)"
+        data => mergedata($(name), '{ "$(todo_$(name))": $(todo_$(name)) }'),
```
tzz: This is exactly the one I tried to keep on a single line because it's much clearer that way (since we iterate on `$(name)`).

nickanderson: Heh, I thought it read cleaner this way. I never use policy => free anymore since vars have been free by default since 3.3 era

PR #638, opened 2016-01-15
Add: nimclient package module for AIX

nickanderson:
This module provides basic functionality for using nimclient as a means
to ensure packages are either present or absent. It does not support
listing package updates available or provide any special caching.

### DIFF
--- a/modules/packages/nimclient
+++ b/modules/packages/nimclient
@@ -1,0 +1,191 @@
+#!/bin/sh
+
+## Licensed under:
+## MIT Public License
+## http://www.opensource.org/licenses/MIT
+
+## Copyright (c) 2015, Nick Anderson <nick@cmdln.org>
+
+## nim package module for cfengine
+##  - Based on work done in pkgsrc module
+## Installs packages using nimclient
+##  - Probably need to implement caching layer for nimclient, it can be slow
+##    and always requires remote connection
+## Removes packages using installp
+
+export PATH=/opt/local/bin:/opt/local/sbin:$PATH
+
+LEVEL=0
+
+fatal () {
+    echo "ErrorMessage=$@"
+    exit 2
+}
+
+warn () {
+    [ $LEVEL -gt 0 ] && echo "[TRACE]: $*" >&2
+}
+
+supports_api_version () {
+    echo 1
+}
+
+repo_install () {
+    # nimclient needs to know which lpp_source to install from.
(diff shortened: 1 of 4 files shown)


nickanderson:
ping @kacf PR as requested.

modules/packages/nimclient
```diff
@@ -0,0 +1,193 @@
...
+  if [ -n "$options" ]; then
+    eval $options
+  fi
+done
+
+# Set up mock environment if necessary
+# This is not well developed as I don't have continuous access to aix and nim
+# nor am I an expert
+CFENGINE_TEST_NIMCLIENT_MOCK=false
+if [ -n "$CFENGINE_TEST_PKGSRC_MOCK" ]; then
```
kacf: These two variables are different, but I assume they should be the same? In that case the `if [ -n` check works only by accident, you should probably correct this to:

```
CFENGINE_TEST_NIMCLIENT_MOCK=false
if [ "$CFENGINE_TEST_NIMCLIENT_MOCK" = "true" ]; then
```

modules/packages/nimclient
```diff
@@ -0,0 +1,193 @@
...
+    list_installed_packages() {
+        cat ./mock_lslpp_Lc
+     }
+    nimclient_showres() {
+        # This lists the AVAILABLE packages in the nim repo
+        cat ./mock_nimclient_showres
```
kacf: The mock files are not in the correct folder for this to work. It should be: `cat ../../tests/unit/mock_lslpp_Lc` and `cat ../../tests/unit/mock_nimclient_showres`.

kacf:
Looks good apart from the comments.

You might also want to think about turning those mock files into a unit test that runs automatically. It can simply be a shell script that calls the package module with the mock environment variable set, and checks for output. But this is very platform specific, so I won't hold it back if you don't have the bandwidth for that right now. Just let me know.

modules/packages/nimclient
nickanderson: ack

modules/packages/nimclient
nickanderson: ack

nickanderson:
@kacf updated per your comments, but i did not make any unit test shell script for this. I will skip it if I can.

kacf:
Sure. I think we should not hold it back because of that. Looks good now!

kacf:
Merged manually after adding changelog entry.

PR #643, opened 2016-01-28
Fix: Suppress errors from unresolved variable

nickanderson:
Ref: https://dev.cfengine.com/issues/7931


kacf:
Thanks, I fixed some funky indentation and submitted manually.

PR #648, opened 2016-02-11
debian: Provide our default gateway interface name in variable

teroka:
Found myself in the need to know the IP address of the interface routed to default gateway.
This should help others bumping in to the same.

Dunno if the inventory is the right place for this.

### DIFF
--- a/inventory/linux.cf
+++ b/inventory/linux.cf
@@ -28,6 +28,15 @@ bundle common inventory_linux
       # this is the same as the original file for non-links
       "proc_1_process" string => filestat($(proc_1_cmdline), "linktarget");
 
+    any::
+      "proc_routes" data => data_readstringarrayidx("/proc/net/route",
+                                                    "#[^\n]*","\s+",40,4k),
+        ifvarclass => fileexists("/proc/net/route");
+      "routeidx" slist => getindices("proc_routes");
+      "dgw_ipv4_iface" string => "$(proc_routes[$(routeidx)][0])",
+        comment => "Name of the interface where default gateway is routed",
+        ifvarclass => strcmp("$(proc_routes[$(routeidx)][1])", "00000000");
+
   classes:
 
     any::


inventory/debian.cf
```diff
@@ -23,6 +23,14 @@ bundle common inventory_debian
       "mint_release" string  => "$(mint_info[RELEASE][1])" ;
       "mint_codename" string => "$(mint_info[CODENAME][1])" ;
 
+    any::
+      "proc_routes" data => data_readstringarrayidx("/proc/net/route",
+                                                    "#[^\n]*","\s+",20,2048);
```
nickanderson: Lets guard the parsing of this file on it's presence:

`ifvarclass => fileexists("/proc/net/route")`

Actually I think that ifvarclass can't use fileexists directly, so you might have to do something like `ifvarclass => fileexists("/proc/net/route")`

nickanderson:
Oh i think this is a nice addition.

But the same should work on other linux platforms so can you move it to inventory/linux.cf?

inventory/debian.cf
teroka: Seems to work directly as well (3.6.6).
Another thing is if you'd prefer bigger maxent / size values?

teroka:
Should be fine on other platforms as well. Only tried on Debian so placed it there.
I'll drop it to `linux.cf`.

tzz:
This will only work on platforms with procfs mounted under `/proc`.

Note that I'm doing the same thing in https://github.com/cfengine/core/pull/2245 but in C, and it will probably be merged. So it will be available without inventory, and **before** policy parsing. If you have time, look through that PR. This PR is still useful, especially if the data it creates can be in the same context as the one that comes from the C code, since it will work in older CFEngine versions too.

inventory/linux.cf
```diff
@@ -28,6 +28,15 @@ bundle common inventory_linux
       # this is the same as the original file for non-links
       "proc_1_process" string => filestat($(proc_1_cmdline), "linktarget");
 
+    any::
+      "proc_routes" data => data_readstringarrayidx("/proc/net/route",
+                                                    "#[^\n]*","\s+",40,4096),
```
nickanderson: Instead of 4096 I think 4K is more readable, what do you think?

nickanderson:
Right, once  cfengine/core#2245 is merged, then we can simply switch to using sys.proc_routes instead of reading the data in again and guard that on version.

nickanderson:
@teroka just waiting to hear your thoughts on 4K vs 4096 with regard to readability,

teroka:
@nickanderson sure, I'll change it

nickanderson:
Thanks for the contribution!

nickanderson:
Hey  @teroka I found that this didnt account for the case when there are multiple interfaces set to be a default gateway, but with different metrics.

See the policy here:
https://gist.github.com/nickanderson/5fe94b03d7b534dd7d06

@tzz might be interested in that as well in consideration of the proc net parsing being proposed in core. Did you handle ensuring you had the lowest metric for which interface was the default gateway?

@LINDEN-IT-NET you might also be interested in this snippet.

tzz:
Good point about the lowest metric. In penance I added that to https://github.com/cfengine/core/pull/2245 and also added the `active_default_gateway` flag to each route, so you can filter the routes for that if you want them regardless of their metric value.

teroka:
@nickanderson Good find indeed :+1: 
Don't have any clients which would have multiple so it didn't come up. Need to do some future proofing..

nickanderson:
Yeah good stumble. And ted has added that to the core PR as well to make much more network infor avaiailable from the start.

tzz:
wooo hooo this is merged thank you so much for your help everyone!

PR #649, opened 2016-02-11
Change: Tag default ipv4 gateway interface for inventory

nickanderson:
Changelog: Inventory of default ipv4 gateway interface when
class cfengine_internal_inventory_dgw_ipv4_iface is defined

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -770,15 +770,15 @@ bundle edit_line set_line_based(v, sep, bp, kp, cp)
       "^$(cp)($(i)$(bp).*|$(i))$"
              comment => "Uncommented the value '$(i)'",
         replace_with => value("$(i)$(sep)$($(v)[$(i)])"),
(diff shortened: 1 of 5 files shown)


nickanderson:
@estenberg I did it like this so that it's easy to augment the mission portal inventory, not something that everyone would want but this allows it to be turned on by just setting a class. Thoughts?

nickanderson:
ping @estenberg

estenberg:
Agree, looks useful and doesn't change that often; so fits within scope of inventory.

However, this is already implemented in C and tagged as part of #2245 (look for default_gateway) so I think we should wait and see what happens with that pull request and close this one if it goes in.

nickanderson:
@tzz @estenberg with the new proc parsing for linux network information is the default gw interface already tagged? If so then I don't think we need this.

nickanderson:
This PR seems to have gotten messed up anyhow

nickanderson:
those commits apper to already be in master from #679

PR #650, opened 2016-02-16
Update README.md with better instructions

nickanderson:


nickanderson:
ping @kacf or @jimis is this correct? There was a github issue where someone noted that they needed to run configure when downloading the release tarball from github.

jimis:
@nickanderson you need to run `./configure` no matter where you got the source code from.

The difference is that git checkout also needs `./autogen.sh` in advance. And the hidden behaviour here is that `./autogen.sh` calls configure inside. But to be on the safe side I would say that calling `./configure` is needed always. Same in core.

kacf:
Yes, it's correct for tarballs.

PR #652, opened 2016-02-18
Redmine #7825: Allow installation of yum packages with versions.

kacf:
Yum has some packages with a version string in the name, so we need to
allow that. An example is java-1.6.0-openjdk.

Changelog: Installing packages containing version numbers using yum
now works correctly.


kacf:
trigger build

kacf:
Cherry-picked to 3.7.x and 3.8.x as well.

PR #655, opened 2016-02-24
masterfiles#654 pkgsrc package module sometimes matches wrong package

### LINKED ISSUE #654: pkgsrc package module sometimes matches wrong package
This happens if the requested package is a substring of another package name.

E.g.:

````
echo File=mtr | ./pkgsrc get-package-data
PackageType=repo
Name=mtree
Version=20130908
Architecture=x86_64
````

### PR DESCRIPTION
**Pull request by bahamat**
Fixes #654, #656, #657.

```
$ echo File=mtr | ./pkgsrc get-package-data
PackageType=repo
Name=mtr
```

### DIFF
--- a/modules/packages/pkgsrc
+++ b/modules/packages/pkgsrc
@@ -102,7 +102,7 @@ get_package_data () {
         # or version, return nothing.
         # There's possibly a bug here because we're already emitting that the
         # PackageType is repo.
-        parse_pkg_data "$(pkgin -pP avail | grep "^$File" | grep "$Version;" | sort -n | tail -1)"
+        parse_pkg_data "$(pkgin -pP avail | grep "^${File}-" | grep "$Version;" | sort -n | tail -1)"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


nickanderson:
[Package Module API Reference](https://docs.cfengine.com/lts/reference-standard-library-package_modules.html) 

> For repository based package names the module should not return Version and Architecture, since they are often ambiguous in repository situations, and any discrepancies will be handled at the install stage instead.

bahamat:
Not returning the version is a bug in the specification since multiple versions may be available, in which case the wrong version may get installed. One wonders what the get-package-data subcommand is for, except  to `| sed s/File/Name/`.

Nevertheless, I've addressed these issues as well.

nickanderson:
@kacf can you comment on the protocol?

bahamat:
It was previously discussed here: https://dev.cfengine.com/issues/7328

The workaround was that version info can be fed to `get-package-data`, but then is discarded if returned.

bahamat:
At any rate, how the protocol *should* work is aside from the changes in this PR. There are two bugs pending to be fixed here, and the module now conforms to the current protocol specification. I don't want to make this PR about changing the spec.

kacf:
Changes look right to me.

PR #659, opened 2016-03-02
Add log rotation policy for state/classes.jsonl log.

maciejmrowiec:
Redmine: #7951


nickanderson:
@maciejmrowiec do you want to fix the Changelog typo?

"rotatio" should be "rotation" I think. Otherwise looks good and I can merge.

PR #663, opened 2016-03-15
Change: Specify Protocol Version

nickanderson:
This was previously commented out because older agents did not
understand the syntax. All currently supported versions understand this
syntax now, so it is safe to uncomment and set to a specific protocol
version.


nickanderson:
@jimis you requested this.

jimis:
Just remove it altogether, the default is protocol "2" today.

nickanderson:
@jimis here

jimis:
Thanks @nickanderson

PR #664, opened 2016-03-15
Fix: Problem with git not dropping privileges soon enough

nickanderson:
If we don't check that we have defined $(repo_uid) when setting the
need_to_drop class we end up with one pass of the agent executing with
elevated privileges. This lean leave artifacts like .git/index with
elevated ownership interfering with subsequent commands executed with
dropped privileges.

Changelog: Title


tzz:
YES

PR #667, opened 2016-03-16
Do not allow anybody to remotely control the hub.

jimis:


jimis:
@nickanderson please review and merge, we have been talking about this but I don't see a ticket. It works as expected though, I manually tested it.

nickanderson:
Looks good to me but missing changelog entry?

jimis:
Does it need one?

PR #668, opened 2016-03-17
Redmine #7961: Acceptance test to make sure def.json:inputs is picked up.

kacf:


tzz:
looks OK to me

tests/acceptance/00_basics/def.json/inputs.cf
```diff
@@ -0,0 +1,62 @@
...
+
+      "" usebundle => file_copy("$(this.promise_filename).json", "$(sys.inputdir)/def.json");
+}
+
+#######################################################
+
+bundle agent check
+{
+  vars:
+    "command" string => "$(sys.cf_promises) --show-classes -f $(sys.inputdir)/promises.cf|$(G.grep) test_class";
```
nickanderson: I think you can use `dcs_passif_output` here instead.

nickanderson: oh duh you did, ignore me.

kacf:
Cherry-picked to 3.8.x, 3.7.x is pending.

PR #661, opened 2016-03-09
Backport feature_def_json_preparse 

nickanderson:

PR #669, opened 2016-03-18
3.7.x backports

kacf:


kacf:
@nickanderson: I forgot about your original https://github.com/cfengine/masterfiles/pull/661 while I did this cherry-picking, but the patches appeared to apply rather easily. Yet I ended up with a different result than you did. Is there a significant difference between this one and https://github.com/cfengine/masterfiles/pull/661?

kacf:
@tzz: You might also want to have a look at this one.

kacf:
Merging so I can start a package build, but I would still appreciate a review.

PR #672, opened 2016-03-18
def.json fixup

kacf:


kacf:
@nickanderson: This is the result.

controls/3.7/def.cf
```diff
@@ -7,10 +7,9 @@
 bundle common def
 {
   classes:
-    # If the augments_file is parsed from C then we do not need to do this work
-    # from policy.
+    # If the augments_file is parsed from C then we do not need ot do this work
+    # from policy
     !feature_def_json_preparse::
-      # all override classes are defined true
       "$(override_classes)" expression => "any", meta => { "override" };
```
nickanderson: The override_classes needs to be removed here

controls/3.7/update_def.cf
```diff
@@ -1,8 +1,7 @@
 bundle common update_def
 {
   classes:
-    !feature_def_json_preparse::
-      # all override classes are defined true
+    !feature_def_json_pre_parse::
       "$(override_classes)" expression => "any", meta => { "override" };
```
nickanderson: override_classes also needs to be removed here

PR #673, opened 2016-03-21
Package module too verbose redmine7485

jimis:


jimis:
@kacf please review

The problem with the quick solution that I implemented is that it always suppresses `stdout` of subcommands, and sometimes you get an error on `stderr` and you can't find out what caused it:
```
[root@localhost ~]# /var/cfengine/bin/cf-agent -KIf ./redmine7485-pack.cf
   error: Error installing package 'vim'
```

The ideal solution would be to redirect stdout always, and have our C code log `stdout` with DEBUG or VERBOSE level, while `stderr` is logged at ERR level. But that change is way more complex and the ticket needs to be re-estimated if we follow that approach.

kacf:
Looks good to me!

kacf:
trigger build

kacf:
trigger build

jimis:
All green. @kacf quick question before merging: the way I've coded it required python 2.4 (because of os.devnull), is that OK?

kacf:
That's fine, the latest one is 2.7.

jimis:
Thanks, merging and cherry-picking down to 3.8.2 and 3.7.3.

PR #674, opened 2016-03-22
Add: file_make_mustache bundle to render mustache templates

nickanderson:


lib/files.cf
```diff
@@ -2165,6 +2165,33 @@ bundle agent file_make_mog(file, str, mode, owner, group)
...
+# @param data data container to use
+#
+# **Example:**
+#
+# ```cf3
+# vars:
+#   "state" data => datastate();
+#
+# methods:
+#      "" usebundle => file_make_mustache( "/tmp/z.txt", "/tmp/z.mustache" @(state) );
```
tzz: typo: missing comma

tzz:
I would also add a bundle to do this with permissions and ownership (since bundles can't take optional parameters).

nickanderson:
Thanks for the review @tzz

tzz:
nice :)

PR #676, opened 2016-03-29
Fix: Only parse services/autorun/*.cf when services_autorun is defined

nickanderson:
CFE-2135 #close Done

Changelog: Title

### DIFF
--- a/promises.cf
+++ b/promises.cf
@@ -297,7 +297,7 @@ bundle common cfengine_controls
 bundle common services_autorun
 {
   vars:
-    !cfengine_3_7::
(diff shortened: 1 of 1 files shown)


kacf:
FYI, I had to revert this because it breaks the build, specifically `tests/acceptance/promises/autorun.cf`. Probably not hard to fix, but I have to get on with pull requests and I need a green build for that.

kacf:
(btw, if `/var/cfengine` exists on the host it seems to pass erronously, so make sure you remove that)

nickanderson:
NEW PR https://github.com/cfengine/masterfiles/pull/681

PR #679, opened 2016-03-30
Scope edit_line bundle classes

nickanderson:
Ref: https://tracker.mender.io/browse/CFE-1959
To prevent class collisions between editing two files with similar formats we
scope the internal implementation classes to bundle scope. By default classes
defined from classes bodies are namespace scoped, so editing two files with
similar formats using the same edit_line bundle can be troublesome.

All of the edit_line bundles are now using the results classes body
consistently.  The results body was chosen for clarity. Note that previous ifok
body usage now uses (_kept|_repaired).

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -770,15 +770,15 @@ bundle edit_line set_line_based(v, sep, bp, kp, cp)
       "^$(cp)($(i)$(bp).*|$(i))$"
              comment => "Uncommented the value '$(i)'",
         replace_with => value("$(i)$(sep)$($(v)[$(i)])"),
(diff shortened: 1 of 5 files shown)


nickanderson:
trigger build

nickanderson:
Build failed due to unrelated change for how services_autorun is implmented.
01:13:09.238 ./promises/autorun.cf FAIL (UNEXPECTED FAILURE)

This should be built again after https://github.com/cfengine/masterfiles/pull/681 is sorted

nickanderson:
Now that  #681 is passing, lets try this again.

nickanderson:
trigger build

nickanderson:
trigger build

nickanderson:
trigger build

PR #680, opened 2016-03-30
Change: Autorun Tests

nickanderson:
It is expected that if the class services_autorun is defined from policy
that it will not be defined soon enough to automatically extend inputs
with the policy files found in services/autorun.

Additionally add a test that when services_autorun is defined from the
augments file it works as expected similar to how we have a test for
defining services_autorun as an option to the agent.


nickanderson:
trigger build

nickanderson:
superseded by https://github.com/cfengine/masterfiles/pull/681 which reverts the reverts :)

PR #681, opened 2016-03-30
Fix: Only parse services/autorun/*.cf when services_autorun is defined

nickanderson:
Ref: CFE-2135

This includes changes to one test so that it is expected to fail if
services_autorun is defined from policy. This also includes a new test that
when services_autorun is defined from def.json it does what is expected similar
to the test when services_autorun is defined as a command line option.

### DIFF
--- a/promises.cf
+++ b/promises.cf
@@ -297,7 +297,7 @@ bundle common cfengine_controls
 bundle common services_autorun
 {
   vars:
-    !cfengine_3_7::
+    !cfengine_3_7.services_autorun::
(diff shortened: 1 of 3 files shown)


nickanderson:
trigger build

nickanderson:
trigger build

nickanderson:
trigger build

nickanderson:
All green: http://jenkins.usdc.cfengine.com/job/testing-enterprise-pr_masterfiles/75/

PR #683, opened 2016-04-05
CFE-2183: Figure out version instead of hardcoding.

kacf:


kacf:
@nickanderson: You might be interested in this.

I'm not super happy with how complex the code is for locating the policy files, but this could be simplified later if we move the policy files into a `src` directory, as we have discussed previously.

kacf:
Right, this breaks Travis because it expects the files without `.in`. I'll take a look at that.

kacf:
Travis issue fixed.

nickanderson:
This would be a change that would make it impossible to use the masterfiles repository directly. I think I would consider adding an extra release artifact for masterfiles. A tarball that could be extracted into place without the use of the automake stuff.

I agree it would make it easier to find all the things if all the policy and templates were under a single directory.

kacf:
> I think I would consider adding an extra release artifact for masterfiles.

We already produce such an artifact, check for example: http://buildcache.usdc.cfengine.com/packages/packages-community-3.8.x/jenkins-bootstrap-packages-3.8.x-31/output/PACKAGES_x86_64_linux_redhat_6/cfengine-masterfiles-3.8.2.tar.gz. I'm unsure if we distribute it though, perhaps @jimis knows.

Using the masterfiles repository directly is anyway something that will be difficult if we want to autogenerate certain parts of the policy, like the version.

jimis:
> We already produce such an artifact

I believe the tarball you pointed at is the result of `make dist`, which means the user needs to run `./configure; make install` to actually install the masterfiles. It would be nice, I agree, to generate a secondary tarball, with masterfiles already installed; a "masterfiles package" I would say, besides the source tarball.

nickanderson:
Right, that's my understanding as well Jimis.

Having runable masterfiles for a given release is handy for users and upgrade planning.

jimis:
> This would be a change that would make it impossible to use the masterfiles repository directly.

@nickanderson exactly, the masterfiles directory should not be used directly, even today.

* If somebody checks it out of github, he should run `./autogen.sh; ./configure; make install`. 
* If somebody untars the source tarball, he should run `./configure; make install`
* I agree we should have a "package tarball", and maybe clarify the above in our README.md.

BasementTrix:
On Tue, Apr 05, 2016 at 04:59:22AM -0700, Kristian Amlie wrote:
> -- Commit Summary --
> 
>   * CFE-2183: Add determine-version.sh script from core.

But it's actually a Python script.  I have nothing against Python, per
se, but would you be willing to look at a Shell script?  AIX, for
example, doesn't have Python by default.  It does have Bourne Shell,
sed, and AWK.

-- 

GregorioDiStefano:
(you can use: https://docs.python.org/2/library/argparse.html for parsing command line arguments)

kacf:
@BasementTrix: You don't need to run it on AIX. This happens before the tarball is produced, in the autogen stage, so all the tarballs already contain the information, and hence will not run the script. If you were to run the autogen stage on AIX, you'd need many other tools as well, like autoconf/automake.

kacf:
> Having runable masterfiles for a given release is handy for users and upgrade planning.

It's still very easy to create a tarball for this, just `./autogen.sh && make dist`.

kacf:
> (you can use: https://docs.python.org/2/library/argparse.html for parsing command line arguments)

Thanks, good to know. I don't think I will use it for this case though, seems overkill for only one argument.

BasementTrix:
On Tue, Apr 05, 2016 at 11:14:53PM -0700, Kristian Amlie wrote:
> @BasementTrix: You don't need to run it on AIX.

I used AIX as my test case because it is a supported platform for
Enterprise.  My actual use case is FreeBSD.

-- 

kacf:
@BasementTrix: Ok, but the point still stands. You can very easily create a tarball for AIX by doing `./autogen.sh && make dist`. However, if you want to do that *on* FreeBSD, you need autoconf, automake and python. I think this is not an unreasonable assumption for a build machine.

kacf:
s/tarball for AIX/tarball for FreeBSD/

jimis:
`s/tarball for FreeBSD/source tarball for any platform/`

Basically either download the source tarball from our website, or go to a linux box and run `make dist`, as @kacf said. The generated tarball should compile on all platforms, be it FreeBSD or AIX. Same thing is true for both core and masterfiles repositories.

PR #686, opened 2016-04-06
Newer versions of Red Hat no longer ship with realpath.

skreuzer:
This binary has been removed from coreutils in favor of readlink -f which offers the same functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,9 @@ If you have cloned the repository from github:
 make install
 ```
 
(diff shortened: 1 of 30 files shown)


nickanderson:
Can you add a line to your commit message:

```
Changelog: Change: readlink -f used for realpath in stdlib paths for redhat
```

skreuzer:
> Can you add a line to your commit message:

I just amended the commit message

PR #687, opened 2016-04-07
Changelog: Change: readlink -e used for realpath in stdlib paths for …

skreuzer:
…redhat

### DIFF
--- a/lib/paths.cf
+++ b/lib/paths.cf
@@ -315,7 +315,7 @@ bundle common paths
       "path[tr]"            string => "/usr/bin/tr";
       "path[wc]"            string => "/usr/bin/wc";
       "path[wget]"          string => "/usr/bin/wget";
-      "path[realpath]"      string => "/usr/bin/realpath";
+      "path[realpath]"      string => "/usr/bin/readlink -e";
 
       #
       "path[chkconfig]" string => "/sbin/chkconfig";


lib/paths.cf
```diff
@@ -315,7 +315,7 @@ bundle common paths
       "path[tr]"            string => "/usr/bin/tr";
       "path[wc]"            string => "/usr/bin/wc";
       "path[wget]"          string => "/usr/bin/wget";
-      "path[realpath]"      string => "/usr/bin/realpath";
+      "path[realpath]"      string => "/usr/bin/readlink -e";
```
nickanderson: Actually I just thought of something this won't work well with.

We define classes based on the presence of the path. `/usr/bin/readline[space]-e` doesn't exist.

I think you might be interested in https://tracker.mender.io/browse/CFE-1804?jql=text%20~%20%22paths%22

I think instead of swapping realpath like that and masking what it really is, it would be better to add the path for readlink.

Thoughts?

tzz: The part that defines the class could look at the `basename()` of the full path.

nickanderson: Or maybe nth(splitstring on a space?

"$(all_paths)" string => nth(splitstring("$(path[$(all_paths)])";

nickanderson: Still it makes me slightly uncomfortable that we are kind of lying about the path to the binary. Its not really realpath, its really readlink.

And that would be OK if we were content to suffer on with the current way of doing things forever, but we would break it for people if we introduce actual PATHs for commands. Similar to how a few people like @zzamboni complained when we re-structured standard_services and some of the aliases like httpd => apache2 were not preserved.

skreuzer: > Still it makes me slightly uncomfortable that we are kind of lying about the path to the binary. Its not really realpath, its really readlink.

The problem is that in older versions of coreutils realpath was a binary on redhat as well as several other platforms as well. If you actually make use of $(paths.path[realpath]) across multiple platforms you will now have to rewrite these agents and jump through hoops.

Seems like changing this to readlink will offer backwards compatibility

nickanderson: Yes it's nice in that regard.

I wonder of it would make sense if paths were imolemted at the command line to allow a paths.path to suffice for a first hit. 

Then the paths bundle could be a user addition with these kinds of overrides.

tzz: After thinking about it, you're trying to do two things in one:

* provide a path to a common binary
* provide a way to find the real path of something

The `paths` bundle **only** does the first thing. So I don't think it should lie about `realpath`.

For the second item, I would provide a bundle in `files.cf` **or** extend the `filestat()` function (but also note it already does quite a bit, see https://docs.cfengine.com/docs/master/reference-functions-filestat.html ). That would not only address the need, but it would also serve all platforms in a consistent way.

nickanderson: This is more of an alias. 

Hummmmm

nickanderson:
Thanks @skreuzer but I am going to go ahead and close this one out.

PR #691, opened 2016-04-13
inventory: rework LLDP inventory to produce a "info" data container through -f json

tzz:
@nickanderson as discussed in #531 and IRC

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -349,7 +349,15 @@ bundle common inventory_control
 # are disabled.
 {
   vars:
-      "lldpctl_exec" string => "/usr/bin/lldpctl";
+      "lldpctl_exec" string => ifelse(fileexists("/usr/sbin/lldpctl"), "/usr/sbin/lldpctl",
+                                     fileexists("/usr/local/bin/lldpctl"), "/usr/local/bin/lldpctl",
+                                     "/usr/sbin/lldpctl");
+
+      "lldpctl_json" string => "$(lldpctl_exec) -f json",
+        unless => isvariable("def.lldpctl_json");
+      "lldpctl_json" string => "$(def.lldpctl_json)",
+        if => isvariable("def.lldpctl_json");
+
       "lsb_exec" string => "/usr/bin/lsb_release";
       "mtab" string => "/etc/mtab";
       "proc" string => "/proc";
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -531,11 +531,9 @@ bundle agent cfe_autorun_inventory_LLDP
   classes:
       "disable_inventory_LLDP" not => fileexists($(inventory_control.lldpctl_exec));
 
-  commands:
+  vars:
     !disable_inventory_LLDP::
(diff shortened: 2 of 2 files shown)


controls/def.cf
```diff
@@ -349,7 +349,8 @@ bundle common inventory_control
 # are disabled.
 {
   vars:
-      "lldpctl_exec" string => "/usr/bin/lldpctl";
```
nickanderson: I think these should be moved inside the bundle they are used in. Set to the default value if there is no def.lldpctl_json defined.

What do you think?
At least lldpctl_json needs to get ifvarclass => not(isvariable(def.lldpctl_json)) so that users can simply define lldpctl_json from def.json to use something like your special perl script.

BasementTrix: On Wed, Apr 13, 2016 at 01:21:47PM -0700, Nick Anderson wrote:
> I think these should be moved inside the bundle they are used
> in. Set to the default value if there is no def.lldpctl_json
> defined.
> 

Probably a good idea.  Would cover the (e.g. FreeBSD) case where the
path is /usr/local/bin/lldpctl.

-- 

tzz: For others like `lsb_exec` we put the executables inside `inventory_control` so users have just one place to change.

I'll add the `def` override.

tzz: @BasementTrix I added a case for `/usr/local/bin/lldpctl` as follows:

* if `/usr/sbin/lldpctl` exists, use it
* else if `/usr/local/bin/lldpctl` exists, use it
* else use `/usr/sbin/lldpctl` as a last resort

Try it out and see if it does the right thing for you. Thanks.

BasementTrix: On Wed, Apr 13, 2016 at 01:30:49PM -0700, Ted Zlatanov wrote:
> @BasementTrix I added a case for `/usr/local/bin/lldpctl` as follows:
> 
> * if `/usr/sbin/lldpctl` exists, use it
> * else if `/usr/local/bin/lldpctl` exists, use it
> * else use `/usr/sbin/lldpctl` as a last resort
> 
> Try it out and see if it does the right thing for you. Thanks.

While I'm sure it will work, why not just add to the 'vars:' section
starting at line 165 which already has class stanzas for 'freebsd::'
and '!freebsd::' where the path for dmidecode is defined?


-- 

controls/def.cf
```diff
@@ -349,7 +349,15 @@ bundle common inventory_control
...
+      "lldpctl_exec" string => ifelse(fileexists("/usr/sbin/lldpctl"), "/usr/sbin/lldpctl",
+                                     fileexists("/usr/local/bin/lldpctl"), "/usr/local/bin/lldpctl",
+                                     "/usr/sbin/lldpctl");
+
+      "lldpctl_json" string => "$(lldpctl_exec) -f json",
+        unless => isvariable("def.lldpctl_json");
+      "lldpctl_json" string => "$(def.lldpctl_json)",
+        if => isvariable("def.lldpctl_json");
+
```
nickanderson: Looks good to me.

controls/def.cf
tzz: I think other platforms may have `/usr/local/bin/lldpctl` and it will Just Work there too.

PR #697, opened 2016-04-18
CFE-2332 Use new package promise by default.

pasinskim:
From 3.9 on the new package promise is the default package promise
implementation used by CFEngine.

Changelog: Use new package promise as default package promise
implementation. (CFE-2332)

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -560,8 +560,7 @@ bundle agent cfe_autorun_inventory_packages
   packages:
     # From 3.7 onwards there is a new package promise implementation using package
     # modules, in which case installing non_existing_package is not needed any more.
-    #(cfengine_3_5|cfengine_3_6).debian::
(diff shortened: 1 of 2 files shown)


pasinskim:
@kacf @nickanderson this is quite straightforward change. Just need to make sure I am not missing anything.

kacf:
Looks good to me, but it looks like Travis doesn't agree.

pasinskim:
@kacf Travis is failing with 3.6 sanity check which is correct as this stuff won't work with 3.6.

kacf:
Ok, then we should remove that check from Travis. 3.9 masterfiles isn't, and won't be, compatible with 3.6.

pasinskim:
@kacf now build is green. Any final comments before I will merge this one?

kacf:
I'm happy with it!

PR #698, opened 2016-04-19
Remove sanity checking for 3.6 version

pasinskim:
3.9 won't be compatible with 3.6 so we need to remove 3.6 check from Travis.


nickanderson:
+1

PR #700, opened 2016-04-20
Fix: Work around impaired class definition from augments

nickanderson:
Ref: Jira #CFE-2333

Changelog: Title


nickanderson:
trigger build

PR #703, opened 2016-04-22
Fix hitting zypper cache for every package promise in the policy.

pasinskim:
Default 'package_list_update_ifelapsed' was set to 0 for OpenSUSE.
This meant that every time package promise was evaluated SUSE repositories
were checked for the new updates.
This should make the behavior unified with other package promises, where
cache is hit every 240 minutes.


pasinskim:
@kacf this looks to be quite obvious and straightforward change.

kacf:
Great stuff! You should add `Changelog: Title` to the commit though.

pasinskim:
Fixed!

kacf:
Cherry-picked to 3.7.x and 3.8.x as well.

PR #704, opened 2016-04-22
Make Database purge schedule more robust

mikeweilgart:
Rather than restricting the database maintenance to specifically
the first 5 minutes of the midnight - 1AM hour, restrict it only
to some point within that hour, and use 'if_elapsed_day' to
prevent multiple runs in a single day.

With the default execution schedule and default splaytime, this
will be effectively the same as before, but this change will
make the DB maintenance promises more robust against execution
schedule changes.  (E.g. if CFEngine only runs every 20 minutes
with an splaytime to match, it may never run in the Min00_05
interval and thus never perform the DB maintenance needed.
This commit fixes that.)

### DIFF
--- a/lib/cfe_internal_hub.cf
+++ b/lib/cfe_internal_hub.cf
@@ -135,33 +135,33 @@ bundle agent cfe_internal_hub_maintain
 
   classes:
       "enable_cfe_internal_reporting_database_purge_old_history" -> { "postgres", "CFEngine Enterprise" }
-        expression => "enterprise_edition.Hr00.Min00_05";
+        expression => "enterprise_edition.Hr00";
 
       "enable_cfe_internal_parition_creation_job" -> { "postgres", "CFEngine Enterprise" }
-        expression => "enterprise_edition.Hr00.Min00_05";
+        expression => "enterprise_edition.Hr00";
 
   methods:
     enable_cfe_internal_reporting_database_purge_old_history::
       "Remove old report history"
       usebundle => cfe_internal_database_cleanup_reports(@(report_settings)),
-      action => if_elapsed(5);
+      action => if_elapsed_day;
 
       "Remove cf-consumer history"
       usebundle => cfe_internal_database_cleanup_consumer_status("50000"),
-      action => if_elapsed(5);
+      action => if_elapsed_day;
 
       "Remove diagnostics history"
       usebundle => cfe_internal_database_cleanup_diagnostics(@(diagnostics_settings)),
-      action => if_elapsed(5);
+      action => if_elapsed_day;
 
       "Remove promise log history"
       usebundle => cfe_internal_database_cleanup_promise_log("7"),
-      action => if_elapsed(5);      
+      action => if_elapsed_day;
 
     enable_cfe_internal_parition_creation_job::
       "Create new partitions for partitioned tables"
       usebundle => cfe_internal_database_partitioning(),
-      action => if_elapsed(5);
+      action => if_elapsed_day;
 }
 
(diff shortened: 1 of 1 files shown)


mikeweilgart:
@estenberg this is the improvement I was thinking of.

nickanderson:
+1

nickanderson:
It would also be nice to expose a variable in def for overriding the expression string. Then someone can adjust the default window for that maint operation without touching masterfiles.

mikeweilgart:
@nickanderson, interesting idea.  So, just have the `Hr00` portion be part of the override, then?  (Not expose the `enterprise_edition` portion to potential mistaken overrides.)

I guess you have a view I haven'

estenberg:
Definitely an improvement that makes it more robust against cf-execd schedule changes!
Thanks for contributing.

nickanderson:
@mikeweilgart

My view is that the masterfiles policy framework should be easy to use. If you want to have this operation run in the afternoon or only when the server is not under abnormally high load you have to dig deep into the policy right now. 

After you make that change you will be responsible for maintaining that difference through each policy framework upgrade. 

In contrast if we use a variable that can be overridden based on a def variables value the maintenance of that customization could move into the augments file,  which is a pure overlay.

mikeweilgart:
@nickanderson 

Makes sense.  However I think there is a point at which the `def` file can become so complicated that it itself is hard to follow.  For instance, if someone wants to restrict the DB cleanup based on load average, it is definitely beyond just a straightforward use of `def`, and gets into some potentially very involved logic.

This reminds me of a [discussion I recently read of so-called "magic numbers" in CSS](http://programmers.stackexchange.com/a/315151/200593).  (In a nutshell, CSS *is* a configuration file, even though it allows logic.  Overzealous avoidance of "magic numbers" in CSS eventually leads to creating a config file for your config file.  The example may not parallel perfectly to CFEngine, but I think it's relevant.)

There's a balance to be struck between allowing simple configuration edits without touching the masterfiles policy framework, and having too much indirection.

My view is that if you want to adjust the CFEngine internal hub maintenance policy, the logical place to look is the `cfe_internal_hub_maintain` bundle, not the `def` bundle.  :)

(P.S.: This is a great argument for allowing scoped variable overrides from `def.json`, by the way, since it would allow you to keep the details relevant to XYZ policy within the XYZ bundle, and allow overrides to the details of that policy without touching the masterfiles policy framework at all.)

nickanderson:
Right, its difficult to find the balance. 

This is what I was thinking:
```
#+begin_src cfengine3
  bundle agent cfe_internal_hub_maintain
  {
    vars:
        "expr_reporting_db_purge_history"
          string => ifelse( isvariable("def.$(this.promiser)"), "$(def.$(this.promiser))",
                            "Hr00" ),
          meta => { "defvar" },
          comment => "This expression identifies when the reporting databases
                      purging of history should be allowed. We look for a user
                      defined value for this in =$(def.$(this.promiser))=
                      (def.expr_reporting_db_purge_history). If no user defined
                      variable is found then set it to a sensible default.";

            # Note: I also tagged the variable with =defvar=. This allows me to do
            # easily find variables that can be specified in def.

    classes:
      enterprise_edition::
       "enable_cfe_internal_reporting_database_purge_old_history" -> { "postgres", "CFEngine Enterprise" }
          expression => "$(expr_reporting_db_purge_history)";
  }
#+end_src
```

mikeweilgart:
@nickanderson 

Awesome, that's a really good idea.  So the variable wouldn't *have* to be in `def`, but if the end user defined it there it would override the default schedule.  Makes sense!  :)

PR #705, opened 2016-04-29
cfengine functions chokes on output generated by systemctl.  

basvandervlies:
This change will limit the output to only the required systemd properties and report if service is unknown, see:
 * https://tracker.mender.io/browse/CFE-2086


nickanderson:
Nice, thanks!

tzz:
YES

PR #706, opened 2016-04-29
CFE-2358: Fix lldpctl error messages due to promise order.

kacf:
The order is vars followed by classes, so the class
'!disable_inventory_LLDP' expression will always be true on the first
iteration. Fix by using a positive class match instead.

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -529,10 +529,10 @@ bundle agent cfe_autorun_inventory_LLDP
 # testing, and your Friendly Network Admin may be of help too.
 {
   classes:
-      "disable_inventory_LLDP" not => fileexists($(inventory_control.lldpctl_exec));
+      "lldpctl_exec_exists" expression => fileexists($(inventory_control.lldpctl_exec));
 
   vars:
-    !disable_inventory_LLDP::
+    !disable_inventory_LLDP.lldpctl_exec_exists::
       "info" data => parsejson(execresult($(inventory_control.lldpctl_json), "noshell"));
 }
 


kacf:
@pasinskim

tzz:
The intent was to let the user define that `disable` class and turn off LLDP inventory explicitly. Unfortunately things are rarely simple in negative-knowledge land. Thank you for fixing this, and note that a lot of the other inventory classes follow the same pattern.

kacf:
Hmm, maybe I should make a more general fix then. Are they all inside `inventory/any.cf`?

tzz:
@kacf yes, but only the LLDP disable is actually set, the rest are up to the user (so I'd expect them to be set in `def.json` for instance). So I think it's OK to leave as is, actually, sorry for the noise.

```
inventory/any.cf:    !disable_inventory_cmdb::
inventory/any.cf:    !disable_inventory_LLDP::
inventory/any.cf:    !disable_inventory_package_refresh::
inventory/any.cf:    !disable_inventory_proc::
inventory/any.cf:    !disable_inventory_cpuinfo::
inventory/any.cf:    !disable_inventory_fstab::
inventory/any.cf:    !disable_inventory_mtab::
inventory/any.cf:    !disable_inventory_dmidecode::
inventory/any.cf:      "disable_inventory_LLDP" not => fileexists($(inventory_control.lldpctl_exec));
inventory/any.cf:    !disable_inventory_LLDP::
```

kacf:
Yes, I think you're right. I tested this and it solves the problem.

PR #707, opened 2016-04-29
Zendesk#2873: fix cf-upgrade_script.sh being empty in SuSE.

jimis:


jimis:
@nickanderson I'm leaving reviewing, merging, cherry-picking to you! :-)

PR #709, opened 2016-04-29
Fix: Stop cfengine choking on systemctl output

nickanderson:
This change will limit the output to only the required systemd properties and report if service is unknown, see:

Ref: Jira #CFE-2806

(cherry picked from commit b77e153fc564d8a43f8af57c0c5439795c8ed642)

ChangeLog: Title


nickanderson:
trigger build

nickanderson:
Build was green.

PR #718, opened 2016-04-29
Add: edit_line contains_literal_block to stdlib

nickanderson:
Changelog: Title


nickanderson:
Without a test because: https://tracker.mender.io/browse/CFE-2331

PR #720, opened 2016-05-02
Change: Define testing_masterfiles_policy_framework

nickanderson:
This defines the testing_masterfiles_policy_framework class. This will
allow us to omit the plucked.cf.sub bundle when running the masterfiles
policy framework. Omitting the bundle allows us to test the framework
directly.

Ref: Jira #CFE-2331


nickanderson:
Dependant on https://github.com/cfengine/core/pull/2602

Will also need to fix some tests before this goes in.

nickanderson:
trigger build

nickanderson:
All green without having to fix anything. Hurah!

PR #721, opened 2016-05-03
Change 'allowlegacyconnects' comment to reflect latest changes.

pasinskim:
By default we are not accepting legacy connections any more.
This is just update of the comment in the masterfiles.


pasinskim:
@kacf last minute change to reflect latest functionality changes.

kacf:
Cherry-picked to 3.9.x as well.

PR #724, opened 2016-05-09
ENT-2737: Make sure modules directory is served via a shortcut.

kacf:


jimis:
@kacf that should be all that's needed!

kacf:
Cherry-picked to 3.9.x as well.

PR #727, opened 2016-05-24
Add: Definition of from_cfexecd for cf-execd initiated runs

nickanderson:
The agent will automatically append ' -Dfrom_cfexecd' if it is not found
in exec_command. This causes unexpected behavior when the appended
option replaces any custom definitions.

Jira #CFE-2386
Changelog: Title

(cherry picked from commit 33b0879e13757d7c854c68fe6a3500d554438ad4)


nickanderson:
The 3.8 series is done. Closing without merging.

PR #730, opened 2016-06-06
Augments controls for update policy

nickanderson:


update.cf
```diff
@@ -10,17 +10,15 @@ body common control
...
-                          "cfe_internal_update_processes",
+                          "mpf_update",
+
       };
 
       version => "update.cf $(update_def.current_version)";
 
       inputs => {
                   @(cfengine_update_controls.update_def_inputs),
+                  @(update_def.inputs),
```
tzz: do you mean `def.update_inputs`?

update.cf
```diff
@@ -31,6 +29,64 @@ body common control
...
+# @description These policies are related to ensuring the environment is
+# configured properly before a policy run. This includes publishing policy with
+# the masterfiles-stage tooling, making sure that $(sys.inputdir) is updated
+# from its upstream masterfiles, updating CFEngine packages, and ensuring the
+# necessary CFEngine components are running,
+{
+  vars:
+      "sorted_update_bundlesequence_end"
+        meta => { "defvar" },
+        slist => sort( @(def.update_bundlesequence_end), "lex" );
```
tzz: Can you use `def.update_bundlesequence_end` without the `@`?

update.cf
```diff
@@ -31,6 +29,64 @@ body common control
...
+
+      "num_custom_bundles" int => length( "sorted_update_bundlesequence_end" );
```
tzz: No need to quote this

update.cf
```diff
@@ -31,6 +29,64 @@ body common control
...
+
+  classes:
+      # TODO After 3.10 is no longer supported compress into a single promise
+      # nesting functions
+      "_have_custom_update_bundles"
+        expression => isgreaterthan( $(num_custom_bundles), 0);
```
tzz: Also `some(def.update_bundlesequence_end, '.*')` will test if a list is not empty.

controls/def.cf
```diff
@@ -126,6 +126,14 @@ bundle common def
                           not(isvariable("trustkeysfrom"))),
         meta => { "defvar" };
 
+      "cf_serverd_maxconnections"
```
tzz: So I'm wondering, shouldn't we try to nest these in `def.json`? So the top-level key is `cf_serverd` and then under that we have `maxconnections`? Or even `controls -> server -> maxconnections`? The `def.json` parser could definitely connect `controls -> AGENTNAME -> SETTING` with `body AGENTNAME control { SETTING => ... }` so we don't have multiple names for the same setting, so we could automatically have **all** the controls easily available, and so they can simply be set without policy.

tzz:
@kacf @estenberg @nickanderson please see my comments about possibly providing a way to set any agent controls (server, runagent, etc.) from `def.json` by using a consistent naming scheme.

controls/def.cf
nickanderson: I've been thinking about this as well.

I think there is value in being able to have policy that determines a setting if no user supplied value is given.

I wonder if it might make sense to have a cfengine:controls bundle get populated with special control settings, and that could be handled like bundle common def is today. If no bundle is present the settings from the augments file are used. If there is a bundle, then the policy can decide what is best.

update.cf
nickanderson: ack that is better.

I kept the check against the sorted list.
```
  classes:
      "_have_custom_update_bundles"
        expression => some( '.*', sorted_update_bundlesequence_end );
```

update.cf
nickanderson: No need to have it :-)

update.cf
nickanderson: I was thinking that I needed to handle definition of empty lists, so I pulled `def.update_inputs` into `update_def.inputs`

https://github.com/cfengine/masterfiles/pull/730/files#diff-8c9a709a840b3d3b6c09f6d4d5e4def6R103

controls/def.cf
tzz: I don't know what that bundle would look like, but it sounds like a pretty typical use case. How would a bundle be different from specifying the body control settings as usual?

update.cf
tzz: Oh, I see, that's confusing!

nickanderson: I tried to clarify it.

update.cf
nickanderson: 3.7:

```
./update.cf:50:27: error: syntax error
        slist => sort( def.update_bundlesequence_end, "lex" );
                          ^
```
It doesn't like the dot.

I can quote it:

```
        slist => sort( "def.update_bundlesequence_end", "lex" );
```

Or I can use the list notation as I am currently.

tzz: Don't use the `@(list)` notation, the second one you showed is better.

nickanderson: Is there a reason other than 1 less character?

tzz: `@(list)` is erratically supported until 3.9 (where functions started taking lists and data containers and parsing var refs more consistently). `list` is supported for sure.

controls/def.cf
nickanderson: The bundle would allow us to specify default values in the MPF that may be more appropriate to general use than the defaults in C.

With the single augments file we currently have only specifying values would be limiting.

That bundle might look something like this:

```
body file control
{
  namespace => "cfengine";
}

bundle common controls
{
  vars:
    "server[maxconnections]" int => ifelse( isvariable("server[maxconnections]"), "$(server[maxconnections])", "200");
}

It's really no different than doing it in def and agreeing on a consistent structure.

Comment (tzz) [inline controls/def.cf]: I see. This seems like a lot of work to get defaults working (3 layers: `def.json` to policy to `body server control`). Would the following work better? It only uses two layers (`def.json` to `body server control`).

```
body server control
{
  maxconnections => ifelse( isvariable("def.server[maxconnections]"), "$(def.server[maxconnections])", "200");
}
```

I think that serves the use case you mentioned and the case where you want to use classes, but I'm maybe missing others?

Comment (nickanderson) [inline controls/def.cf]: Shit, I should have thought about trying that directly. Does it work currently?

Comment (tzz) [inline controls/def.cf]: Functions should work in bodies, yes. I haven't tested it, but if it doesn't work I'd consider that a bug.

I would make the variable `def.cfe_controls[server][maxconnections]` so we have just one new variable in `def`.

Comment (nickanderson) [inline controls/def.cf]: OK, that doesn't work in 3.7.3 or 3.9.0

```
      ignore_missing_inputs => ifelse( isvariable("def.ignore_missing_inputs"),
                                                  $(def.ignore_missing_inputs),
                                       "false");

```

```
./promises.cf:12:47: error: Function does not return the required type
                                       "false");

```

Issue filed here: https://tracker.mender.io/browse/CFE-2400

**Comment by nickanderson** [comment]
ping @mikeweilgart

File: cfe_internal/core/log_rotation.cf [hunk 6]
```diff
@@ -5,35 +5,70 @@
...
+
+      "_max_log_size"
+        int => "$(def.cfengine_internal_max_log_size)",
+        ifvarclass => isvariable("def.cfengine_internal_max_log_size"),
+        comment => "If a user has defined a custom max log size we use that
+                    instead.";
+
+      "_max_log_rotations"
+        int => "10",
+        comment => "By default we will keep up to 30 rotated logs";
```
mikeweilgart: Is this a typo?  "Default up to 30" but the int is set to 10.

mikeweilgart:
@nickanderson, thanks for the ping.  I agree with @tzz's proposal for hierarchical use of def.json; I think it would simplify things.

His comments about removing layers of abstraction also inspired some other ideas that don't directly pertain to this pull request, so I emailed you about them separately.  :)

nickanderson:
I had too many issues trying to use one big data container for this and i moved to standardized individual variable names.

PR #747, opened 2016-06-20
Fix systemctl path when `/bin/systemctl` does not exists.

pasinskim:


pasinskim:
@nickanderson @kacf what do you think about this one?

nickanderson:
@kacf there is a ticket now: https://tracker.mender.io/browse/CFE-2427

nickanderson:
@pasinskim 

We could add a class check for `/usr/bin/systemctl` like we already have for `/bin/systemctl`.

See: https://github.com/pasinskim/masterfiles/blob/d91fcf5daae191f7cd4c79f38d6ec2e621f1d1b4/lib/paths.cf#L517

Then we could change the guard on `/usr/bin/systemctl` appropriately.

Things like this just make me yearn for https://tracker.mender.io/browse/CFE-1804

tzz:
@nickanderson yeah maybe it's time to work on CFE-1804... see if you can get it prioritized so the syntax can be settled before we can implement it.

kacf:
See [ticket](https://tracker.mender.io/browse/CFE-2427).

PR #754, opened 2016-06-23
Change: Add systemd services bundle

nickanderson:
Changelog: Title

This refactors standard_services pulling systemd related policy into its
own bundle. standard_services will call the systemd_services bundle if
systemd is detected.

This should improve maintainability of the systemd services
implementation.

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -125,6 +125,99 @@ body service_method standard_services
 
 ##
 
+body service_method systemd_services
+{
+      service_autostart_policy => "none";
+      service_dependence_chain => "ignore";
+      service_type => "generic";
+      service_args => "";
+}
+
+bundle agent systemd_services(service,state)
+# @brief Manage standard systemd services
+{
+  vars:
+
+    # We explicitly guard for systemd to avoid unnecessary agent time in
+    # pre-eval
+
+    systemd::
+      "call_systemctl" string => "$(paths.systemctl) --no-ask-password --global --system";
+      "systemd_properties" string => "-pLoadState,CanStop,UnitFileState,ActiveState,LoadState,CanStart,CanReload";
+      "systemd_service_info" slist => string_split(execresult("$(call_systemctl) $(systemd_properties) show $(service)", "noshell"), "\n", "10");
+
+  classes:
+
+    systemd::
+
+      # define a class named after the desired state
+      "$(state)" expression => "any";
+      "non_disabling" or => { "start", "stop", "restart", "reload" };
+
+# A collection of classes to determine the capabilities of a given systemd
+# service, then start, stop, etc. the service. Also supports a custom action
+# for anything not supported
+
+      "service_enabled" expression => reglist(@(systemd_service_info), "UnitFileState=enabled");
+      "service_active"  expression => reglist(@(systemd_service_info), "ActiveState=active");
+      "service_loaded"  expression => reglist(@(systemd_service_info), "LoadState=loaded");
+      "service_notfound" expression => reglist(@(systemd_service_info), "LoadState=not-found");
+
+      "can_stop_service"   expression => reglist(@(systemd_service_info), "CanStop=yes");
+      "can_start_service"  expression => reglist(@(systemd_service_info), "CanStart=yes");
+      "can_reload_service" expression => reglist(@(systemd_service_info), "CanReload=yes");
+
+      "request_start"   expression => strcmp("start", "$(state)");
+      "request_stop"    expression => strcmp("stop", "$(state)");
+      "request_reload"  expression => strcmp("reload", "$(state)");
+      "request_restart" expression => strcmp("restart", "$(state)");
+
+      "action_custom"  expression => "!(request_start|request_stop|request_reload|request_restart)";
+      "action_start"   expression => "request_start.!service_active.can_start_service";
+      "action_stop"    expression => "request_stop.service_active.can_stop_service";
+      "action_reload"  expression => "request_reload.service_active.can_reload_service";
+      "action_restart"         or => {
+                                      "request_restart.service_active",
+
+                                      # Possibly undesirable... if a reload is
+                                      # requested, and the service "can't" be
+                                      # reloaded, then we restart it instead.
+                                      "request_reload.!can_reload_service.service_active",
+                                     };
+
+      # Starting a service implicitly enables it
+      "action_enable"  expression => "request_start.!service_enabled";
+
+      # Respectively, stopping it implicitly disables it
+      "action_disable" expression => "request_stop.service_enabled";
+
+  commands:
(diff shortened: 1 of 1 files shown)


nickanderson:
@moonkid196 thoughts?

lib/services.cf
```diff
@@ -125,6 +125,99 @@ body service_method standard_services
...
+                                      # requested, and the service "can't" be
+                                      # reloaded, then we restart it instead.
+                                      "request_reload.!can_reload_service.service_active",
+                                     };
+
+      # Starting a service implicitly enables it
+      "action_enable"  expression => "request_start.!service_enabled";
+
+      # Respectively, stopping it implicitly disables it
+      "action_disable" expression => "request_stop.service_enabled";
```
moonkid196: Since you explicitly support here **action_enable**, **action_disable**, perhaps you should also support **request_enable**, **request_disable**, instead of letting that be handled implicitly by the **action_custom** class? Perhaps: `=> (request_start|request_enable).!service_enabled`

I'm just thinking this will be a common case: people setting a service to enabled/disabled (which is system state). Might as well not call _systemctl_ if it's not necessary?

lib/services.cf
```diff
@@ -125,6 +125,99 @@ body service_method standard_services
...
+
+  commands:
+    systemd.service_loaded:: # note this class is defined in `inventory/linux.cf`
+      # conveniently, systemd states map to `services` states, except
+      # for `enable`
```
moonkid196: _except for `enable`_ ... is this comment still true/relevant?

lib/services.cf
```diff
@@ -125,6 +125,99 @@ body service_method standard_services
...
+
+  reports:
+    DEBUG|DEBUG_systemd_service::
+      "DEBUG $(this.bundle): using systemd layer to $(state) $(service)";
+
+      "DEBUG $(this.bundle): Service $(service) unit file is not loaded; doing nothing"
+        ifvarclass => "systemd.!service_loaded";
+
+      "DEBUG $(this.bundle): Could not find service: $(service)"
+        ifvarclass => "systemd.service_notfound";
```
moonkid196: It's been a few months since I've really spent much time with cfengine (changed jobs, hope to correct this soon), but is there not an output level between normal and debug? Or at least `!quiet_mode` or some such? I feel like those first two reports are good, affirmative outputs that cfengine is doing what it's expected to; the 3rd does feel more like a debug message, though.

_(sorry, I don't expect that's exactly the feedback you're looking for, but I thought I'd throw it in)_

lib/services.cf
```diff
@@ -216,77 +304,7 @@ bundle agent standard_services(service,state)
...
                     automatically try to add the service at this time.";
 
-### BEGIN ###
-# @brief probe the state of a systemd service
-# @author Bryan Burke
-#
-# A collection of classes to determine the capabilities of a given systemd
-# service, then start, stop, etc. the service. Also supports a custom action
-# for anything not supported
-#
```
moonkid196: I lose my attribution...! _NOOOO_ ;)

moonkid196:
I put in some line-comments on the diff page. Overall, I support the idea that specific tools get their own bundle/namespace (see [my fork of the contrib repo](https://github.com/moonkid196/contrib)), then system detection logic can apply the right one for the system.

lib/services.cf
```diff
@@ -125,6 +125,99 @@ body service_method standard_services
...
+      "service_notfound" expression => reglist(@(systemd_service_info), "LoadState=not-found");
+
+      "can_stop_service"   expression => reglist(@(systemd_service_info), "CanStop=yes");
+      "can_start_service"  expression => reglist(@(systemd_service_info), "CanStart=yes");
+      "can_reload_service" expression => reglist(@(systemd_service_info), "CanReload=yes");
+
+      "request_start"   expression => strcmp("start", "$(state)");
+      "request_stop"    expression => strcmp("stop", "$(state)");
+      "request_reload"  expression => strcmp("reload", "$(state)");
+      "request_restart" expression => strcmp("restart", "$(state)");
```
tzz: The above will only work for those four states and it's verbose. To create a class for any state in one line:

```
"request_$(state)" expression => "any";
```

lib/services.cf
tzz: `enable` was added to the allowed states recently, so the comment can remove the "except" part

lib/services.cf
```diff
@@ -125,6 +125,99 @@ body service_method standard_services
...
+    systemd.service_loaded:: # note this class is defined in `inventory/linux.cf`
+      # conveniently, systemd states map to `services` states, except
+      # for `enable`
+
+      "$(call_systemctl) -q start $(service)"   ifvarclass => "action_start";
+      "$(call_systemctl) -q stop $(service)"    ifvarclass => "action_stop";
+      "$(call_systemctl) -q reload $(service)"  ifvarclass => "action_reload";
+      "$(call_systemctl) -q restart $(service)" ifvarclass => "action_restart";
+      "$(call_systemctl) -q enable $(service)"  ifvarclass => "action_enable";
+      "$(call_systemctl) -q disable $(service)" ifvarclass => "action_disable";
```
tzz: I think this can be reduced to

```
"$(call_systemctl) -q $(state) $(service)"   ifvarclass => "action_$(state)|action_custom";
```

lib/services.cf
```diff
@@ -125,6 +125,99 @@ body service_method standard_services
...
+
+      # Custom action for any of the non-standard systemd actions such a
+      # status, try-restart, isolate, et al.
+      "$(call_systemctl) $(state) $(service)" ifvarclass => "action_custom";
```
tzz: this can also be merged into the above

tzz:
@nickanderson maybe on systemd we should pre-load as much as possible from a single systemctl call to tell us what's loaded, enabled, etc. as Mr. Svoboda suggested.

nickanderson:
Oh somehow i missed all these comments, will come back and review. Thanks!

And yes, there are probably other improvments that can be made. But generally you like the idea to seperate systemd functionality into its own bundle, and have standard_services leverage it like it does for classic services?

nickanderson:
@mikeweilgart ping

lib/services.cf
mikeweilgart: @tzz that seems like a good idea at first glance, but there may be an interesting edge case: what happens if the variable `$(state)` is not defined but the class `action___state_` is set?  :)  (This may never happen; I haven't explored in detail how the `state` variable gets set, but it's worth considering.)

tzz: The promise will fail to evaluate if `$(state)` is not defined. The logic for class expressions is different but I'm pretty sure unexpanded variables are evaluation failures. As an edge case, I think this can be ignored, but we could enforce it with an explicit test like `if => some("^$(state)$", '["start", "stop", ...]')` (untested but you see what I mean).

PR #766, opened 2016-07-12
Add jUnit bundles and include them in stdlib.cf

tzz:
Needs review and testing @nickanderson

### DIFF
--- a/lib/stdlib.cf
+++ b/lib/stdlib.cf
@@ -71,6 +71,9 @@ bundle common stdlib_common
       "input[cfe_internal_hub]"           string => "$(this.promise_dirname)/cfe_internal_hub.cf";
       "input[cfengine_enterprise_hub_ha]" string => "$(this.promise_dirname)/cfengine_enterprise_hub_ha.cf";
 
+      # jUnit and TAP reports
+      "input[testing]"                      string => "$(this.promise_dirname)/testing.cf";
+
     any::
       "inputs" slist => getvalues(input);
 
--- a/lib/testing.cf
+++ b/lib/testing.cf
@@ -1,0 +1,330 @@
+############################################################################
+#  Copyright (C) CFEngine AS
+#
+#  This program is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License LGPL as published by the
+#  Free Software Foundation; version 3.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  To the extent this program is licensed as part of the Enterprise
+#  versions of CFEngine, the applicable Commercial Open Source License
+#  (COSL) may apply to this file if you as a licensee so wish it. See
+#  included file COSL.txt.
+###########################################################################
+#
+# CFEngine Community Open Promise-Body Library
+#
+# This initiative started by CFEngine promotes a
+# standardized set of names and promise specifications
+# for template functionality within CFEngine 3.
+#
+# The aim is to promote an industry standard for
+# naming of configuration patterns, leading to a
+# de facto middleware of standardized syntax.
+#
+# Names should be intuitive and parameters should be
+# minimal to assist readability and comprehensibility.
+
+# Contributions to this file are voluntarily given to
+# the cfengine community, and are moderated by CFEngine.
+# No liability or warranty for misuse is implied.
+#
+# If you add to this file, please try to make the
+# contributions "self-documenting". Comments made
+# after the bundle/body statement are retained in
+# the online docs
+#
+
+# For CFEngine Core: 3.9.0 or later
+
+# Test reporting bodies
+
+# TAP specification at
+# http://testanything.org/tap-specification.html
+
+###################################################
+# If you find CFEngine useful, please consider    #
+# purchasing a commercial version of the software.#
+###################################################
+
+@if minimum_version(3.9)
+
+bundle agent testing_ok_if(classname, message, error, trace, format)
+# @brief Report outcome for test on `classname` in format `format` with optional `error` and `trace`
+# @param classname The class name
+# @param message The test message
+# @param error The error to report, if the class is not defined
+# @param trace The error trace detail to report, if the class is not defined
+# @param format TAP (immediate output) or jUnit or anything else for delayed TAP
+#
+# **See also:** `testing_junit_report`, `testing_tap_report`, `testing_ok`, `testing_todo`, and `testing_skip`
+{
+  vars:
+      "next_testing_$(classname)" int => length(classesmatching("testing_.*", "testing_(passed|failed|skipped|todo)"));
+      "next_testing_$(classname)_failed" int => length(classesmatching("testing_.*", "testing_(passed|failed|skipped|todo)"));
+
+  classes:
+      "tap" expression => strcmp('tap', string_downcase($(format)));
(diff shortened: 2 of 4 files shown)


lib/junit.cf.mustache
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nickanderson: I would put this in a templates folder inside lib.

tzz: You mean `lib/templates/junit.mustache`?

nickanderson: Yeah thats what I was thinking. Thoughts?

I have done similar in cfe_internal/enterprise/templates

tzz: No problem. Fixed and pushed as a reworked commit.

tzz:
Yeah this will fail the 3.7 test build...

lib/stdlib.cf
```diff
@@ -71,6 +71,9 @@ bundle common stdlib_common
       "input[cfe_internal_hub]"           string => "$(this.promise_dirname)/cfe_internal_hub.cf";
       "input[cfengine_enterprise_hub_ha]" string => "$(this.promise_dirname)/cfengine_enterprise_hub_ha.cf";
 
+      # jUnit reports
+      "input[junit]"                      string => "$(this.promise_dirname)/junit.cf";
```
tzz: We could make this optional if the `junit` class is defined but that seemed unnecessary since they are all agent bundles that do nothing by default.

nickanderson: Yeah agent bundles get pre-eval for vars. But they should all be quickly skipped since the vars wont have values unless its passed as a parameter.

tzz: All right, give this a try on a 3.9 install... thanks for the review :)

tzz:
ping?

tzz:
@nickanderson updated with ideas from https://github.com/brontolinux/cfengine-tap to support both jUnit and TAP output, to a file or to STDOUT.

tzz:
@brontolinux please look at the latest code here, it has pretty good overlap with your cfengine-tap library. There's a usage example towards the end.

brontolinux:
Hey @tzz , all!

I have looked at your code. Some parts are not very clear to me (e.g., all chicanery with `format`, `mergedata`, `regexp_replace` and the like) but I didn't stop too long to understand them better, I know you know what you are doing and I have to rush to other stuff today ;-)

One thing that I don't like, which you can well consider a "minor" is that one gets the TAP report only after all the tests have run. It's OK for short tests but for very long ones (and I've seen some in the Perl world) it can be annoying. Annoyances are not necessarily bugs though.

Reasonably sized examples would also help greatly. Is there any plan to put some in the documentation?

Finally, I was considering to implement the [bail out](http://testanything.org/tap-specification.html#bail-out) directive in cfengine-tap when I get a little time to add the code and write an example. How hard would it be to implement it in your framework?

Ciao!
-- bronto

tzz:
@brontolinux responses:

* we can't have automatic test planning (the `1..N` count) if we produce the TAP output immediately. I think it's a platform limitation: the meta attributes are created only when we call the bundle, so there's no way to find out how many tests we'll have in advance.

* there's a pretty big example bundle. Do you mean you want a specific testing scenario? Or are the examples unclear?

* immediate bailing out can be supported. I'd implement it as simply another form of the finishing bundle, it's not hard.

brontolinux:
@tzz:
> we can't have automatic test planning (the `1..N` count) if we produce the TAP output immediately

You can: the protocol specifies that [the plan line can also be written at the end](http://testanything.org/tap-specification.html#the-plan), instead of the beginning, e.g. for those cases where the number of tests is dynamic:
> In certain instances a test file may not know how many test points it will ultimately be running. In this case the plan can be the last non-diagnostic line in the output.

.
> there's a pretty big example bundle. Do you mean you want a specific testing scenario? Or are the examples unclear?

You mean `bundle agent testing_usage_example`? If so, then I don't think it's good enough. Some close-to-real-life examples would be definitely more helpful in understanding how to use the framework in a real case.

> immediate bailing out can be supported. I'd implement it as simply another form of the finishing bundle, it's not hard.

Good stuff!!!

tzz:
@brontolinux @nickanderson OK, rebased and ready:

* TAP output to file is supported (it will be done entirely at the end)
* immediate TAP output to STDOUT is supported
* `testing_tap_bailout` bundle will do what you expect, but it won't bail out of the CFEngine bundle sequence (because I think that would be wrong in the CFEngine context)
* please submit real-world examples you think are useful, I can't think of anything better

nickanderson:
This is failing in 3.9.1 too.

```
    info: Cannot open file for hashing '/home/travis/cf_install/masterfiles/lib/junit.cf'. (fopen: No such file or directory)

   error: Can't stat file '/home/travis/cf_install/masterfiles/lib/junit.cf' for parsing. (stat: No such file or directory)
```

lib/stdlib.cf
nickanderson: I think this is supposed to be testing.cf now that you added tap support right?

tzz:
Duh, fixed. Thanks.

lib/stdlib.cf
```diff
@@ -71,6 +71,9 @@ bundle common stdlib_common
       "input[cfe_internal_hub]"           string => "$(this.promise_dirname)/cfe_internal_hub.cf";
       "input[cfengine_enterprise_hub_ha]" string => "$(this.promise_dirname)/cfengine_enterprise_hub_ha.cf";
 
+      # jUnit reports
```
nickanderson: jUnit and/or TAP reports ?

lib/testing.cf
```diff
@@ -0,0 +1,355 @@
...
+
+###################################################
+# If you find CFEngine useful, please consider    #
+# purchasing a commercial version of the software.#
+###################################################
+
+@if minimum_version(3.9)
+
+bundle agent testing_ok_if(classname, message, error, trace, format)
+# @brief Report outcome for test on `classname` with optional `error` and `trace`
```
nickanderson: with optional `error`, `trace`, and `format`

tzz: yeah, fixed in all the places, thanks!

lib/testing.cf
```diff
@@ -0,0 +1,359 @@
...
+# and body classes classes_generic from lib/3.6/common.cf
+# Diffing two files is quite common to test if two different versions
+# of the same policy produced the same result. This version of classes
+# generic will help.
+#
+# From man diff:
+# Exit status is 0 if inputs are the  same,  1  if
+# different, 2 if trouble.
+#
+# @brief Define `x` prefixed/suffixed with promise outcome, Set command to "kept" instead of "repaired" if it returns 0
```
nickanderson: Instead of having a full additional body can't we use inherit_from => classes_generic or more preferable results

tzz: That's a question for @brontolinux since it's his code. From my viewpoint this body can be moved out to `common.cf` since it's generically useful.

tzz:
I'm cutting out `testing_diff_classes_generic` since it's not used by anything else in `testing.cf`.

nickanderson:
I am adding it back to common.cf but slightly modified.

https://github.com/cfengine/masterfiles/pull/807

nickanderson:
@tzz can you  add `Changelog: Title` to your commit message please?

tzz:
OK, commit message amended. Thanks.

nickanderson:
Regarding the classes body tailored for diff: https://docs.cfengine.com/docs/master/reference-standard-library-common.html#diff_results

nickanderson:
And docs for the testing library are also published now: https://docs.cfengine.com/docs/master/reference-standard-library-testing.html

PR #767, opened 2016-07-18
Adapt autoconf code to autoconf 2.63.

kacf:
Autoconf 2.63 is lacking both m4_ifblank and m4_esyscmd_s, so work
around the former with m4_ifval (which has reversed semantics) and the
latter with m4_normalize.

Also bump the version requirement while we're at it. Although I don't
know if the new macros were introduced in 2.63, we at least know they
are present there. Bump to 2.63 because:

* 2.64 has been (secretly) required for a long time, since the macros
  I replaced in this commit were only available in that version and
  later. I doubt anyone is using anything older.

* 2.63 is already eight years old.

* At least 2.63 is tested, we don't know about the older ones.

* It doesn't affect tarballs.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,20 +6,20 @@ dnl # Run ./autogen.sh to build configure script
 dnl #
 dnl ##########################################################################
 
-AC_PREREQ(2.59)
+AC_PREREQ(2.63)
 
-m4_define([revision], m4_esyscmd_s([git rev-list -1 --abbrev-commit HEAD]))
+m4_define([revision], m4_normalize(m4_esyscmd([git rev-list -1 --abbrev-commit HEAD])))
 
-m4_define([cfversion_from_env], m4_esyscmd_s([echo $EXPLICIT_VERSION]))
-m4_ifblank(cfversion_from_env, [
-    m4_define([cfversion_from_detect], m4_esyscmd_s([3rdparty/core/determine-version.py]))
+m4_define([cfversion_from_env], m4_normalize(m4_esyscmd([echo $EXPLICIT_VERSION])))
+m4_ifval(cfversion_from_env, [
+    m4_define([cfversion], cfversion_from_env)
+], [
+    m4_define([cfversion_from_detect], m4_normalize(m4_esyscmd([3rdparty/core/determine-version.py])))
     m4_if(m4_sysval, 0, [], [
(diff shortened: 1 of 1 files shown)


kacf:
Will submit this to the other repos as well if this passes, in order to keep them similar.

kacf:
trigger build

kacf:
Darn, Jenkins is currently down.

configure.ac
```diff
@@ -6,20 +6,20 @@ dnl # Run ./autogen.sh to build configure script
 dnl #
 dnl ##########################################################################
 
-AC_PREREQ(2.59)
+AC_PREREQ(2.63)
```
jimis: Are we using specific macros from 2.63? If so it should be listed in the commit message.

kacf: No, the original was using macros from 2.64 (and has been for a long time). The new macros I use are present in 2.63, but I don't know if they were introduced there. I bumped to 2.63 mainly because:

* 2.64 has been (secretly) required for such a long time, I doubt anyone is using anything older.
* 2.63 is already eight years old.
* At least 2.63 is tested, we don't know about the older ones.
* It doesn't affect tarballs.

All in all I just thought it was a good time to bump. I can document it the way I described in the commit message, or if you think it's a bad idea I can also remove it.

jimis: > At least 2.63 is tested, we don't know about the older ones.

This is the argument that persuades me. :-) Better know for sure that what we require works. Commit +1 from me, but please mention it in the commit message.

jimis: BTW do we really test 2.63? Our "bootstrap" buildslave has autoconf-2.68.

kacf: Uhm no. :-)

Read it as "has been tested", not as "being continuously tested"! If we want to be sure we could bump it even higher, but it works on 2.63 right now and this stuff changes very rarely.

I think I'll do what you suggested, leave as is and edit the commit message. I'll be submitting the same patch to all other repos as well, to keep them in sync.

jimis: ok!

kacf: Commit message updated!

kacf:
trigger build

kacf:
Also cherry-picked to 3.7.x and 3.9.x, as well as the same branches on all the other repositories with a `configure.ac` file.

PR #772, opened 2016-07-19
Add: Inventory for system product name (model)

nickanderson:
Jira #ENT-2780


nickanderson:
Needs to be merged with https://github.com/cfengine/mission-portal/pull/282

mikeweilgart:
Wow, that's simpler than I expected it would be.

nickanderson:
pretty simple :)

PR #781, opened 2016-08-02
Introduce `make tar-package`.

jimis:


jimis:
@nickanderson review? :-)

Now how do we generate and distribute it?

jimis:
Merging, needs to be cherry-picked to stable branches in order to generate package tarballs there.

PR #784, opened 2016-08-03
Improve Travis-CI usage.

jimis:


jimis:
OK this seems to be ready. I'm not so sure how deployments will work when we tag, we need @maciejmrowiec 's opinion on this one. I believe nothing will happen now, since we never tag the master branch, but if we cherry-pick this `deploy:` snippet to branches, it should trigger. I wonder if there is a way to test it, without actually tagging anything.

.travis.yml
```diff
@@ -1,15 +1,56 @@
...
+
+    # JOB 2
+    - env: CF_VERSION=lts
+      addons:
+        apt:
+          sources:
+            - sourceline: "deb https://cfengine-package-repos.s3.amazonaws.com/pub/apt/packages stable main"
+            - key_url: https://cfengine-package-repos.s3.amazonaws.com/pub/gpg.key
+          packages:
+            - cfengine-community=3.7.3-1
```
nickanderson: not for long ;)

maciejmrowiec:
@jimis you need this config in branch that you are building and they can differ which is helpful with having adequate compatibility matrix.

PR #787, opened 2016-08-19
CFE-2278 Split systemd units

nickanderson:
This introduces seperate systemd units for each CFEngine component.

cf-apache.service
cf-consumer.service
cf-execd.service
cf-hub.service
cf-monitord.service
cf-postgres.service
cf-redis-server.service
cf-runalerts.service
cf-serverd.service

cfengine3.service acts as an umbrella service

NOTE: Currently cfengine components are configured to run with verbose output.
I expect that this will be changed before merging unless there is significant
user support for leaving in verbose mode by default for easier inspection.

The `cfe_internal_update_processes` bundle is responsible for determining if the
cfengine services are `systemd_supervised` (based on the presence of the
`systemd` class and not running an older version (3.6 through 3.9). It is also
responsible for determining which components should be managed, enabled or
disabled. For example only enterprise policy servers would manage the
`cf-apache`, and `cf-postgres` services, and `cf-hub` and `cf-runalerts` should
be disabled on enterprise HA hubs that are not the active hub.

The `cfe_internal_systemd_service_unit_files` bundle is responsible for
rendering the appropriate systemd units, setting their permissions, reloading
systemd when changes are made, and restarting any service currently running whos
unit has changed.

The `cfe_internal_systemd_service_unit_state` bundle is responsible for managing
the correct state for each managed cfengine service unit. It leverages
`u_systemd_services` which was extracted from the standard library for use in
the update policy.

These policy changes were tested against a 3.7.4 hub.

Note: The split systemd units are only manged when the systemd_supervised class
is definded. So if you want to try this on a centos 7 3.7.4 hub simply ensure
that systemd_supervised gets defined for your runs.


nickanderson:
@jimis @GregorioDiStefano @kacf

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -121,7 +121,7 @@ bundle agent cfe_internal_update_folders
 
   files:
 
-      "$(sys.workdir)/master_software_updates/$(dirs)/." -> { "goal_updated" }
+      "$(sys.workdir)/master_software_updates/$(dirs)/."
```
tzz: unrelated

tzz:
Looking at this, it's very complicated and does several unrelated things. I would at least split it into the systemd piece and the rest.

The systemd code really worries me. That's a lot of code copied with the `u_` prefix. Maybe that can be avoided?

Rather than writing dynamic **code** with Mustache templates, I would consider writing data files with Mustache and reading them into static code. That's much less likely to break, and bugs or misbehavior are not as fatal.

Sorry if I don't understand the way everything fits together, this was just my reaction after reading the PR.

nickanderson:
Hum, I thought this was much less complex than the other existing process management bundles.

One bundle is responsible for figuring out which services should be in what state.
One bundle is present to manage the unit configuration (the files).
One bundle is present to manage the service states.

And yes, other shuffling was mixed in as I went.

I agree about the u_systemd_services bundle. But it's also risky to share policy between seperate entries. I brought this up with @estenberg and the we decided to keep the current pattern of not loading any of the stdlib from update.cf.

What do you mean writing dynamic code with mustache? Are you talking about the systemd unit templates, or the one template for the runalerts service?

I thought that way of managing the systemd units would make it relatively easy to maintain by users if necessary.

For runalerts, it was previously a bash script that was promised with an edit_line block. I switched from bash to php to make the little loop work with systemd, and there is still probably better ways to handle it, but I did think a mustache template was more clear than a deeply burried edit_line.

tzz:
> What do you mean writing dynamic code with mustache? Are you talking about the systemd unit templates, or the one template for the runalerts service?

The runalerts. It's easier to let users hack it your way, sure. I didn't know that was the intended use case--it's a good approach, thanks for explaining.

tzz:
Regarding systemd management, considering the amount of CFE code dedicated to it, the critical nature of systemd service modifications, and the growing market share of systemd, I would consider putting some facilities for systemd service management in the C core. For now, just functions that can probe: list services, list units, get details, that kind of thing--these are annoyingly verbose in CFE. And maybe put some of the info in the `sys` namespace by default, like we did with the `/proc` and `/proc/net` parsing. If you agree, open a ticket and we can spec it out :)

BasementTrix:
On Mon, Aug 22, 2016 at 08:02:53AM -0700, Ted Zlatanov wrote:
> Regarding systemd management, considering the amount of CFE code
> dedicated to it, the critical nature of systemd service
> modifications, and the growing market share of systemd,

Just remember that not everybody runs systemd and that what you are
considering for cthe C core shouldn't adversely affect systems on which
systemd is neither native nor installed.

Not all CFEngine systems run Linux.

-- 

tzz:
@BasementTrix thanks, we can discuss all of that in the ticket if @nickanderson wants to propose the feature. FWIW I agree with you; my work on the procfs and Linux-specific `/proc/net` parsing hasn't affected any non-Linux systems adversely and I'd expect the same here.

templates/cf-apache.service.mustache
```diff
@@ -0,0 +1,18 @@
...
+Type=forking
+ExecStart={{{vars.sys.workdir}}}/httpd/bin/apachectl start
+ExecStop={{{vars.sys.workdir}}}/httpd/bin/apachectl stop
+PIDFile={{{vars.sys.workdir}}}/httpd/logs/httpd.pid
+Restart=always
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
+WantedBy=cfengine3.service
```
kacf: Why keep an additional set of files here in masterfiles? This just makes it more complicated, and messes with files that the package manager is ultimately responsible for. I'm sure at some point someone will make a fix to those files in core and forget it here.

nickanderson: The reason for having the templates in masterfiles is so that they can be easily tweaked by users. For example to run cf-serverd in verbose mode you can simply update the template and the unit will be re-rendered and the service re-started.

ageorgop: I know I'm late to the party but just looking over the code and I notice we are dropping the units in /lib vs /etc.  I was always under the impression that custom systemd units should end up in /etc and not clobber the ones in /lib.

https://www.freedesktop.org/wiki/Software/systemd/FrequentlyAskedQuestions/

> Q: I want to change a service file, but rpm keeps overwriting it in /usr/lib/systemd/system all the time, how should I handle this?

> A: The recommended way is to copy the service file from /usr/lib/systemd/system to /etc/systemd/system and edit it there. The latter directory takes precedence over the former, and rpm will never overwrite it. If you want to use the distributed service file again you can simply delete (or rename) the service file in /etc/systemd/system again.

nickanderson: Yes, I have the same thing in my own notes. This is actually disabled right now (the bundles are never called). I was still debating on actively managing the unit states by default or not. Figured I would start off with not having them active and while testing was going on.

nickanderson: And I appreciate the note, thanks!

PR #788, opened 2016-08-26
Fix double-parsing of options

mikeweilgart:
Passing un-double quoted $options to eval is not what was intended here.

The contents of "options" in the previous version will be subjected to split, glob, and then *again* split, glob.


mikeweilgart:
@nickanderson

PR #789, opened 2016-08-31
Travis whitespace check

jimis:
This is just an intro, adding a very simple and very quick trailing-whitespace check for pull requests.

The purpose is to extend this with lint, coding style checks and nice checks for the C code if we port it to Core.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: c
 matrix:
   include:
 
+    # JOB 0  --  first because it's the fastest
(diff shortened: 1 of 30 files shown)


jimis:
This seems to work good, check the "details" button above, it should show 3 tests now. 

@nickanderson please review and merge. :-)

jimis:
I'm merging just to make your life harder @nickanderson ;-)

nickanderson:
I like it

PR #790, opened 2016-09-02
Make package manager unit tests parallel-safe.

kacf:


kacf:
Cherry-picked to 3.7.x and 3.9.x as well.

PR #797, opened 2016-09-08
Test: Add same class guard for CentOS 5 as for Redhat 5.

kacf:


kacf:
Cherry-picked to 3.7.x and 3.9.x as well.

PR #798, opened 2016-09-09
Fix for CFE-2445

ageorgop:
Not sure about style but this fixes the issue.  Let me know if you need me to adjust it.

### DIFF
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -8,7 +8,6 @@ do
 	if [ -x $python_bin ];
 	then
 		PYTHON_BIN="$python_bin"
-		echo $python_bin
 		break
 	fi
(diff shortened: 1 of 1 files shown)


kacf:
Nice! Looks good to me in principle, but need to wait for Jenkins to be back up to test this.

ageorgop:
I had a feeling about issue 1 after I submitted the PR.  Now for 2 I will try and think of a cleaner way to do this.

kacf:
Thanks, @ageorgop!

kacf:
I haven't looked at this in a while. Has there been any luck in resolving point 2? From looking at it, it seems like stdin is still occupied by the pipe that feeds the script.

PR #799, opened 2016-09-13
CFE-2458: Allow specification of files_single_copy via augments

nickanderson:
Changelog: Title

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -345,7 +345,7 @@ of an augments file.
 
 For example:
 
-```
+```json
 {
   "vars":{
     "control_common_bundlesequence_end": [ "mybundle1", "mybundle2" ]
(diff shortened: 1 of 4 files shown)


nickanderson:
@jimis or @kacf any idea about this autogen failure all the sudden?

kacf:
Can't say that I do, but it must be related to the `determine-version.sh` script. Can you try running it locally while at the same commit and see what result it gives?

nickanderson:
@kacf it works locally on that commit, and it also works in 3.10.x

kacf:
I can't reproduce it even in a Ubuntu 12 docker instance, which is what Travis uses. Also the message `fatal: No tags can describe '9884667a989ffa87f4975df25a5d00113ce5cf8e'.` is completely bogus, clearly it can be described. I don't have any other explanation than that Travis is using a buggy version of Git. Maybe you can tweak exactly which container Travis uses for doing the build.

nickanderson:
@jimis thoughts?

PR #800, opened 2016-09-13
CFE-2460: Add ability to append to bundlesequnece with def.json

nickanderson:
Changelog: Title

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -316,3 +316,29 @@ This class can be defined by an augments file. For example:
 This class enables policy that cleans up report diffs when they exceed
 `def.maxclient_history_size`. By default is is **off** unless a CFEngine
 Enterprise agent is detected.
+
+
+## Main Policy (promises.cf)
+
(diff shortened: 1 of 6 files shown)


tzz:
YES for me. Maybe implementing this in C makes sense, and the setting name is really long, but I like it.

nickanderson:
@jimis this seems like a regression in core. Note that LTS passes, while master fails. I might be able to do some policy acrobatics to get around this, but I would rather not do acrobatics when it looks like a regression.

jimis:
Travis is failing this pull request with CFEngine 3.9.1, if this is a bug in Core it should be fixed there first. Else all Travis builds will keep coming red.

nickanderson:
@tzz Yeah the name is long. But it follows a pattern, and works with 3.7+. I just kept running into issues when I was trying to use a data container as suggested in #730

nickanderson:
rebased

PR #801, opened 2016-09-13
CFE-2461: Allow specifying agent maxconnections via def.json

nickanderson:
Changelog: Title


nickanderson:
Picked to 3.9.x 3.8.x and 3.7.x

PR #802, opened 2016-09-14
CFE-2468: Pass package promise options to underlying apt-get call

jpvlsmv:
Changelog: Title

Commit can be cherry-picked cleanly into 3.7.x, 3.8.x, 3.9.x as well.

### DIFF
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -251,6 +251,10 @@ def package_arguments_builder(is_apt_install):
         elif line.startswith("Architecture="):
             arch = line.split("=", 1)[1].rstrip()
 
+        elif line.startswith("Options="):
+            global apt_get_options
+            apt_get_options.append(line.split("=", 1)[1].rstrip())
+
     if name:
         args.extend(one_package_argument(name, arch, version, is_apt_install))
 
@@ -258,9 +262,9 @@ def package_arguments_builder(is_apt_install):
 
 
 def repo_install():
-    cmd_line = [apt_get_cmd] + apt_get_options + ["install"]
-
     args = package_arguments_builder(True)
+
+    cmd_line = [apt_get_cmd] + apt_get_options + ["install"]
 
     if (not args):
         return 0
@@ -276,9 +280,9 @@ def repo_install():
     return 0
 
 def remove():
-    cmd_line = [apt_get_cmd] + apt_get_options + ["remove"]
-
     args = package_arguments_builder(False)
+
+    cmd_line = [apt_get_cmd] + apt_get_options + ["remove"]
 
     if (not args):
         return 0
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -251,7 +251,7 @@ def package_arguments_builder(is_apt_install):
         elif line.startswith("Architecture="):
             arch = line.split("=", 1)[1].rstrip()
 
-        elif line.startswith("Options="):
+        elif line.startswith("options="):
             global apt_get_options
             apt_get_options.append(line.split("=", 1)[1].rstrip())
 
--- a/tests/unit/test_package_module_apt_get
+++ b/tests/unit/test_package_module_apt_get
@@ -117,6 +117,14 @@ dpkg-query --showformat ${Architecture}=${Status}
 apt-get -o Dpkg::Options::=--force-confold -o Dpkg::Options::=--force-confdef -y --force-yes install a:x=1 b:y c=3 netcat:amd64 netcat:i386 netcat:amd64=3 netcat:i386=3 d
 """])
 
+assert check("repo-install", [ "Name=netcat",
+                              "Options=-qq"],
+             0, [], 4, ["""dpkg --print-architecture
+dpkg-query --showformat ${Architecture}=${Status}
+ -W netcat:*
+apt-get -o Dpkg::Options::=--force-confold -o Dpkg::Options::=--force-confdef -y --force-yes -qq install netcat:amd64 netcat:i386
+"""])
+
 assert check("remove", ["Name=a\nVersion=1\nArchitecture=x",
                         "Name=b\nArchitecture=y",
                         "Name=c\nVersion=3",


nickanderson:
Perhaps I am doing something wrong. I don't think this is working.

If I use this policy with your updated module screen is still installed.

```
  packages:
    "screen"
      policy => "present",
      options => { "APT::Get::Download-Only" };
```

I get the same behaviour with:

```
  packages:
    "screen"
      policy => "present",
      options => { "--download-only" };
```

For the first to work I think when executed apt needs to be called with `-o APT::Get::Download-Only` and for the second it would be simply appended.

I wonder how this difference in "options" should be handled.

It seems that any configuration option can be specified by the first format and only a subset of options have command line arguments associated. So maybe it makes sense to only support the first. Otherwise I think the module will need to differentate between the types either with a specific key like `args=--download-only` and `config=APT::Get::Download-Only`

jpvlsmv:
For ease of implementation, I didn't try to determine which  command line options need a -o and which ones stand alone.

Options => [ "-o", "APT::Get::Download-Only=true" ];

They're just passed right on to the popen () calls in the python scripts.

neilhwatson:
My Python is not great, but I don't see where those arguments are checked for insecure options like ';' and '&'.

nickanderson:
I am not a pythonista but i played around and came up with the following patch.

It will exit with an error if a dangerous character is found in an Options= line.

Additionally it trys to be smart about the `-o`. If the option starts with a `-` then `-o` is not automagically used, but if the option doesn't start with `-` its automagically used.

```
diff --git a/modules/packages/apt_get b/modules/packages/apt_get
index 66dfb8b..b7b0cd9 100755
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -252,8 +252,22 @@ def package_arguments_builder(is_apt_install):
             arch = line.split("=", 1)[1].rstrip()
 
         elif line.startswith("Options="):
-            global apt_get_options
-            apt_get_options.append(line.split("=", 1)[1].rstrip())
+            # If the option starts with a - we assume this will be a command argument
+            # We abort if we see any dangerous looking characters.
+            if (';' in line) or ('&' in line) or ('$' in line) or ('`' in line):
+                sys.stderr.write("Dangerous character(s) found in Options. Aborting package module\n")
+                return 2
+            else:
+                opt = options.split("=", 1)[1].rstrip()
+                global apt_get_options
+                if opt.startswith("-"):
+                    # If the option starts with a - we assume it will be a
+                    # command argument and it will be passed directly on
+                    apt_get_options.extend([ options.split("=", 1)[1].rstrip() ])
+                else:
+                    # For usability we take care of appending the -o to the
+                    # command automatically.
+                    apt_get_options.extend(["-o", options.split("=", 1)[1].rstrip() ])
 
     if name:
         args.extend(one_package_argument(name, arch, version, is_apt_install))
```

nickanderson:
@jpvlsmv ref: https://github.com/jpvlsmv/masterfiles/pull/1

jpvlsmv:
I'm not sure that there is a security risk created by passing tokens from cfengine to one of its modules... But since Python calls `subprocess_Popen(cmd, stdout, stderr)` and the cmd parameter is an array, there is no subshell involved.  (Unless you explicitly pass shell=True, or are on Windows where the underlying CreateProcess() call only operates on strings)

https://docs.python.org/2/library/subprocess.html#popen-constructor

I don't mind the additional checks either way.

jpvlsmv:
As for automatically -o'ifying options, there are other apt-get command-line options: `-c config_file`, `-t target_release`, `-a architecture` which are not required to be conjoined to their optarg by an =, (just like -o isn't conjoined).  It is complex to remember "APT::Get::Download-Only=True" gets -o'ified, and "-c=/mnt/target/etc/apt.conf" does not, but make sure you don't forget the = sign there, because if you do, it won't work even though you can omit it at the shell.

There should be limited use (and this is 100% package-module-specific) for such options anyway... It's an odd case to turn off HTTPS certificate checking for a specific package install, and one definitely wouldn't expect `yum` to understand this APT::whatever option flag.

And I would say that it goes against the documented behavior of the options:

> This attribute has no inherent meaning to CFEngine, and **will be passed verbatim**. It is meant as a mechanism to communicate special attributes to the package module that are not covered by the main API. For example, for certain package modules it may be used to pass a repository URL. (https://docs.cfengine.com/lts/reference-standard-library-package_modules.html#options-attribute, emphasis added)

nickanderson:
Spoke with @estenberg and he agrees with you @jpvlsmv . I asked for additional feedback from the debian team and the dev-cfengine mailing list.

estenberg:
Yes, I think we should keep it simple in this case to avoid confusion; we can simply say that the apt-get manual documents the exact arguments.

Are the options used in all calls to apt-get? E.g. on remove, update, add? And will all these calls accept the same options? Or do we expect the end user to supply options depending on what he needs to do (I suppose download only is a good example of this; only relevant for present?).

jpvlsmv:
And interestingly, setting an options => [ "anything" ] results in the package detection code to choke with "   error: Mixed old and new package promise attributes inside one package promise. " from core/verify_packages.c@227.

jpvlsmv:
--download-only would not converge to "present", though... :)  

The apt_get_options python variable is only used when calling from repo_install and remove, which should be disjoint enough in policy that they would not be confusing.  You might need `policy => "absent", options => {"--purge"}` sometimes, but that wouldn't get in the way of `policy => "present", options => {"--build"}`

kacf:
This looks like a good improvement to me, consider me in favor!

nickanderson:
We should add docs for this.

Any thoughts on where a section for specific package modules should be found? The first place I thought of was in the packages reference page and adding a an entry for Package Modules -> apt_get

estenberg:
Yes, good point with the docs @nickanderson.
As it is now I think the "Package Modules" doc (https://docs.cfengine.com/lts/reference-standard-library-package_modules.html) is more for the developer of the module, not the user. This split makes sense.

I would suggest adding it to the packages reference page for now (at least until we have many package modules).

E.g.

**Package** modules **out-of-the-box**
CFEngine comes with out-of-the-box support for the following package managers...
apt_get: options are passed directly to...

nickanderson:
@jpvlsmv When I try this unexpectedly the promise actually installs the package. I expected the promise would not converge and would either error because the package didn't get installed (I believe the agent checks this internally) or repaired because the command returned 0.

```
  packages:
    "screen"
      policy => "present",
      package_module => apt_get,
      options => { "-o", "APT::Get::Download-Only=true" };
```

```
root@nickanderson-ThinkPad-W550s:/var/cfengine/inputs# apt-cache policy screen
screen:
  Installed: (none)
  Candidate: 4.3.1-2
  Version table:
     4.3.1-2 0
        500 http://us.archive.ubuntu.com/ubuntu/ wily/main amd64 Packages
        100 /var/lib/dpkg/status
root@nickanderson-ThinkPad-W550s:/var/cfengine/inputs# cf-agent -KIf ./promises.cf -b main
    info: Using command line specified bundlesequence
    info: Successfully installed package 'screen'
```


Running the following does work as expected. The package is NOT installed and the command returns 0

```
root@nickanderson-ThinkPad-W550s:/usr/bin/apt-get -o Dpkg::Options::=--force-confold -o Dpkg::Options::=--force-confdefs -y --force-yes -o APT::Get::Download-Only=true install screen
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following package was automatically installed and is no longer required:
  mysql-common
Use 'apt-get autoremove' to remove it.
Suggested packages:
  iselect screenie byobu
The following NEW packages will be installed:
  screen
0 upgraded, 1 newly installed, 0 to remove and 4 not upgraded.
Need to get 0 B/567 kB of archives.
After this operation, 971 kB of additional disk space will be used.
Download complete and in download only mode
root@nickanderson-ThinkPad-W550s:/var/cfengine/inputs# echo $?
0
```

Downloadonly is an edge case for sure, but can you give that a try and see if you get the same behaviour?

kacf:
I got some spurious (but seemingly unrelated) errors when running this through Jenkins. So it looks like it passes tests.

jpvlsmv:
I see the problem now. https://github.com/cfengine/core/blob/master/cf-agent/package_module.c#L227 sends in the option value as option=whatever instead of Option=whatever.

Pull request branch updated to use the lower-case version.

jpvlsmv:
Policy testing is described at https://github.com/jpvlsmv/masterfiles/wiki/Troubleshooting-apt_get-module

nickanderson:
Thats great. Love your notes too. Now its erroring on me with dry run (since the package wasn't). @kacf already approved this so I will merge to master. I am OK with picking this back to 3.{9,8,7} as well but will wait for some additional feedback in the ticket

kacf:
@nickanderson: Did you force push after you merged? This pull request has references to commits that don't exist in the repository...

kacf:
Followup commit to fix broken test: https://github.com/cfengine/masterfiles/pull/808

Always rerun tests when there has been changes.

nickanderson:
No I didn't force push. At least I don't think I did. It definitely wasn't intentional and I rarely force push except when squashing.

kacf:
Ok, weird...

PR #803, opened 2016-09-16
CFE-2466: Change: Set tests to soft fail for package building

nickanderson:


jimis:
Thanks for this, inspite of you being on vacation!

jimis:
please merge

PR #806, opened 2016-09-28
Clarify comment on "ifelapsed" in controls/cf_agent.cf

atsaloli:
It was confusing because it said "ifelapsed" takes an interval, but it doesn't; it takes an integer.


nickanderson:
Thanks! Picked to 3.9.x, 3.8.x, and 3.7.x

PR #807, opened 2016-09-28
Add: Classes body tailored for use with diff

nickanderson:
Changelog: Title


nickanderson:
@bronto @tzz

tzz:
I love it. I would pass a scope parameter through, though.

nickanderson:
you mean diff_results(scope, x)

tzz:
Yes, exactly. I wish there was a way to default body parameters! But lacking that, yeah, let the user specify it.

PR #811, opened 2016-10-05
CFE-953 policy changes for custom ports.

olehermanse:
@nickanderson Would like review on this. Some changes to make custom ports work for cf-serverd and cf-agent. Not sure if this is the best way, but it seems to be working.

Ref: cfengine/core#2699

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -216,8 +216,10 @@ body copy_from u_rcp(from,server)
       trustkey    => "false";
 
     !am_policy_hub::
-
       servers => { "$(server)" };
+
+    !am_policy_hub.sys_policy_hub_port_exists::
+      portnumber => "$(sys.policy_hub_port)";
 
     cfengine_internal_encrypt_transfers::
       encrypt => "true";
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ stamp-h1
 /revision
 /ylwrap
 /test-driver
+CFVERSION
 
 # Misc.
 /cf_promises_*
--- a/cfe_internal/update/update_bins.cf
+++ b/cfe_internal/update/update_bins.cf
@@ -323,7 +323,7 @@ bundle agent cfe_internal_update_bins
     !am_policy_hub.enterprise.trigger_upgrade.(cfengine_3_6_0|cfengine_3_6_1)::
       "$(sys.bindir)/cf-upgrade"
       handle => "cfe_internal_update_bins_files_cf_upgrade_i386_linux",
-      copy_from => u_rcp("$(master_software_location)/cf-upgrade/linux.i386/cf-upgrade", @(update_def.policy_servers)),
+      copy_from => u_rcp("$(master_software_location)/cf-upgrade/linux.i386/cf-upgrade", @(update_def.policy_servers), "$(sys.policy_hub_port)"),
       perms => u_m("0755"),
       ifvarclass => "linux.i686",
       comment => "The cf-upgrade binary that shipped with 3.6.0 and 3.6.1 was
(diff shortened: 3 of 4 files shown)


cfe_internal/update/update_policy.cf
```diff
@@ -95,7 +95,7 @@ bundle agent cfe_internal_update_policy
       "$(inputs_dir)/cf_promises_validated"
       comment => "Check whether a validation stamp is available for a new policy update to reduce the distributed load",
       handle => "cfe_internal_update_policy_check_valid_update",
-      copy_from => u_rcp("$(master_location)/cf_promises_validated", @(update_def.policy_servers)),
+      copy_from => u_rcp("$(master_location)/cf_promises_validated", @(update_def.policy_servers), "$(sys.policy_hub_port)"),
```
neilhwatson: Assumes all servers in the list will use the same port. Might it be better if servers element could take the form of 10.0.0.0 or 10.0.0.0:5309 and do away with the port parameter.

olehermanse: Yes we discussed this, and agree that this would be better. But it will take some time to implement.

olehermanse:
@jimis @nickanderson Would like your input as well.

nickanderson:
No this won't work because it's not backwards compatible. If this policy is run on a host running a version that does not have sys.policy_hub_port  they will fail.

We very much expect that this version of masterfiles will be run on prior versions of the agent especially during upgrades.

nickanderson:
A thought to keep policy server and port bound closely would be to populate a data container where the keys are the addresses and the values are the port.

jimis:
> What will happen if you make no changes to masterfiles and someone bootstraps to an ip with a port? It will fail when it comes around to update right?

If these changes to masterfiles are not committed then bootstrap succeeds and update fails, that's right.
Furthermore on the policy hub itself, the `cf-serverd` will not be listening to the proper port.

jimis:
@nickanderson thanks for reviewing this, I can now see that this is very important:

> No this won't work because it's not backwards compatible. If this policy is run on a host running a version that does not have sys.policy_hub_port they will fail.

I'm not sure how exactly to address this, besides what you are suggesting. It might be a good discussion topic for the cabin trip!

> A thought to keep policy server and port bound closely would be to populate a data container where the keys are the addresses and the values are the port.

The keys are the addresses? Hmmm indeed that would be quite elegant way of having multiple policy hubs defined. However this is regarding @neilhwatson 's comment and by no means does it block this pull request. In fact it is a problem in every `body copy_from` with multiple servers, so I'd suggest these ideas go in a relevant ticket. :-)

nickanderson:
Looking at this again, I think there is an easier way to add this support.

Until now the MPF has assumed port `5308`. If you wanted to change that, you
have to edit the MPF and update the `copy_from` body to specify a port.

If this feature introduces a feature class (similar to
`feature_def_json_preparse`) I think that we could simplify these changes so
that they only affect the `copy_from` body itself, instead of making changes to
the parameters it takes.

For example:

    body copy_from u_rcp(from,server)
     {
           source      => "$(from)";
           compare     => "digest";
           trustkey    => "false";
    
         !am_policy_hub::
    
           servers => { "$(server)" };
    
         !am_policy_hub.feature_bootstrap_specific_port::
           portnumber => "$(sys.policy_hub_port)";
    
         cfengine_internal_encrypt_transfers::
           encrypt => "true";
    
         cfengine_internal_purge_policies::
           purge => "true";
    
         cfengine_internal_preserve_permissions::
           preserve => "true";
    }

This way old clients not having the feature will behave exactly the same as
today. New clients that have the feature, will use the port that populates
`sys.policy_hub_port` (and `sys.policy_hub_port` will be `5308` if not specified
in `policy_server.dat`.

It might be more elegant to apply a constraint to the body attribute directly,
but that's not valid in the language.

    bundle agent main
    {
      reports:
          "test";
    }
    
    body copy_from test
    {
          portnumber => $(sys.policy_hub_port)
            ifvarclass => isvariable("sys.policy_hub_port");
    }

Then for future development:

1.  Populate `sys.policy_servers` with a map of addresses to ports. Initially
    this would look like this:
    
        {
            "$(sys.policy_hub)": "$(sys.policy_hub_port)"
        }

2.  Support multiple addresses in `$(sys.workdir)/policy_server.dat` in the
    following format:
    
        IP1:port
        IP2:port
    
    -   `sys.policy_hub` should contain the first entry in the file
    
    -   `sys.policy_hub_port` contain the port of the first entry in the
        file (defaulted appropriately if not specified).

tzz:
@nickanderson you can try:

```
body copy_from test
{
      portnumber => ifelse(isvariable("sys.policy_hub_port"), $(sys.policy_hub_port), 5308);
}
```

olehermanse:
@nickanderson I agree, and the future stuff looks good. Will work with you in person to get the immediate solution. 

Small note - the new `policy_server.dat` format is:
```
host:port
```
where host can be ip or host name, and lookup will be done for hostnames.

PR #812, opened 2016-10-07
apt_get option --force-yes is deprecated since version 1.1

sberthelot:
Check bugreport CFE-2360
This PR changes options unconditionally. Older apt_get versions (<1.1) if supported by current cfengine LTS masterfiles should be checked to change options accordingly...

### DIFF
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -16,7 +16,9 @@ dpkg_status_format = "Status=${Status}\n" + dpkg_output_format
 apt_get_options = ["-o", "Dpkg::Options::=--force-confold",
                    "-o", "Dpkg::Options::=--force-confdef",
                    "-y",
-                   "--force-yes"]
+                   "--allow-downgrades",
+                   "--allow-remove-essential",
+                   "--allow-change-held-packages"]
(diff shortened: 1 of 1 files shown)


nickanderson:
I think that we should detect the version and DTRT.

Perhaps something like this:

```
import sys
import subprocess
from distutils.version import LooseVersion


apt_version = subprocess.Popen(['/usr/bin/apt-get', '-v'], stdout=subprocess.PIPE).communicate()[0]

apt_version = apt_version.splitlines()[0].split(' ')[1]

if LooseVersion( apt_version ) < LooseVersion( "1.1" ):
	sys.stdout.write("Running version " + apt_version + " that supports --force-yes\n")

if LooseVersion( apt_version ) > LooseVersion( "1.1" ):
	sys.stdout.write("Running version " + apt_version + " that does not support --force-yes\n")


```

nickanderson:
Hey @sberthelot thanks a lot for the suggestion and work on implementation.

We did end up implementing this in a separate pull request:

Please reference:
https://github.com/cfengine/masterfiles/blob/7e13f4ff388fecd62ccf1043519e9e9aac820953/modules/packages/apt_get#L34-L45

I am going to go ahead and close out this PR since it was superseded, but please let us know if that works in your testing as well.

Thanks!

PR #815, opened 2016-10-20
Skip tests that cause RPM DB corruption on old OSes.

jimis:


jimis:
Test build is mostly green, besides a couple of spurious rsync errors. Merging.

PR #816, opened 2016-10-25
CFE-2609: lib/event.cf: add event handling framework

tzz:
@nickanderson similar to `testing.cf`

@yann-soubeyrand maybe this is useful regarding your questions about event handling...

### DIFF
--- a/lib/event.cf
+++ b/lib/event.cf
@@ -1,0 +1,205 @@
+############################################################################
+#  Copyright (C) CFEngine AS
+#
+#  This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


tzz:
@yann-soubeyrand updated with event persistence as discussed

nickanderson:
@tzz can i haz a test for this?

tzz:
I was waiting for a review of the API. Does the library look reasonable?

nickanderson:
I think it seems reasonable, it reminds me very much of the [feature bundles](https://github.com/cfengine/masterfiles/blob/master/lib/feature.cf)

PR #818, opened 2016-10-26
ENT-2831: Fix bootstrap hanging on non-systemd platforms.

kacf:


kacf:
Cherry-picked to 3.7.x and 3.10.x as well.

kacf:
Err, I mean 3.10.x only.

PR #824, opened 2016-10-31
sys.policy_hub renamed to sys.policy_hub_ip

olehermanse:
Must merge with: cfengine/core#2727

This change is done so I can reintroduce sys.policy_hub to be same as policy_server.dat.

### DIFF
--- a/cfe_internal/enterprise/ha/ha_def.cf
+++ b/cfe_internal/enterprise/ha/ha_def.cf
@@ -31,7 +31,7 @@ bundle common ha_def
 
 
   classes:
-      "ha_replication_only_node" expression => some($(sys.policy_hub), replication_only_nodes);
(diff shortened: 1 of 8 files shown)


olehermanse:
@nickanderson Please review, wait for core PR before merging.

olehermanse:
@nickanderson Should we introduce something like sys.policy_hub_ip exists like we did for port number?

jimis:
We just had an extensive discussion with @olehermanse: ideally these changes should not be needed.  `sys.policy_hub` will be IP or hostname (depending on what you bootstrapped do), but will not contain `:port` suffix. ACLs should work with the `hostname/16` notation. This way we make sure that other policy frameworks and custom policy will keep working.

So I'm closing this. If one or two changes will be needed after all, we can introduce them then.

PR #825, opened 2016-11-02
3.10.0 changelog

kacf:


kacf:
@nickanderson: Not sure why this changelog has a different format. I used the standard one from the generator, which I think we should use for all of them.

kacf:
Cherry-picked to master as well.

PR #830, opened 2016-11-15
CFE-2466: adapt tests for new behaviour

jimis:
Changelog: Behaviour change: when used with CFEngine 3.10.0 or greater,
bundles set_config_values() and set_line_based() are appending a
trailing space when inserting a configuration option with empty value.


jimis:
Ugh, the whitespace detector is failing the Travis build. :-@

nickanderson:
I just ran this on my system with the 3.10.0b1.

I do not experience the behaviour originally reported in https://tracker.mender.io/browse/CFE-2466.

I do not think the trailing space is significant in this case, so I support the alteration of the acceptance test.

PR #836, opened 2016-11-23
CFE-2360: Make apt_get package module version aware

nickanderson:
Options for behaviour we expect have changed in newer versions of
apt-get. This makes it version aware so that the correct options are
used by default.

Changelog: Title

### DIFF
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -4,6 +4,7 @@ import sys
 import os
 import subprocess
 import re
+from distutils.version import LooseVersion
 
 
 dpkg_cmd = os.environ.get('CFENGINE_TEST_DPKG_CMD', "/usr/bin/dpkg")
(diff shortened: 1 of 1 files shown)


nickanderson:
This would superspeed #812

nickanderson:
triggered PR build 98

nickanderson:
@kacf bootstrap-enterprise-pr/98/ passed a monthish ago.

Kicking off new PR bootstrap-enterprise-pr/134

nickanderson:
I couldnt find the build so testing-enterprise-pr/122

nickanderson:
@jimis or @kacf I am having issues with the PR builds not coming up green. But all the tests seem to pass?

kacf:
@nickanderson: The unit test for apt_get is not passing, probably because arguments to apt-get were changed. You should try running `test_package_module_apt_get` locally and check the return code / output.

kacf:
Try #855 instead.

PR #837, opened 2016-11-24
Fixed ACLs for hostnames in cf-serverd (CFE-2495)

olehermanse:
Using the IP in policy is the fastest way to get this working. The only drawback is it will not change the ACL before policy is re-evaluated. This can in some cases (but not all) cause connections to be refused after the hostname policy server has changed IP. 

Note that if the hostname policy server is within the same ACL, for example 192.168.x.x, it should still work fine.

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -92,7 +92,7 @@ bundle common def
                    # ".*$(def.domain)",
 
                    # Assume /16 LAN clients to start with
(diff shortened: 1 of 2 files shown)


olehermanse:
@jimis What do you think?

jimis:
+1 Merging.

PR #840, opened 2016-12-02
Add class to enable post transfer verrification during policy updates

nickanderson:
Changelog: Title


nickanderson:
@jimis FYI I merged this and picked it back to 3.10.x. I would prefer the ability to easily enable this. This has the potential to complicate some environemnts especially mixed enterprise and community installs so thats why i am changing it last minute. We need to be sure to redact "        - Change: Verify transfered files during policy update" from the changelog, as with this class guard, it's outdated.

PR #846, opened 2016-12-15
Fix apt_get package module incorrectly using interactive mode.

kacf:
Changelog: Title

### DIFF
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -6,6 +6,8 @@ import subprocess
 import re
 
 
+dpkg_options = ["--force-confold", "--force-confdef"]
+
 dpkg_cmd = os.environ.get('CFENGINE_TEST_DPKG_CMD', "/usr/bin/dpkg")
 dpkg_deb_cmd = os.environ.get('CFENGINE_TEST_DPKG_DEB_CMD', "/usr/bin/dpkg-deb")
 dpkg_query_cmd = os.environ.get('CFENGINE_TEST_DPKG_QUERY_CMD', "/usr/bin/dpkg-query")
@@ -306,7 +308,7 @@ def remove():
     return ret
 
 def file_install():
-    cmd_line = [dpkg_cmd, "-i"]
+    cmd_line = [dpkg_cmd] + dpkg_options + ["-i"]
 
     found = False
     for line in sys.stdin:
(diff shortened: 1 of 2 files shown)


jimis:
Jenkins disagrees:

```
00:26:49.459 FAIL: test_package_module_apt_get
00:26:49.459 =================================
00:26:49.459 
00:26:49.459 Fail: Does not appear in mock log: "dpkg -i /path/to/pkg"
00:26:49.459 Actual output:
00:26:49.459 "dpkg --force-confold --force-confdef -i /path/to/pkg
00:26:49.459 "
00:26:49.459 Traceback (most recent call last):
00:26:49.459   File "./test_package_module_apt_get", line 94, in <module>
00:26:49.459     assert check("file-install", ["File=/path/to/pkg"], 0, [], 1, ["dpkg -i /path/to/pkg"])
00:26:49.459 AssertionError
```

zofrex:
I reported this bug on IRC this morning - here's the error log I posted:

```
Configuration file '/etc/inspircd/inspircd.conf'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** inspircd.conf (Y/I/N/O/D/Z) [default=N] ? dpkg: error processing package inspircd (--install):
 EOF on stdin at conffile prompt
Errors were encountered while processing:
 inspircd
    info: Some error occurred while communicating with package module while installing package.
   error: Error installing package '/root/inspircd_2.0.23-1_amd64.deb'
```

What the log file is showing is a package failing to install. The context leading up to the failure is that the configuration file was created first, then the package was installed, using the "apt_get" package module but installing from a .deb file, not using apt-get and a repo. The cause of the failure is dpkg spawning an interactive prompt asking what to do with the configuration file, which the package module (obviously) doesn't respond to correctly.

After reporting this on IRC this PR was made and I was asked to test it.

I just did some manual testing on the same setup that produced the above, and can confirm this fix works - it now keeps the existing config file and installs the package, which is (imo) the expected behaviour here.

kacf:
Thanks for testing @zofrex!

Pull request updated to fix the unit test failure, and started new build in Jenkins.

jimis:
Jenkins is green now, merging. Thanks!

jimis:
Cherry-picked in all maintained branches.

PR #847, opened 2016-12-15
Fixes CFE-2445: Make apt_get module compatible with Ubuntu 16.04

amousset:
This commit allows using the python3 executable when python
executable is not there, like in the default with Ubuntu 16.04
install. It also fixes some incompatibilities of the code with Python 3.

### DIFF
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -1,10 +1,23 @@
-#!/usr/bin/python
+#!/bin/sh
+"true" '''\'
+PYTHON_BINS="/usr/bin/python /usr/bin/python3"
+
+for python_bin in $PYTHON_BINS;
+do
+  if [ -f "$python_bin" ]; then
+    exec "$python_bin" "$0" "$@"
+  fi
+done
+
+exit 127
+'''
 
 import sys
 import os
 import subprocess
 import re
 
+PY3 = sys.version_info > (3,)
 
 dpkg_cmd = os.environ.get('CFENGINE_TEST_DPKG_CMD', "/usr/bin/dpkg")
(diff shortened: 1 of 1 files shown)


ageorgop:
LGTM as well.  Thanks for picking this up.

modules/packages/apt_get
```diff
@@ -1,10 +1,23 @@
-#!/usr/bin/python
+#!/bin/sh
+"true" '''\'
+PYTHON_BINS="/usr/bin/python /usr/bin/python3"
+
+for python_bin in $PYTHON_BINS;
+do
+  if [ -f "$python_bin" ]; then
+    exec "$python_bin" "$0" "$@"
```
kacf: Lol! Yes, that's clever! +1 from me. :-)

But stick to four space indentation, like the rest of the file!

modules/packages/apt_get
```diff
@@ -1,10 +1,23 @@
-#!/usr/bin/python
+#!/bin/sh
+"true" '''\'
+PYTHON_BINS="/usr/bin/python /usr/bin/python3"
```
kacf: In the interest of being future oriented, I would reverse the order of these, so we prefer python3 over python ("python" is always python2, AFAIK).

jimis:
I didn't know about this one, so I posted a similar solution in [CFE-2445](https://tracker.mender.io/browse/CFE-2445). I think it's better because it's searching current PATH for the binaries, it doesn't need them hardcoded.

I have tested it in custom scripts though, not in the CFEngine package module. I'll post a pull request.

jimis:
@amousset can you split out the part that makes the code python3 compatible? That part can be merged immediately.

jimis:
I'm merging, and will amend modifications for the nitpicks. Thanks!

PR #848, opened 2016-12-20
Fix issue with upgrading from 3.7 to 3.10 on RHEL.

pasinskim:
The client package can not be updated as legacy package promise
is not comparing packages versions correctly.

Signed-off-by: Marcin Pasinski <marcin.pasinski@cfengine.com>


nickanderson:
Thanks!

Looks good to me, lets wait for travis to agree.

PR #850, opened 2016-12-29
CFE-2550 Change: Suppress undesired errors

nickanderson:


inventory/any.cf
```diff
@@ -787,8 +787,13 @@ bundle agent inventory_cmdb_load(file)
       "$(ckeys)" expression => "any", scope => "namespace";
 
   vars:
-      "cmdb" data => readjson($(file), "999999");
-      "cmdb_string" string => format("%S", cmdb);
+      "cmdb"
+        data => readjson($(file), "999999"),
```
tzz: Does the number need to be quoted?

tzz:
YES for me

inventory/any.cf
nickanderson: Probably not, but it was already quoted so i left it.

nickanderson:
Syntax error .... double parenthesis slipped in, pushed fix

PR #851, opened 2016-12-30
Adjust Travis to test both LTS releases.

jimis:


jimis:
Travis likes it so I'm merging.

PR #854, opened 2017-01-05
Small fixes to the previous commit.

jimis:
1. First try python3, then python
2. Search for the python binaries in PATH


jimis:
@amousset @ageorgop @nickanderson to be honest I didn't check this on a system lacking python2 (I tested with a custom binary name though), so please verify if the package_module still works for you.

nickanderson:
@jimis it works fine in my testing on ubuntu 16

PR #855, opened 2017-01-09
CFE-2360: Make apt_get package module version aware

kacf:
Options for behaviour we expect have changed in newer versions of
apt-get. This makes it version aware so that the correct options are
used by default.

Changelog: Title

Modified-by: Kristian Amlie <kristian.amlie@cfengine.com>
* Adapted unit test to accept new "apt-get -v" calls.

### DIFF
--- a/modules/packages/apt_get
+++ b/modules/packages/apt_get
@@ -16,6 +16,7 @@ import sys
 import os
 import subprocess
 import re
+from distutils.version import LooseVersion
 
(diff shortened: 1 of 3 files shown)


kacf:
Currently running in Jenkins.

kacf:
All green!

kacf:
@nickanderson: I'll let you handle the cherry-picking and the task (which can probably be closed).

PR #858, opened 2017-01-27
Add: missing pkg and nimclient package modules

nickanderson:
@BasementTrix we forgot to add these to the Makefile so that they would
be shipped with the package.


nickanderson:
Picked to 3.10.x

PR #859, opened 2017-01-28
Fix python3 incompatibility.

jimis:
Popen() returns a "bytes" object in python3 which fails to
split(). Passing the universal_newlines parameter makes it return a
string.


nickanderson:
Niiice

jimis:
[Green build](https://ci.cfengine.com/job/testing-enterprise-pr/145/), merging.

PR #864, opened 2017-02-09
CFE-2582: Add a shortcut templates to MPF framework

basvandervlies:
See also ticket: https://tracker.mender.io/projects/CFE/issues/CFE-2582

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -196,9 +196,9 @@ bundle common def
       comment => "Define plugins path",
       handle => "common_def_vars_dir_plugins";
 
-      "dir_templates"
-        string => ifelse( isvariable( "mpf_dir_templates"),
-                          $(mpf_dir_templates),
+      "server[dir_templates]"
+        string => ifelse( isvariable( "def.server[dir_templates]"),
+                          $(def.server[dir_templates]), 
                           "$(sys.workdir)/templates"),
       comment => "Define templates path",
       handle => "common_def_vars_dir_templates";
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -196,9 +196,9 @@ bundle common def
       comment => "Define plugins path",
       handle => "common_def_vars_dir_plugins";
 
-      "server[dir_templates]"
-        string => ifelse( isvariable( "def.server[dir_templates]"),
-                          $(def.server[dir_templates]), 
+      "server_controls[dir_templates]"
+        string => ifelse( isvariable( "def.server_controls[dir_templates]"),
+                          $(def.server_controls[dir_templates]), 
                           "$(sys.workdir)/templates"),
       comment => "Define templates path",
       handle => "common_def_vars_dir_templates";
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -198,7 +198,7 @@ bundle common def
 
       "server_controls[dir_templates]"
         string => ifelse( isvariable( "def.server_controls[dir_templates]"),
-                          $(def.server_controls[dir_templates]), 
+                          $(def.server_controls[dir_templates]),
                           "$(sys.workdir)/templates"),
       comment => "Define templates path",
       handle => "common_def_vars_dir_templates";
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -196,6 +196,14 @@ bundle common def
       comment => "Define plugins path",
       handle => "common_def_vars_dir_plugins";
 
+      "dir_templates"
+        string => ifelse( isvariable( "mpf_dir_templates"),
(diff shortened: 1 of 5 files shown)


controls/def.cf
```diff
@@ -196,6 +196,14 @@ bundle common def
       comment => "Define plugins path",
       handle => "common_def_vars_dir_plugins";
 
+      "dir_templates"
```
nickanderson: I think this should be the same as what the override is. mpf_dir_templates

nickanderson: I think this should be the same as what the override is. mpf_dir_templates

nickanderson:
also just fyi the current fails in travis we are still investigating the cause of.

controls/def.cf
tzz: I don't understand why only this variable needs the `mpf_` prefix. Is it used anywhere else?

tests/acceptance/00_basics/def.json/server_dirs.cf
```diff
@@ -0,0 +1,55 @@
...
+{
+      inputs => { "../../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent test
+{
```
nickanderson: Lets skip this test on 3.7.x (oldlts) by tagging it:

```
meta:
  "test_skip_unsupported" string => "cfengine_3_7";
```

nickanderson:
Hey look at that, its green.

tests/acceptance/00_basics/def.json/server_dirs.cf.json
```diff
@@ -0,0 +1,7 @@
+{
+    "vars": {
+        "server_controls": {
```
nickanderson: @basvandervlies @tzz lets bikeshed this name a bit.

0) I feel like I am terrible at picking names 
1) I don't consider this a "server control" because access rules can be in any server bundle.
2) Why nest this deeply instead of `mpf_dir_templates`, `mpf_templates` or `templates`?
3) Why do I suggest MPF anything? Because highlighting the fact that its a MPFism and not a core thing seems like a good thing to do.

controls/def.cf
```diff
@@ -196,6 +196,14 @@ bundle common def
       comment => "Define plugins path",
       handle => "common_def_vars_dir_plugins";
 
+      "server_controls[dir_templates]"
+        string => ifelse( isvariable( "def.server_controls[dir_templates]"),
+                          $(def.server_controls[dir_templates]),
+                          "$(sys.workdir)/templates"),
```
nickanderson: Since the MPF ships with a template directory in the root do you think that would be a more appropriate default?

tests/acceptance/00_basics/def.json/server_dirs.cf.json
tzz: Works for me!

nickanderson: This is why I have some aversion to the deep nesting. I am looking at how this would work if we get [CFE-2084](https://tracker.mender.io/browse/CFE-2084) implemented.

This example takes two augments data sets and merges them together using `mergedata()` to merge vars together. Note how `vars.server_controls.option_a` defined in `augments_0` is missing from the merged result. Please correct me if I am wrong, but I believe this is the current behavior implemented in [PR 2515](https://github.com/cfengine/core/pull/2515).

```
    bundle agent main
    {
      vars:
        # def.json in the root of inputs
        "augments_0" data => '{ "vars": { "server_controls": { "option_a": "default", "dir_templates" : "/var/cfengine/my_templates" } } }';    
    
        # a supplementary augments file as proposed in CFE-2084
        "augments_1" data => '{ "vars": { "server_controls": { "dir_templates" : "/var/cfengine/my_templates" } } }';    
    
        "merged" data => mergedata( augments_0, augments_1 );
        "s_merged" string => format("%S", merged);
    
      reports:
    
      # Because of how mergedata works simply merging vars picked up from different
      # augments files would cause the last object to win. This is the main reason I
      # was trying to avoid deep nesting in the augments file.
    
        "CFEngine $(sys.cf_version)";
        "$(s_merged)";
    
    }
```

Here is the result:

```
    R: CFEngine 3.11.0a.334d1fc
    R: {"vars":{"server_controls":{"dir_templates":"/var/cfengine/my_templates"}}}
```

nickanderson:
To be superseded by https://github.com/cfengine/masterfiles/pull/891

nickanderson:
Merged in  #891

PR #873, opened 2017-02-21
CFE-2587: Make stock policy update more resiliant

nickanderson:
Changelog: Title

This change ensures that a promise failure while updating inputs does
not leave the remote agent with the impression that
cf_promises_validated is reflective of the current state.


nickanderson:
Picked to 3.10.x in 898d924

PR #874, opened 2017-02-21
ENT-3047: Do not symlink agents to /usr/local/bin on coreos

nickanderson:
Changelog: Title

CoreOS has a read only filesystem there.


nickanderson:
ping @estenberg

estenberg:
I think this is the best we can do, thanks.

estenberg:
Picked to 3.10.x in 74e79dfc85d3e2244b5361a1c1371c82f14158a8

PR #875, opened 2017-02-22
Fix missing body

nickanderson:
Manually backport missing body.


jimis:
+1 thanks @nickanderson !

PR #877, opened 2017-03-03
CFE-2608: Add prunetree bundle to stdlib

nickanderson:
The prunetree bundle allws you to delete files and directories up to a
sepcified depth older than a specified number of days.

Changelog: Commit

### DIFF
--- a/lib/bundles.cf
+++ b/lib/bundles.cf
@@ -301,6 +301,39 @@ bundle agent prunedir(dir, max_days)
       depth_search  => recurse("1");
 }
 
+bundle agent prunetree(dir, depth, max_days)
+# @brief Delete files and directories inside "dir" up to "depth" older than "max_days".
+# @depends delete tidy
+# @depends depth_search recurse_with_base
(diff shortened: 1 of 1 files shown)


nickanderson:
@jimis

lib/bundles.cf
```diff
@@ -301,6 +301,39 @@ bundle agent prunedir(dir, max_days)
...
+# }
+# ```
+{
+  files:
+      "$(dir)"
+      comment => "Delete files and directories under $(dir) up to $(depth)
+                    depth older than $(max_days)",
+
+      delete        => tidy,
+      file_select   => days_old( $(max_days) ),
```
jimis:
```
lib/files.cf-# @param days Number of days
lib/files.cf-{
lib/files.cf-      mtime       => irange(0,ago(0,0,"$(days)",0,0,0));
lib/files.cf-      file_result => "mtime";
lib/files.cf-}
```

Why do I read this as selecting whatever was modified up to these many days old? So the opposite of what we want.

PR #881, opened 2017-03-24
ENT-3082: Stop trying to symlink binaries into sbin on windows

nickanderson:


nickanderson:
@olehermanse see if this fixes your promises not kept.

olehermanse:
Seems to have fixed the issue. LGTM @nickanderson

PR #889, opened 2017-04-11
CFE-2605: Available upgrade parsing is incorrect with some packages with apt_get module

amousset:


kacf:
I suspect that the regex can be shortened even further. I'll take a look at this tomorrow.

kacf:
Superseded by #899. @amousset: Can you verify that this works for you?

amousset:
@kacf It works for me 👍

kacf:
Thanks for testing, I have merged #899.

PR #891, opened 2017-04-12
CFE-2582: Add templates shortcut

nickanderson:


nickanderson:
@basvandervlies I went back to your original dir_templates proposal and I added some light documentation.

PR #898, opened 2017-04-20
Switch to POSIX compliant shell tool paths on Solaris 9 and 10.

kacf:
CFE-2616

Changelog: Title

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -386,6 +386,9 @@ bundle common def
                     the disk may fill up or cause additional collection
                     issues.";
 
+      # Enable paths to POSIX tools instead of native tools when possible.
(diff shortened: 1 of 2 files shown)


kacf:
Don't merge this yet, I'm still working out the paths.

kacf:
@nickanderson: I have updated the PR, perhaps you can take a look?

@mikeweilgart: Can you try that this works to your expectations?

kacf:
Merging this so we can get it into the beta. Nick didn't think this was a bug fix so this will be in master / 3.11 only.

mikeweilgart:
Great, I like the simplified logic (compared to handling Solaris 11 specially).  Thank you!

PR #899, opened 2017-04-20
CFE-2605: apt_get: Fix broken parsing of updates with extra sources.

kacf:
Changelog: apt_get package module: Fix bug which prevented updates
from being picked up if there was more than one source listed in the
'apt upgrade' output, without a comma in between.

Based on a patch by Alexis Mousset <contact@amousset.me>.


kacf:
Cherry-picked to 3.7.x and 3.10.x as well.

PR #912, opened 2017-05-01
ENT-3143: Use package_module for self upgrade with rpm or deb

nickanderson:


nickanderson:
Need to take a different approach to self upgrade because of systemd. Closing this.

PR #914, opened 2017-05-09
ENT-3184: Use sparse in tar backup during CFE upgrades as lmdb uses spare files

estenberg:
Proven to improve performance, at least in one case (large lmdb files).
https://cfengine.zendesk.com/agent/tickets/3409

### DIFF
--- a/cfe_internal/update/update_bins.cf
+++ b/cfe_internal/update/update_bins.cf
@@ -408,7 +408,7 @@ bundle edit_line u_backup_script
       "#!/bin/sh
 
 if [ $(const.dollar)1 = \"BACKUP\" ]; then
- tar cfz $(const.dollar)2 $(sys.workdir) > /dev/null
+ tar cfzS $(const.dollar)2 $(sys.workdir) > /dev/null
 fi
 if [ $(const.dollar)1 = \"RESTORE\" ]; then
  tar xfz $(const.dollar)2


estenberg:
The file may be read twice (on older Linux systems), and concerned if all tar utilities on Linux supports sparse (S).
So we should run this through CI before merging.

estenberg:
GNU tar supports spase (https://www.gnu.org/software/tar/manual/tar.html#IDX526), however Solaris tar does not seem to (http://ibgwww.colorado.edu/~lessem/psyc5112/usail/man/solaris/tar.1.html).

nickanderson:
@Lex-2008 test failing in travis because of fakeroot?

Lex-2008:
yes, @jimis researched something in this area. I suggest `sudo apt-get install fakeroot`

jimis:
Yes, see my relevant commit in Core: cfengine/core:9786610673b28cdd29734a1925a727a34b4c23ce

nickanderson:
OK closeing and opeing to retrigger travis

nickanderson:
@jimis  https://github.com/cfengine/masterfiles/pull/915 ?

cfe_internal/update/update_bins.cf
```diff
@@ -408,7 +408,7 @@ bundle edit_line u_backup_script
       "#!/bin/sh
 
 if [ $(const.dollar)1 = \"BACKUP\" ]; then
- tar cfz $(const.dollar)2 $(sys.workdir) > /dev/null
+ tar cfzS $(const.dollar)2 $(sys.workdir) > /dev/null
```
jimis: Is the `S` flag working in all our platforms? Is it defined in POSIX?

estenberg: See the comment above; I don't know but I think it is a GNU extension.

nickanderson:
Passed travis. Merging.

PR #915, opened 2017-05-15
Use fakeroot with travis

nickanderson:


jimis:
https://github.com/cfengine/masterfiles/pull/916

PR #917, opened 2017-05-16
CFE-2533: Add zypper package module

nickanderson:
Changelog: Title


jimis:
Thank you Normation for this ( @jooooooon @ncharles etc).

@nickanderson the blargh at the top is kind of confusing. Is it GPLv3 or MIT licensed? We should also update it for year 2017.

PR #919, opened 2017-05-16
ENT-3155: Refactor self upgrade

nickanderson:
Also removes symlinking into /usr/local/sbin (CFE-2636)

### DIFF
--- a/cfe_internal/update/update_bins.cf
+++ b/cfe_internal/update/update_bins.cf
@@ -14,63 +14,7 @@
 #
 ################################################################################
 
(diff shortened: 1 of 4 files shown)


nickanderson:
built fine in Jenkins

Findings:
1) Need to add promise to run agent after standalone self upgrade policy. This
   should cut the time for data to reach mission portal by one scheduled
   interval.
2) Need to autoconf `cfe_internal/update/standalone_self_upgrade.cf` for hard
   coded version.

nickanderson:
Make it so.

PR #922, opened 2017-05-20
Add /var/log/postgresql.log to list of enterprise_log_files

atsaloli:
As reported in https://cfengine.zendesk.com/hc/en-us/requests/3436,
CFEngine does not rotate the Postgres log file which will eventually
cause the disk to go 100% full and then stuff breaks (e.g., can't
download CSV reports anymore).

Nick Anderson suggested adding postgresql.log to enterprise_log_files,
which this commit does.  I've already implemented this change and tested it
on one of our hubs as documented in support ticket 3436.


nickanderson:
Thanks for bringing it to our attention @atsaloli  I merged something nearly identical.

PR #925, opened 2017-05-23
Zypper module fixes

jimis:


jimis:
Lets wait a bit before merging.

jimis:
The downgrade operation does not work on very old zyppers, and it's not straightforward to warn about it, since the package module does not know whether it performs an upgrade or downgrade. Currently it just fails silently.

PR #928, opened 2017-06-13
CFE-2326: Add bundle to track component variables to restart when necessary

nickanderson:
Changelog: Title


nickanderson:
Ah fails on 3.7.x because the function variablesmatching_as_data() doesnt exist. Will have to re-factor to make it 3.7.x compatible.

PR #930, opened 2017-06-13
Do not use bindtointerface on windows (CFE-2662)

olehermanse:
Note that this does not fix the issue, it's just a workaround on windows.

### DIFF
--- a/controls/cf_serverd.cf
+++ b/controls/cf_serverd.cf
@@ -32,9 +32,6 @@ body server control
       # Suggested value >= total number of hosts
       maxconnections        => "$(def.control_server_maxconnections)";
 
-      # Bind to all interfaces including ipv6
-      bindtointerface => "::";
-
       # Allow connections from nodes which are out-of-sync
       denybadclocks         => "false";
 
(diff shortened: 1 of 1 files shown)


olehermanse:
@nickanderson This should be merged and cherry picked to 3.11.x

controls/cf_serverd.cf
```diff
@@ -32,9 +32,6 @@ body server control
       # Suggested value >= total number of hosts
       maxconnections        => "$(def.control_server_maxconnections)";
 
-      # Bind to all interfaces including ipv6
```
nickanderson: If you remove this definition from the implicit any:: context then the agent will only bind to ipv4 interfaces, so this one should remain, and then be overridden for windows below.

olehermanse: @nickanderson Did you see the whole context? This is inside `!windows` context. So should apply just like before on everything except windows.

olehermanse: Also, this is `body server control` so why do you mention agent?

jimis: @olehermanse I think @nickanderson 's suggestion will have the same effect with what you did in this patch.

olehermanse: @jimis @nickanderson I have tested with and without this patch on linux now. Even tested ipv6. Cannot see any difference, works fine with and without this patch.

PR #932, opened 2017-06-16
Fix Masterfiles travis failures

olehermanse:
@jimis Maybe this fixes it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
     - env: STYLE_CHECK_JOB=1
 
     # JOB 1
-    - env: CF_VERSION=latest
+    - env: CF_VERSION=master
       addons:
         apt:
           sources:
@@ -20,7 +20,7 @@ matrix:
             - cfengine-community
 
     # JOB 2
-    - env: CF_VERSION=lts
+    - env: CF_VERSION=3.10.x
       addons:
         apt:
           sources:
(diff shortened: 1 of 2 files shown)


jimis:
This change looks good, but it's not enough. When `testall`  invokes `testall --bindir` then cf-net binary fails is not found and that one fails. This happens because we install packages of 3.10 and 3.7 here that don't have cf-net at all.

Another change is needed, maybe in core cf-net binary could be optional.

olehermanse:
@jimis cf-net doesn't exist in 3.10. So I think core is correct as is? cf-net exists in master and 3.11, not 3.10 or 3.7

olehermanse:
@nickanderson @jimis So the problem here is that we're testing 3.10 package with 3.11 tests / testall. My suggestion is to comment out the latest/master test case for now, and re-enable it when 3.11 includes cf-net. We can run into the same error if we add new features with tests, but we also kind of want to test master with latest, even though they can be quite different.

nickanderson:
Well at least I find the actual versions easier to read than lts/oldlts.

We do need want to make sure that 3.11 MPF works on 3.10, and 3.7 binaries (because of upgrade)

jimis:
This is a good compromise. We should remember to uncomment the top session when 3.11 is out, in both 3.11 and master branches.

nickanderson:
This should be picked into 3.11 3.10  and 3.7 branches?

olehermanse:
@nickanderson The cf-net commit should only be in 3.11.x and master. The others you can try to cherry pick back.

PR #936, opened 2017-07-18
ENT-3283: copy test_package_module_apt_get file from master 

Lex-2008:
I completely reworked the commit - now it's just plain file copy, no fancy cherry-picking or merge conflict resolution

### DIFF
--- a/tests/unit/test_package_module_apt_get
+++ b/tests/unit/test_package_module_apt_get
@@ -87,12 +87,16 @@ def check(operation, input, no_lines, expected_output, mock_lines, mock_output):
 
     return True
 
-apt_get_options = "-o Dpkg::Options::=--force-confold -o Dpkg::Options::=--force-confdef -y --force-yes"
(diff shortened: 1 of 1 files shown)


jimis:
Is this a cherry-pick? If so use `-x` switch when cherry-picking. The contents are for @nickanderson to review.

Lex-2008:
git help says that "-x" works only for cherry picks without conflicts, so i've added relevant info manually

nickanderson:
The content looks really different from what is in master. Since nothing in the protocol has changed, to me it makes sense that the modules in maintained branches match what is in master.

Lex-2008:
hm, then let's copy just tests/unit/test_package_module_apt_get file and see if it helps!

nickanderson:
That's what I would probably do. I can't think of any current reason why we wouldnt want the package modules to be the same across all the maintained branches.

Lex-2008:
ok, the PR updated - now it's just file copy :)

nickanderson:
It passed travis testing, so merging.

PR #941, opened 2017-08-03
Update changelog from 3.11.0b1.

jimis:


jimis:
Takk @nickanderson, fixed and merging.

PR #947, opened 2017-08-17
ENT-3313: Ensure that Mission Portal can export pdf and csv reports

nickanderson:
Changelog: Title


nickanderson:
@jimis  @Lex-2008

Lex-2008:
I'm not expert of masterfiles language, but I can correlate the fix described by Nick in the bug to the changes in the policy. So to speed up the release, let's integrate it and test!

PR #950, opened 2017-08-22
CFE-2685: Ignore empty options in apt_get module

amousset:


nickanderson:
Thanks for working on this!
Change does not seem to suppress the error, comments in ticket.

nickanderson:
Thank you!

PR #951, opened 2017-08-26
CFE-2690: Be less noisy when a promised service is not found

nickanderson:
The failed promise messages are likely to make it clear enough what is going on
without this extra report. It is nice to have in inform_mode and verbose_mode.

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -371,7 +371,7 @@ bundle agent standard_services(service,state)
     verbose_mode.fallback::
       "$(this.bundle): falling back to classic_services to $(state) $(service)";
 
-    systemd.service_notfound::
+    systemd.service_notfound.(inform_mode|verbose_mode)::
         "$(this.bundle): Could not find service: $(service)";
 }
 


mikeweilgart:
@nickanderson, is that message really necessary even in inform mode?

As we have many services promises that are simply to promise "stop," i.e. to ensure the services are NOT running, we don't care when these services aren't found.  If they're not found, that means they won't be running, which is what's wanted after all.

The noise is too noisy for "stop" promises, probably even too noisy for verbose mode.  On the other hand, if the service is promised to be "started," it should probably give some complaint even in "normal" (not verbose and not inform) mode.

What would it take to make these messages quieter and more in line with the UNIX tradition?

> [Rule of Silence](http://catb.org/~esr/writings/taoup/html/ch01s06.html#id2878450): When a program has nothing surprising to say, it should say nothing.

nickanderson:
Probably not in inform mode, it's not actually modifying the system but let's keep it in verbose.

Feel free to open a pull request.

PR #955, opened 2017-09-12
ENT-3368: Fix self upgrade for hosts older than 3.7.4

nickanderson:
This patch avoids the use of the ifelse function on hosts running
versions older than 3.7.4 so that the correct variables are defined.
Additionally this patch updates the url patterns for enterprise packages
so that enterprise hubs can automatically download the packages and
clients reference the newer naming convention.

Changelog: Title


nickanderson:
Superceded by #962

PR #960, opened 2017-09-13
add server control allowlegacyconnects to def.json

estenberg:
Not too experience with this, so would appreciate your review @nickanderson 
We should backport to 3.10 if ok.

Ref https://cfengine.zendesk.com/agent/tickets/3589


nickanderson:
superceded by https://github.com/cfengine/masterfiles/pull/960

PR #964, opened 2017-10-03
ENT-3390: Backport Fix to frequent logging of enterprise license utilization

nickanderson:


nickanderson:
travis is having provisioning issues. I am confident this is compatible with 3.7.x, so I am merging without waiting any longer for travis to complete its tests.

PR #972, opened 2017-10-14
CFE-2497: Make mpf more tuneable from augments

nickanderson:


nickanderson:
hum i fubard the commit history

PR #982, opened 2017-10-25
fixed an error in the file_make_mustache_*, must be data instead of state

basvandervlies:


nickanderson:
Thank you! Superceeded by https://github.com/cfengine/masterfiles/pull/983 (added changelog entry, and issue id to commit)

PR #994, opened 2017-11-09
Allow multiple sections in insert_ini_section

mvpel:
When insert_ini_section is used on a file which contains other sections, or isoperating on a new file with no sections and is given an slist of multiple sections and values to add, the values were not associated with the correct file sections, but intermingled with others. The [section] for multiple sections is placed at the top of the file, but then the values are placed anywhere else.

Example:

```
bundle agent ini_edit
{
  vars:
    "options[test][option_one]" string => "content_one";
    "options[fake][fake_option]" string => "example";
    "sections" slist => getindices("options");

  files:
    "/tmp/foo.ini"
    create => "true",
    edit_line => insert_ini_section("$(sections)", "ini_edit.options[$(sections)]");
}
```

Result in foo.ini:

> [fake]
> [test]
> option_one=option_one_content
> option_one=fake_one
> option_two=option_two_content

By tying the location of the config values to the section header, insert_ini_section is now able to be iterated.

PR #995, opened 2017-11-09
CFE-2721: Add test for insert_ini_section and mutliple sections

nickanderson:


nickanderson:
Adding soft fail test to go with https://github.com/cfengine/masterfiles/pull/994

PR #998, opened 2017-11-09
CFE-2721: Allow multiple sections in insert_ini_section

nickanderson:


nickanderson:
Thanks @mvpel ! merging

PR #1004, opened 2017-11-13
CFE-2722: Fix systemctl path detection

amousset:
f704e39e broke systemctl path detection.

Changelog: Title


nickanderson:
Thank you !

PR #1008, opened 2017-12-01
WIP ENT-3451: upgrade to php 7.2.0

craigcomstock:


cfe_internal/enterprise/templates/httpd.conf.mustache
```diff
@@ -223,9 +223,9 @@ LogLevel warn
 </IfModule>
 
 
-LoadModule php5_module       modules/libphp5.so
-AddHandler php5-script       .php
-AddType    application/x-httpd-php-source php5
+LoadModule php7_module       modules/libphp7.so
```
nickanderson: We need to make this conditional so that it only renders the php7 one if php7 is present, and will render php5 if thats present. Probably we should do that inside policy to set some vars, then we can just reference the vars that cfengine derives from here.

nickanderson:
@craigcomstock  I think this PR can be superseded by https://github.com/cfengine/masterfiles/pull/1010

PR #1009, opened 2017-12-12
Explicitly set perms for augments bundle test

olehermanse:
@nickanderson This was causing the test to fail on my VM, dunno why.


nickanderson:
I don't know, but its OK to specify it imho, the expectation was that files copied without specifying preserve true would all be created with 600 perms.

PR #1013, opened 2017-12-18
ENT-3251: Add external watchdog support for stuck cf-execd

nickanderson:
Changelog: Title

### DIFF
--- a/templates/cfengine_watchdog.mustache
+++ b/templates/cfengine_watchdog.mustache
@@ -1,3 +1,7 @@
 # This file is managed by CFEngine
 # If cf-execd is executable and if no process matching cf-execd can be found then restart cf-execd
 * * * * * root [ -x {{{vars.sys.cf_execd}}} ] && if ! {{{vars.paths.pgrep}}} cf-execd > /dev/null; then {{{vars.sys.cf_execd}}}; fi
+
+# If there is a cf-execd process running and promise_summary.log has not been
+# updated in the last 30 minutes then kill and re-start cf-execd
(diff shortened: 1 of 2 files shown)


templates/cfengine_watchdog.mustache
```diff
@@ -1,3 +1,7 @@
 # This file is managed by CFEngine
 # If cf-execd is executable and if no process matching cf-execd can be found then restart cf-execd
 * * * * * root [ -x {{{vars.sys.cf_execd}}} ] && if ! {{{vars.paths.pgrep}}} cf-execd > /dev/null; then {{{vars.sys.cf_execd}}}; fi
+
+# If there is a cf-execd process running and promise_summary.log has not been
+# updated in the last 30 minutes then kill and re-start cf-execd
+* * * * * root if find {{{ vars.sys.workdir}}}/promise_summary.log -mmin +30; then if [ $(pgrep -c cf-execd) -gt 0 ]; then echo "cf-execd does not seem to be triggering agent runs, restarting"; pkill cf-execd; {{{sys.cf_execd}}} fi; fi
```
jimis: `pgrep` and `pkill` are not portable. I'm also not sure about the `find` options, you'll have to try. Also 30 minutes is a bit low, some people even have a 30-min schedule for regular runs.

Here is a script that should be portable:

```
get_cfexecd_pids()
{
    ps axo pid,comm | grep '[c]f-execd' | sed -e 's/^ *\([0-9][0-9]*\) .*/\1/'`
}

if whatever_logic_is_true
then
    for p in `get_cfexecd_pids`
    do
        kill $p  &&  sleep 2
        if get_cfexecd_pids | grep $p >/dev/null    # not dead yet
        then
            kill -9 $p
        fi
    done
fi
````

nickanderson: It's mostly targeted at linux as it is now. Mostly trying to keep the cron jobs to one line entries.

PR #1015, opened 2017-12-20
Fix old packages matching tests

nickanderson:
Thes old tests are for package inventory using the package_method
implementation. The bundles excercisded were removed from the framework
because they were no longer necessry. They were simply moved in to the
tests themselves in order to preserve the functionality.


nickanderson:
@jimis I added this PR yesterday because @aleksandrychev showed me some test failures. Curiously the MPF tests in travis didn't catch these.

PR #1022, opened 2018-01-05
Tidy up unused bundles and missing inline docs

nickanderson:


nickanderson:
Failure in 3.7.x because variablesmatching_as_data isn't available.

Wrapping in if minimum_version(3.10) should solve it.

PR #1027, opened 2018-01-15
Sort indicies for insert_ini_section()

vpodzime:
The first commit makes the tests more deterministic, the second commit makes the produced INI files potentially better readable.

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -515,7 +515,9 @@ bundle edit_line insert_ini_section(name, config)
 # @param config The fully-qualified name of an associative array containing `v[LHS]="rhs"`
 {
   vars:
-      "k" slist => getindices($(config));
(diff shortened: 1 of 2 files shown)


vpodzime:
See https://github.com/cfengine/core/pull/2915 for some details.

vpodzime:
> For that you'll need to use an intermediary variable as the nested function calling didn't work back then.

I did that, but it's still failing with 3.7. Any idea how to make this compatible with all the versions, @nickanderson?

nickanderson:
The current changes look 3.7.x compatible to me. Lets see what travis says.

nickanderson:
Now can you squash and rebase?

vpodzime:
> Now can you squash and rebase?

Sure, initially, I wasn't sure if the change in behaviour in ``insert_ini_section()`` would be okay or not.

PR #1028, opened 2018-01-16
ENT-3572: Maintain access to exported CSV reports in older versions

nickanderson:


mikeweilgart:
I see the [change that caused this break](https://github.com/nickanderson/masterfiles/blob/4f0cddf989297532ab224ba59b89db683bd7d4dc/cfe_internal/enterprise/CFE_knowledge.cf#L126-L130) was made in April 2016.  The comment on the relevant promise reads:

> The ssl certificates only need to be readable by root since the webserver reads them before dropping privileges.

Presumably that only applies to newer CFEngine versions, not older ones?  Hence the issue observed with 0 byte CSVs (on e.g. 3.7.3).

Since your patch only includes 3.7.3, 3.7.4, 3.7.5 and 3.10.0, I had a look at the changelog - I don't see anything about certificates or webserver changes in the 3.10.1 or the 3.7.3 changelog entries.

Are you expecting that 3.7.0, 3.7.1, 3.7.2, 3.10.1, 3.10.2 and the eventual 3.10.3 and 3.7.6 releases are not/will not be affected?  I would be surprised if that were the case.

On the face of it it seems that a more sensible class expression might be:

```
      "ENT_3572" -> { "ENT-3572" }
        comment => "Hosts with this class need to be sure that the ssl directory
                    is readable and executable by other users",
        or => {
                "enterprise_3_7", "enterprise_3_10"
              };
```

But if there *is* some behavior change in the webserver between 3.10.0 and 3.10.1 (and between 3.7.5 and the upcoming 3.7.6, and between 3.7.2 and 3.7.3), then the class expression in the diff is perfect.  Can you please comment?

nickanderson:
In the listed versions testing shows that removing read and execute bit from the SSL directory causes 0byte csvs. In older versions the change is not specifically known to cause 0 byte csvs. In newer versions removing the read and execute bit for other users does not create a condition producing 0 byte csvs.

PR #1030, opened 2018-01-16
CFE-2759 : prevent yum from locking when possible

ncharles:
https://tracker.mender.io/browse/CFE-2759

### DIFF
--- a/lib/packages.cf
+++ b/lib/packages.cf
@@ -250,6 +250,22 @@ bundle common rpm_knowledge
       "rpm3_arch_regex" string => "\S+\s+(\S+).*";
 }
 
+bundle common redhat_no_locking_knowledge {
+  vars:
+    redhat_setopt_available::
+      "no_locking_option" string => "--setopt=exit_on_lock=True";
+    !redhat_setopt_available::
+      "no_locking_option" string => "";
+
+
+  classes:
+    # Use the class 'agent' as a guard, to prevent the command to be executed during cf-promises/cf-serverd
+    # Class agent is a hard class set by cf-agent
+    redhat.agent::
+      "redhat_setopt_available" not => returnszero("LANG=C ${paths.path[yum]} --setopt=exit_on_lock=True 2>&1 | ${paths.path[grep]} -q -E '(Command line error: no such option: --setopt|Main config did not have a exit_on_lock attr. before setopt)'", "useshell");
+
+}
+
 bundle common redhat_knowledge
 # @depends paths
 # @brief common Red Hat knowledge bundle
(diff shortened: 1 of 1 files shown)


ncharles:
for the record, we've been using this option for several years in Rudder

lib/packages.cf
```diff
@@ -250,6 +250,22 @@ bundle common rpm_knowledge
...
+    redhat_setopt_available::
+      "no_locking_option" string => "--setopt=exit_on_lock=True";
+    !redhat_setopt_available::
+      "no_locking_option" string => "";
+
+
+  classes:
+    # Use the class 'agent' as a guard, to prevent the command to be executed during cf-promises/cf-serverd
+    # Class agent is a hard class set by cf-agent
+    redhat.agent::
```
nickanderson: Why did I not know about this `agent` class?

Of course, you can't see it from `cf-promises --show-classes` because that's not the agent.
But you can see it with the new `--show-evaluated-classes` option to `cf-agent`. 

```
> $ sudo cf-agent --show-evaluated-classes=^agent
Class name                                                   Meta tags                               
agent                                                        cfe_internal,source=agent,hardclass   
```


Thanks for the lesson!

lib/packages.cf
```diff
@@ -264,9 +280,9 @@ bundle common redhat_knowledge
       "call_yum" string => "$(paths.path[yum])";
       "call_rpmvercmp" string => "$(sys.bindir)/rpmvercmp";
 
-      # on RHEL 4, Yum doesn't know how to be --quiet
-      "yum_options" string => ifelse("centos_4|redhat_4", "",
-                                     "--quiet");
+      # on RHEL 3/4, Yum doesn't know how to be --quiet
+      "yum_options" string => ifelse("centos_4|redhat_4|centos_3|redhat_3", "",
+                                     "--quiet ${redhat_no_locking_knowledge.no_locking_option}");
```
nickanderson: What about `redhat_5? In the research I have done the option doesn't seem to apply on el5.

lib/packages.cf
ncharles: I discovered this option last week, when I detected that cf-promises ran the yum command.
I searched and digged, and found this class - it is very useful

lib/packages.cf
ncharles: We failed to find if the option was there on CentOS5 or not - so we assumed it *could* be there and stayed on the safe side (we've seen people recompiling their yum in the wild)
As we are checking if option is there or not, impact is quite minimal.
What do you think ?

ncharles:
I've amended the ticket commit
We have had any issues with the package module yet, but that may be sheer luck. We surely need to backport this option there as well - are the classes available in the python module ?

lib/packages.cf
nickanderson: OK thas all I was looking for. I suspected people might be re-building. Makes sense to me. Ill add note to ticket as well.

nickanderson:
Great, looks good.

Regarding the package_module I believe that locking is handled differently. For package_module based promises I believe there is a global lock to prevent concurrent executions of the same package module.

However, if we ever find a similar issue then yes, we should look to implement similar in the module itself.

ncharles:
my bad for the dash
I added "Changelog: Title" at the end of the commit message - not sure if that's what was expected

nickanderson:
Perfect.

Thanks!

ncharles:
Thank you !

PR #1040, opened 2018-02-01
ENT-3602: execute runalerts.php as cfapache, to be identical to systemd_supervised systems.

jimis:


jimis:
@nickanderson please review and merge. Which host should we test it in? Which hub platform is non-systemd supervised?

nickanderson:
Centos 6 doesn't use systemd. Ubuntu 14 I think, Unsure about debian 7

jimis:
Thanks @nickanderson, done.

nickanderson:
Ok are we waiting for a build, or can we merge this now?

PR #1041, opened 2018-02-01
ENT-3602: execute runalerts.php as cfapache, to be identical to systemd_supervised systems.

jimis:


jimis:
Updated with the change that Nick requested in #1040.

PR #1049, opened 2018-02-16
CFE-2776 Fix cf-runagent during 3.7.x -> 3.10.x migration

nickanderson:
This fixes an issue introduce by the addition of the ability to specify the bundlesequence cf-agent should use during a request made by cf-runagent.

3.10.x and greater needs an open cfruncommand specification so that it can append the arguments like no locks and remote bundles.

### DIFF
--- a/controls/cf_serverd.cf
+++ b/controls/cf_serverd.cf
@@ -45,12 +45,28 @@ body server control
       cfruncommand => "$(sys.cf_agent) -I -D cfruncommand -f \"$(sys.update_policy_path)\"  &
                        $(sys.cf_agent) -I -D cfruncommand";
 
-    !windows::
-      # last single quote in cfruncommand is left open, so that
-      # arguments (like -K and --remote-bundles) are properly appended.
-      cfruncommand => "$(def.cf_runagent_shell) -c \'
+    !windows.!cfengine_3_7::
+
+      # In 3.10 the quotation is properly closed when EOF is reached. It is left
+      # open so that arguments (like -K and --remote-bundles) can be appended.
+      # 3.10.x does not automatically append -I -Dcfruncommand
+
+        cfruncommand => "$(def.cf_runagent_shell) -c \'
                            $(sys.cf_agent) -I -D cfruncommand -f $(sys.update_policy_path)  ;
                            $(sys.cf_agent) -I -D cfruncommand";
+
+    !windows.cfengine_3_7::
+
+      # 3.7.x does not properly close open quotes when EOF is reached preventing
+      # policy execution, so we must provide a trailing quote for cfruncommand.
+
+      # 3.7.x Automatically appends -I -Dcfruncommand to the cfruncommand. This
+      # results in a benign error about not being allowed to specify inputs for
+      # remote execution.
+
+        cfruncommand => "$(def.cf_runagent_shell) -c \'
+                           $(sys.cf_agent) -I -D cfruncommand -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I -D cfruncommand\'";
 
       # Bind to all interfaces including ipv6
       # Adding this on windows will force ipv6 only
--- a/controls/cf_serverd.cf
+++ b/controls/cf_serverd.cf
@@ -42,8 +42,8 @@ body server control
       allowusers            => { "root" };
 
     windows::
-      cfruncommand => "$(sys.cf_agent) -I -D cfruncommand -f \"$(sys.update_policy_path)\"  &
-                       $(sys.cf_agent) -I -D cfruncommand";
+      cfruncommand => "$(sys.cf_agent) -I -D cf_runagent_initiated -f \"$(sys.update_policy_path)\"  &
+                       $(sys.cf_agent) -I -D cf_runagent_initiated";
 
     !windows.!cfengine_3_7::
 
@@ -52,8 +52,8 @@ body server control
       # 3.10.x does not automatically append -I -Dcfruncommand
 
         cfruncommand => "$(def.cf_runagent_shell) -c \'
-                           $(sys.cf_agent) -I -D cfruncommand -f $(sys.update_policy_path)  ;
-                           $(sys.cf_agent) -I -D cfruncommand";
+                           $(sys.cf_agent) -I -D cf_runagent_initiated -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I -D cf_runagent_initiated";
 
     !windows.cfengine_3_7::
 
@@ -61,12 +61,14 @@ body server control
       # policy execution, so we must provide a trailing quote for cfruncommand.
 
       # 3.7.x Automatically appends -I -Dcfruncommand to the cfruncommand. This
-      # results in a benign error about not being allowed to specify inputs for
-      # remote execution.
+      # results in an error about not being allowed to specify inputs for remote
+      # execution from the update policy. In order to preserve the ability for
+      # policy to know its execution provenance we define a different class
+      # `cf_runagent_initiated` that does not elicit the error.
 
         cfruncommand => "$(def.cf_runagent_shell) -c \'
-                           $(sys.cf_agent) -I -D cfruncommand -f $(sys.update_policy_path)  ;
-                           $(sys.cf_agent) -I -D cfruncommand\'";
+                           $(sys.cf_agent) -I -D cf_runagent_initiated -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I -D cf_runagent_initiated\'";
 
       # Bind to all interfaces including ipv6
       # Adding this on windows will force ipv6 only
--- a/MPF.md
+++ b/MPF.md
@@ -32,6 +32,11 @@ request to expose the tunable into the ```def``` bundle.
 **Note:** `controls/def.cf` contains the defaults and settings for `promises.cf`
 and `controls/update_def.cf` contains the defaults and settings for `update.cf`.
 
+Since 3.7.8, 3.10.4, and 3.12.0 The class `cf_runagent_initated` is defined by
+default in the MPF for agent executions initiated by `cf-runagent` through
+`cf-serverd`. Previously the class `cfruncommand` was defined. See `body server
+control cfruncommand` in `controls/cf_serverd.cf`.
+
 ## Update Policy (update.cf)
 
 Synchronizing clients with the policy server happens here, in


atsaloli:
:+1:

controls/cf_serverd.cf
```diff
@@ -45,12 +45,24 @@ body server control
...
-                           $(sys.cf_agent) -I -D cfruncommand";
+    !windows.!cfengine_3_7::
+
+      # in cfengine 3.10.0 and later last single quote in cfruncommand is left
+      # open, so that arguments (like -K and --remote-bundles) are properly
+      # appended.
+
+        cfruncommand => "$(def.cf_runagent_shell) -c \'
+                           $(sys.cf_agent) -I -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I ";
```
jimis: why did you remove `-Dcfruncommand`? If you want the class defined in 3.10, you have to append it manually. In older versions it used to be the case that cf-serverd was appending it before spawning cf-agent, when commanded via a remote cf-runagent.

jimis: Here is what changed, copied from Core's changelog:

```
        - cf-serverd no longer appends "-I -Dcfruncommand" to
                     cfruncommand, this has to be done manually in masterfiles
                     body server control. (Redmine #7732)
```

jimis: in other words, with CFEngine 3.10.x, if you want the class defined in cf-agent, when triggered via a remote cf-runagent, you have to manually append that in the policy.

If you remove it, it will result to the `cfruncommand` class being undefined when triggered by remote cf-runagent. Just this, nothing will blow up.

craigcomstock: Hmm... I worked on this a bit and mostly payed attention to getting the cfruncommand to work at all, not on the effect on the remote cf-agent's environment. Some more testing is in order?

jimis: Why? Didn't it work before with 3.7.x ? I recall there were several fixes to cfruncommand parsing that were cherry-picked to 3.7.x.

craigcomstock: I'm not sure. Looks like Nick or I (or both) will have to dig back into this a bit.

nickanderson: Good catch @jimis, I re-added the cfruncommand definition.

controls/cf_serverd.cf
```diff
@@ -45,12 +45,25 @@ body server control
...
+      # appended.
+
+        cfruncommand => "$(def.cf_runagent_shell) -c \'
+                           $(sys.cf_agent) -I --define cfruncommand -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I --define cfruncommand";
+
+    !windows.cfengine_3_7::
+
+      # 3.7 doesn't support specifying the remote bundlesequence and does not
+      # know to add a trailing single quote so it must be accounted for here.
```
jimis: No, that's not the difference. 3.10 does not auto-append a closing trailing quote either. But when EOF is reached, the shell closes all quotation implicitly.

controls/cf_serverd.cf
```diff
@@ -45,12 +45,25 @@ body server control
...
-                           $(sys.cf_agent) -I -D cfruncommand";
+    !windows.!cfengine_3_7::
+
+      # in cfengine 3.10.0 and later last single quote in cfruncommand is left
+      # open, so that arguments (like -K and --remote-bundles) are properly
+      # appended.
+
+        cfruncommand => "$(def.cf_runagent_shell) -c \'
+                           $(sys.cf_agent) -I --define cfruncommand -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I --define cfruncommand";
```
jimis: So you only changed the style from `-D` to `--define` ? I don't like the new style :-p

controls/cf_serverd.cf
```diff
@@ -45,12 +45,25 @@ body server control
...
+
+    !windows.cfengine_3_7::
+
+      # 3.7 doesn't support specifying the remote bundlesequence and does not
+      # know to add a trailing single quote so it must be accounted for here.
+      # 3.7.x Automatically appends -I -Dcfruncommand to the cfruncommand
```
jimis: That is true. So now that you are manually closing the quote, are you sure that it works?

controls/cf_serverd.cf
jimis: I can see in the serverd code that `-D` is treated specially to check if "role activation" is permitted. I can't see `--define` there, so I would suggest leaving it `-D`.

nickanderson: Bah, fine. Yeah, I much prefer the long options. It's way easier to read than knowing all the abbreviations.

controls/cf_serverd.cf
nickanderson: @craigcomstock  and @atsaloli  are sure.

atsaloli:
Is the `cfruncommand` class needed by CFEngine?  Why are we defining the `cfruncommand` class?  Aha, found it in Volker's post of 2014 at https://groups.google.com/forum/#!topic/help-cfengine/8vpnVJ_x1JU : 

--start quote--
This is to prevent remote DoS attacks.

I made a few fixes in 3.6 that enforced that behaviour in cf-agent:

https://github.com/cfengine/core/pull/1708

The previous code tried to check if either -K, -f or --no-lock was passed to the server by cf-runagent via -o, but failed to correctly replicate getopt's parsing behaviour, so it was possible to execute a remote agent ignoring locks by passing "-o"--no-loc". The new code passes a class "cfruncommand" to cf-agent, and cf-agent looks for that class when processing options, and if set, refuses to accept no-lock or file-input. It's a bug fix to existing behaviour, rather than a change of behaviour. However, the new logic makes it impossible to specify -K or -f options in the cfruncommand (and I admit, the ChangeFile doesn't point this out, it only mentions the new cfruncommand class).

--end quote--

atsaloli:
I am confused, the comments say 3.7.x automatically adds -D cfruncommand, so then we are you adding -D  cfruncommand in masterfiles?

```
      # 3.7.x Automatically appends -I -Dcfruncommand to the cfruncommand
      # DTV Customization to add DEBUG class to the update run

        cfruncommand => "$(def.cf_runagent_shell) -c \'
                           $(sys.cf_agent) -I -D cfruncommand,DEBUG -f $(sys.update_policy_path)  ;
                           $(sys.cf_agent) \'";
```

atsaloli:
The latest version from this pull request does not work!

```
[redacted]> cf-serverd executing cfruncommand: /bin/sh -c '
                           "/var/cfengine/bin/cf-agent" -I -D cfruncommand,DEBUG -f /var/cfengine/inputs/update.cf  ;
                           "/var/cfengine/bin/cf-agent" ' -I -Dcfruncommand
[redacted]> ->    error: Specifying input files is not allowed for remote execution
```

controls/cf_serverd.cf
nickanderson: I spun up a 3.10.3 hub with a 3.7.3, 3.7.7 and 3.10.3 remote agents running the
stock 3.10.3 MPF.

From the hub I did a runagent against the 3.7.3 agent.

     verbose: Server is TRUSTED, received key 'SHA=a97346372843a7dee5f58980a9c88bca36538022b48c2e618590a04b0c07a76a' MATCHES stored one.
    192.168.33.3> cf-serverd executing cfruncommand: /bin/sh -c '
                               "/var/cfengine/bin/cf-agent" -I -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                               "/var/cfengine/bin/cf-agent" -I -Dcfruncommand
    192.168.33.3> -> sh: -c: line 0: unexpected EOF while looking for matching `''
    192.168.33.3> -> sh: -c: line 3: syntax error: unexpected end of file
     verbose: Connection to 192.168.33.3 is closed

From the hub I did a runagent against the 3.7.7 agent.

    192.168.33.3> cf-serverd executing cfruncommand: /bin/sh -c '
                              "/var/cfengine/bin/cf-agent" -I -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                              "/var/cfengine/bin/cf-agent" -I -Dcfruncommand
    192.168.33.3> -> sh: -c: line 0: unexpected EOF while looking for matching `''
    192.168.33.3> -> sh: -c: line 3: syntax error: unexpected end of file
    verbose: Connection to 192.168.33.3 is closed

From the hub I did a runagent against a 3.10.3 agent.

    192.168.33.3> cf-serverd executing cfruncommand: /bin/sh -c '
                              "/var/cfengine/bin/cf-agent" -I -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                              "/var/cfengine/bin/cf-agent" -I -D cfruncommand
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/passwd' (SHA=8580c7f630bb7cb7252e9e1f7e129eb9c591f4c4132261ca76ebfb66f8f6c59c)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/group' (SHA=665c4e56c7ff3e5fcd19a937ff40208d16c8fd9371a7e779feb3e33d8de0719d)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/services' (SHA=f0c4af9261e7e6193db0bbb8f335c0f0c094d8d7aa428e5d8cbb674f2e13b2ef)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/shadow' (SHA=8be6b00d56db641797111598a203dd80b6a285802c14c3155d4e1bceca7f57b0)
    verbose: Connection to 192.168.33.3 is closed
    You have new mail in /var/spool/mail/root

I applied the patch where on 3.7 the quote is terminated explicitly and because
`-I -Dcfruncommand` is supposed to be appended by default I modified
services/main.cf to report when `cfruncommand` was defined.

    reports:
      cfruncommand::
        "Executed by cf-runagent (I see that cfruncommand is a defined class)";

     verbose: Server is TRUSTED, received key 'SHA=a97346372843a7dee5f58980a9c88bca36538022b48c2e618590a04b0c07a76a' MATCHES stored one.
    192.168.33.3> cf-serverd executing cfruncommand: /bin/sh -c '
                               "/var/cfengine/bin/cf-agent" -I -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                               "/var/cfengine/bin/cf-agent" ' -I -Dcfruncommand
    192.168.33.3> ->    error: Specifying input files is not allowed for remote execution
     verbose: Connection to 192.168.33.3 is closed

In the above output we can see that indeed `-I -Dcfruncommand` is appended to
`cfruncommand` after the closing quote. Unfortunately since it comes AFTER the
closing quote, it isn't considered an option to the command. Instead the agent
errors and tells you that it's not possible to specify input files for remote
execution.

So, I adjusted `cfruncommand` to define `-I -Dcfruncommand` inside of the
closing quote.

    !windows.cfengine_3_7::
    
      # 3.7 doesn't support specifying the remote bundlesequence and does not
      # know to add a trailing single quote so it must be accounted for here.
      # 3.7.x Automatically appends -I -Dcfruncommand to the cfruncommand
    
        cfruncommand => "$(def.cf_runagent_shell) -c \'
                           $(sys.cf_agent) -I -D cfruncommand -f $(sys.update_policy_path)  ;
                           $(sys.cf_agent) -I -Dcfruncommand\'";

I got this result when executing `cf-runagent` from the hub.

    192.168.33.3> cf-serverd executing cfruncommand: /bin/sh -c '
                               "/var/cfengine/bin/cf-agent" -I -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                               "/var/cfengine/bin/cf-agent" -I -Dcfruncommand' -I -Dcfruncommand
    192.168.33.3> ->    error: Specifying input files is not allowed for remote execution
    192.168.33.3> -> R: Executed by cf-runagent (I see that cfruncommand is a defined class)

We can see that the agent still complained about not allowing remote execution
to specify input files, but importantly we see that the policy was actually
executed and that `cfruncommand` was defined as desired.

The auto appending of `-I -Dcfruncommand` in 3.7.x seems to be useless since
3.7.x is not able to close the quote implicitly like we can in 3.10.x

     verbose: Server is TRUSTED, received key 'SHA=5934f85e77fa1636365761bacf4c7827bec87f547cc0c95ac81292e45ccbe341' MATCHES stored one.
    192.168.33.3> cf-serverd executing cfruncommand: /bin/sh -c '
                               "/var/cfengine/bin/cf-agent" -I -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                               "/var/cfengine/bin/cf-agent" -I -Dcfruncommand' -I -Dcfruncommand
    192.168.33.3> ->    error: Specifying input files is not allowed for remote execution
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/passwd' (SHA=8580c7f630bb7cb7252e9e1f7e129eb9c591f4c4132261ca76ebfb66f8f6c59c)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/group' (SHA=665c4e56c7ff3e5fcd19a937ff40208d16c8fd9371a7e779feb3e33d8de0719d)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/services' (SHA=f0c4af9261e7e6193db0bbb8f335c0f0c094d8d7aa428e5d8cbb674f2e13b2ef)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/shadow' (SHA=8be6b00d56db641797111598a203dd80b6a285802c14c3155d4e1bceca7f57b0)
    192.168.33.3> -> R: Executed by cf-runagent (I see that cfruncommand is a defined class)

And we see that it works fine in 3.10.3 -> 3.10.3

     verbose: Server is TRUSTED, received key 'SHA=a23be6f85cf3660b4c5a82b3d465a8fcb3ccb8906c88d4a2ac30285021d8d591' MATCHES stored one.
    192.168.33.3> cf-serverd executing cfruncommand: /bin/sh -c '
                               "/var/cfengine/bin/cf-agent" -I -D cfruncommand -f /var/cfengine/inputs/update.cf  ;
                               "/var/cfengine/bin/cf-agent" -I -D cfruncommand
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/passwd' (SHA=8580c7f630bb7cb7252e9e1f7e129eb9c591f4c4132261ca76ebfb66f8f6c59c)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/group' (SHA=665c4e56c7ff3e5fcd19a937ff40208d16c8fd9371a7e779feb3e33d8de0719d)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/services' (SHA=f0c4af9261e7e6193db0bbb8f335c0f0c094d8d7aa428e5d8cbb674f2e13b2ef)
    192.168.33.3> ->   notice: Storing sha256 hash for '/etc/shadow' (SHA=8be6b00d56db641797111598a203dd80b6a285802c14c3155d4e1bceca7f57b0)
    192.168.33.3> -> R: Executed by cf-runagent (I see that cfruncommand is a defined class)
     verbose: Connection to 192.168.33.3 is closed

atsaloli:
If you define the `cfruncommand` class _in the masterfiles_, you get  `error: Specifying input files is not allowed for remote execution`.

mikeweilgart:
I would like to point out that the underlying complexity that is causing all of this additional complexity, is the original decision to make the cfruncommand attribute accept a _command to be executed._  In actual fact, this is too much abstraction, as cf-runagent is neither designed nor expected to operate at all correctly if anything is specified in the cfruncommand attribute besides a cf-agent command.

A cleaner design from the start would have been, in order to allow options or classes to be defined in policy, to have an *slist* attribute which could contain flags to be added on to the cf-agent executions—but to hardcode the cf-agent binary (with promises.cf and update.cf hardcoded as well) into cf-serverd as the commands to be run.

The architecture of using a string containing `sh -c` and _another_ embedded string containing both cf-agent commands, and then leaving the single quoted string unterminated to allow the appending of further attributes or arguments *inside the binary,* is just extremely unclean.  Icky.

Or, alternatively and probably better, don't allow the cfruncommand to be globally customized _at all_ in policy.  YAGNI and it just adds complexity and a maintenance burden—witness this pull request.

Also see CFE-2664, in particular [this comment](https://tracker.mender.io/browse/CFE-2664?focusedCommentId=80730&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-80730), which explains another bug/problem caused by using a shell "in between" rather than hardcoding the execution of cf-agent in the daemon binaries.

In summary: *I contend that it is neither useful nor desirable to allow policy-based customization of the commands to be triggered by cf-runagent.*

Customization of the *arguments* to be passed to the cf-agent instances is more useful, but should be handled by cf-runagent—not by policy writers.  That's the purpose of cf-runagent, after all.

If cf-runagent had flags to allow arguments (_not_ arbitrary arguments as in the old `-o`) to be passed to the update.cf run, and to the promises.cf run, separately—you wouldn't need to fiddle with quoting.

And the "magic" behavior of the cfruncommand _class_ (currently hardcoded in the cf-agent binary) could be done away with, also, because `cf-serverd` would have its own sanity checking built in (as would cf-runagent) and would only pass arguments with _known meaning_ (whitelist) to the cf-agent command.

A win for simplicity all around.

nickanderson:
Pull requests are always welcome.

atsaloli:
```
$ cf-agent -D cfruncommand -f /tmp/no/such/file.cf
   error: Specifying input files is not allowed for remote execution
$
```

mikeweilgart:
Aleksey just found that the brokenness was caused by having cfruncommand (the class, not the attribute :P) defined for the update run in cfruncommand (the attribute, not the class).

Quick demonstration of the magic behavior of the cfruncommand class:

```
# cat temp.cf
bundle agent main {
  reports:
    apple::
      "Apple is set";
    orange::
      "Orange is set";
}
# cf-agent -V
CFEngine Core 3.7.3
CFEngine Enterprise 3.7.3
# cf-agent -Kf ./temp.cf
# cf-agent -Kf ./temp.cf -D apple
R: Apple is set
# cf-agent -Kf ./temp.cf -D apple -D cfruncommand
   error: Remote execution cannot ignore locks
# cf-agent -f ./temp.cf -D apple -D cfruncommand
   error: Specifying input files is not allowed for remote execution
#
```

nickanderson:
Yes, that's weird. Please always try to reproduce with the most recent binaries.

```
> $ cf-agent -Kf ./test.cf -D talk -D cfruncommand                                           
R: Hello world
                                                                                              
nickanderson@nickanderson-ThinkPad-W550s /tmp                                      [15:44:24] 
> $ cf-agent -V                                                                              
CFEngine Core 3.11.0
CFEngine Enterprise 3.11.0
                                                                                              
nickanderson@nickanderson-ThinkPad-W550s /tmp                                      [15:44:42] 
> $ cat test.cf                                                                              
bundle agent main
{
  reports:

  talk:: "Hello world";
}
                                                                                              
nickanderson@nickanderson-ThinkPad-W550s /tmp                                      [15:44:45] 
> $ cf-agent -V                                                                              
CFEngine Core 3.11.0
CFEngine Enterprise 3.11.0
```

nickanderson:
@atsaloli  you were testing while I was in the middle of pushing changes and testing. Feel free to test again. This should work across 3.7 and 3.10 according to my tests.

controls/cf_serverd.cf
```diff
@@ -45,13 +45,29 @@ body server control
...
                            $(sys.cf_agent) -I -D cfruncommand";
 
+    !windows.cfengine_3_7::
+
+      # 3.7.x does not properly close open quotes when EOF is reached preventing
+      # policy execution, so we must provide a trailing quote for cfruncommand.
+
+      # 3.7.x Automatically appends -I -Dcfruncommand to the cfruncommand. This
+      # results in a benign error about not being allowed to specify inputs for
+      # remote execution.
```
mikeweilgart: I don't understand why you say it's a benign error.  This won't run the update.cf file at all on CFEngine 3.7.3.

mikeweilgart:
Nick, this really doesn't make sense to me.  As you demonstrated, the special behavior that `cf-agent` manifests when the `-D cfruncommand` flag is set *in 3.7* was *removed* in the 3.10 cf-agent binaries.  Given that, **what is the purpose of retaining the `-D cfruncommand` flag in the `cfruncommand =>` attribute for 3.10?**

I understand that the 3.7 version of `cf-serverd` automatically appends `-I -D cfruncommand` to the value of the `cfruncommand` attribute before executing it, and that 3.10 `cf-serverd` does not do this.  However, why does that matter?  The 3.10 version of `cf-agent` doesn't do anything special with that class *anyway.*  So why does it need to be defined in policy?

Also, what is the purpose of defining `-D cfruncommand` for the `update.cf` run?  That was never the behavior before this pull request, under any circumstances.  In 3.7 that wouldn't work at all, because in `cf-agent` 3.7 you can't combine `-D cfruncommand` and `-f update.cf`.  And in 3.10 it doesn't do anything, because `cf-agent` 3.10 doesn't have any special behavior for `-D cfruncommand` anyway.  So why have it there?

Why have `-D cfruncommand` *anywhere*?  At the moment, out of the four places where you have it:

- update.cf run on 3.10
- promises.cf run on 3.10
- update.cf run on 3.7
- promises.cf run on 3.7

The first two are no-op, since 3.10 binaries don't do anything special with `-D cfruncommand`.

The third one breaks the update run on 3.7, because you can't use `-D cfruncommand` in combination with `-f`.

And the fourth one is unnecessary, because the 3.7 cf-serverd binary will append `-D cfruncommand` to the `cfruncommand => ` attribute anyway, before executing.

Unless you want to handle the bizarre corner case of a host running 3.10 cf-serverd and 3.7 cf-agent.  But for goodness' sake, that doesn't make very much sense.  And even then, all it will do is prevent someone *who has root access to your CFEngine Hub* from triggering the `-K` option on a remote agent run.  Which doesn't seem like a necessary feature in the first place.  (Oh, and they would have to use a 3.7 cf-runagent binary on the hub, also, unless I'm much mistaken, because 3.10 cf-runagent won't even pass the `-K` to the remote cf-serverd in the first place, will it?)

Let's just get rid of `-D cfruncommand` entirely.  There never was a time or a version when it was necessary to have it specified in masterfiles.

mikeweilgart:
(Or does the `cfruncommand` class still have some other special (and undocumented) behavior in the 3.10 cf-agent binary?  From [your comment earlier](https://github.com/cfengine/masterfiles/pull/1049#issuecomment-368147537) I am under the impression that it doesn't.)

controls/cf_serverd.cf
nickanderson: I thought the error was related to the automatic appending of  `-I -Dcfruncommand`. But it was indeed from the update policy activation.

nickanderson:
I didn't read your whole post.

I changed from cfruncommand to cf_runagent_initated

That avoids the landmines related to the literal cfruncommand string, and it preserves the ability for the policy (both update.cf and promises.cf) to know they were initated from cf-runagent as opposed to their normal polices.

controls/cf_serverd.cf
mikeweilgart: Great!  I would just recommend removing the inaccurate comment, otherwise it looks good to me.  I'm not convinced that the `cf_runagent_initiated` class is *necessary*, but it's not harmful.  :)

atsaloli:
Thanks @nickanderson 

Please do read Mike's post, as it expresses some design concerns, specifically with regard to complexity. I understand the real world is not simple but since complexity is now the limitation to scalability, not CPU power or RAM or network bandwidth, needless complexity is killer.

I agree that removing `-D cfruncommand` (or replacing it with `-D anythingelse`) ought to render the update policy runnable via cf-runagent on 3.7.

Thanks!! Really appreciate your folks attention and help on this issue.

controls/cf_serverd.cf
atsaloli: That's right.  Thanks, Nick.

atsaloli:
@nickanderson I tested the following with 3.7.3 host and 3.10.3 host and cf-runagent works on both -- we prefer to have DEBUG enabled during update runs (it helps us to have visibility, we have policy channels, remember?) -- plus I omitted `cfruncommand` class entirely as 3.7 cf-serverd supplies that, and the magic around it was removed in 3.10 -- and we don't have any policy that behaves differently when run via cf-runagent:

```
    !windows.!cfengine_3_7::
      # last single quote in cfruncommand is left open, so that
      # arguments (like -K and --remote-bundles) are properly appended.
      cfruncommand => "$(def.cf_runagent_shell) -c \'
                           $(sys.cf_agent) -I -D DEBUG -f $(sys.update_policy_path)  ;
                           $(sys.cf_agent) -I ";

    !windows.cfengine_3_7::
      # 3.7 doesn't know to add a trailing single quote so add it here.
      cfruncommand => "$(def.cf_runagent_shell) -c \'
                           $(sys.cf_agent) -I -D DEBUG -f $(sys.update_policy_path)  ;
                           $(sys.cf_agent) -I \'";

```

atsaloli:
@nickanderson I also tested with the contents of this merge request, and found it works against 3.7 and 3.10 hosts:

```
Hub [root@dc1247fe-c618-4da4-83c5-90b31ac56c05 masterfiles]# cf-runagent -H 3_10_host  -t 99999
165.225.128.183> cf-serverd executing cfruncommand: /bin/sh -c '
                           "/var/cfengine/bin/cf-agent" -I -D cf_runagent_initiated -f /var/cfengine/inputs/update.cf  ;
                           "/var/cfengine/bin/cf-agent" -I -D cf_runagent_initiated
165.225.128.183> ->     info: Edit file '/tmp/cfe_internal_update_policy.txt'
165.225.128.183> -> R: Hello world from cfe_internal_update_policy bundle at Fri Feb 23 23:45:02 2018
165.225.128.183> ->     info: Edit file '/etc/motd'
Hub [root@dc1247fe-c618-4da4-83c5-90b31ac56c05 masterfiles]# cf-runagent -H 3_7_host  -t 99999
165.225.128.110> cf-serverd executing cfruncommand: /bin/sh -c '
                           "/var/cfengine/bin/cf-agent" -I -D cf_runagent_initiated -f /var/cfengine/inputs/update.cf  ;
                           "/var/cfengine/bin/cf-agent" -I -D cf_runagent_initiated' -I -Dcfruncommand
165.225.128.110> ->     info: Edit file '/tmp/cfe_internal_update_policy.txt'
165.225.128.110> -> R: Hello world from cfe_internal_update_policy bundle at Fri Feb 23 23:45:34 2018
165.225.128.110> ->     info: Edit file '/etc/motd'
Hub [root@dc1247fe-c618-4da4-83c5-90b31ac56c05 masterfiles]#
```

As you can see, each host runs the update policy (which prints a report and edits a file) and the regular policy (which updates /etc/motd).

controls/cf_serverd.cf
```diff
@@ -42,15 +42,31 @@ body server control
...
-                           $(sys.cf_agent) -I -D cfruncommand";
+    !windows.!cfengine_3_7::
+
+      # In 3.10 the quotation is properly closed when EOF is reached. It is left
+      # open so that arguments (like -K and --remote-bundles) can be appended.
+      # 3.10.x does not automatically append -I -Dcfruncommand
+
+        cfruncommand => "$(def.cf_runagent_shell) -c \'
+                           $(sys.cf_agent) -I -D cf_runagent_initiated -f $(sys.update_policy_path)  ;
+                           $(sys.cf_agent) -I -D cf_runagent_initiated";
```
jimis: Instead of introducing a new class name because of 3.7.x, I would rather keep the class named `cfruncommand` and honour the legacy. :-) Specifically for `update.cf` I would completely remove `-D` argument for 3.7.x.

Changing the class name also nullifies the 3.7.x "protection" of refusing to run other policy files, so anybody will be able to append arbitrary arguments (and potentially `-f` argument).

If you change the class name after all, ensure a loud changelog message and a section in the release notes.

atsaloli:
@jimis : I am sure your clients that use the "cfruncommand" class would appreciate not having to update their code.

@nickanderson : We tested with 3.7.3 because that is a version we have to support.  I suppose an ideal test would have included 3.7.3 (for me) and 3.7.7 (for you).  :) I'll keep that in mind!

controls/cf_serverd.cf
mikeweilgart: This still doesn't make sense to me, @jimis.  The options go from `cf-runagent` (on the hub) to `cf-serverd` (on the client) and from there to `cf-agent` on the client.  If `cf-serverd` is version 3.7, it will append `-I -D cfruncommand` to the arguments it passes to `cf-agent` regardless of what's in masterfiles, so it doesn't have to be in policy.  And if `cf-serverd` is version 3.10, then presumably `cf-agent` is also 3.10 (they're on the same host) and in that case `-D cfruncommand` doesn't cause any special behavior in the agent at all, right?

So in what circumstances will omitting `-D cfruncommand` from *policy* allow arbitrary arguments to be passed via `cf-runagent` through `cf-serverd` to `cf-agent` and be acted upon?

jimis: You are right @mikeweilgart so the danger is not there as I initially thought. I withdraw my objection. :+1:

nickanderson:
Thanks for all the reviews everyone. I think I have captured the knowledge in the commit messages, and inline comments. Merging to master will backport to 3.10.x.

atsaloli:
Very good, thank you, Nick.

jimis:
> Merging to master will backport to 3.10.x.

The class name change will break existing deployments so I would suggest to keep the same name while removing `-D` from update.cf invocation in your cherry-picking.

PR #1050, opened 2018-02-16
Let amazon linux use yum package module

ageorgop:
Amazon linux doesn't work currently with new package module and nothing installs as a result.


lib/packages.cf
```diff
@@ -60,7 +60,7 @@ bundle common package_module_knowledge
     freebsd::
       "platform_default" string => "pkg";
 
-    redhat::
+    redhat|amazon_linux::
```
nickanderson: +1

ageorgop:
yeah i had put in a pr for that, not sure why it's showing up here :/.  I may abort and start over

PR #1057, opened 2018-02-28
Changelog: Add amazon_linux class to yum package module

nickanderson:


nickanderson:
Thanks @ageorgop  Merging.

PR #1058, opened 2018-03-20
ENT-3554: Tooling to upgrade MPF files (WIP)

olehermanse:
Internal ticket: tracker.mender.io/browse/ENT-3554

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ CFVERSION
 
 # vim
 .*.sw[po]
+
+# Atom-IDE
+.ropeproject
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,4 @@
+[style]
+based_on_style = pep8
+split_before_first_argument = true
+column_limit = 78
--- a/contrib/policy_upgrade/generate_hashes.py
+++ b/contrib/policy_upgrade/generate_hashes.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+import json
+import hashlib
+
+from collections import OrderedDict
+
+
+def json_string(data):
+    return json.dumps(data, indent=2)
+
+
+def json_print(data):
+    print(json_string(data))
+
+
+def json_save(data, path):
+    string = json_string(data)
+    with open(path, "w") as f:
+        f.write(string)
+
+
+def json_load(path):
+    with open(path, "r") as f:
+        data = f.read()
+        return json.loads(data, object_pairs_hook=OrderedDict)
+
+
+def file_hash(path):
+    if not os.path.isfile(path):
+        raise ValueError
+
+    hasher = hashlib.sha256()
+    block_size = 4096
+    with open(path, 'rb') as f:
+        buf = f.read(block_size)
+        while len(buf) > 0:
+            hasher.update(buf)
+            buf = f.read(block_size)
+    return hasher.hexdigest()
+
+
+def get_file_paths(root_path):
+    file_paths = []
+    for subdir, dirs, files in os.walk(root_path):
+        for f in files:
+            path = os.path.join(subdir, f)
+            if ".git/" in path:
+                continue
+            file_paths.append(path)
+    file_paths.sort()
+    return file_paths
+
+
+def generate_db(root_path, version):
+    db = OrderedDict()
+    file_paths = get_file_paths(root_path)
+
+    for path in file_paths:
+        digest = file_hash(path)
+        if digest not in db:
+            db[digest] = []
+        if path not in db:
+            db[path] = []
+        file_info = "{} {}".format(version, path)
+        db[digest].append(file_info)
+        db[path].append(file_info)
+        print("{} {}".format(digest, path))
+    return db
+
+
+def merge_databases(dbs):
+    master = OrderedDict()
+    for db in dbs:
+        for digest in db:
+            if digest not in master:
+                master[digest] = []
+            for file_info in db[digest]:
+                master[digest].append(file_info)
+    return master
+
+
+def main(args):
+    assert (len(args) > 1)
+    versions = args[1:]
+    path = "./"
+    dbs = []
+
+    for version in versions:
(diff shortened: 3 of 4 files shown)


olehermanse:
@nickanderson This is by no means done, but feel free to discuss desired functionality and command line usage here.

nickanderson:
@mikeweilgart  you might be interested in whats happening here. Understand this is very rough. But the idea is a tool to help you upgrade your policy sets by identifying which policy files have/haven't changed between versions.

mikeweilgart:
Thanks for the ping!  Yes, it's interesting.

What I would love to see is some before-the-fact documentation that could then be discussed/reviewed.  :)

What I usually do when writing a script or command line utility is to *first* write the usage message (that gets shown when `--help` is passed), and then go ahead and implement it.  It's like a lightweight spec, and it has the added advantage that it forces me to write the tool so it's usable, because I first have to make the documentation make sense, figure out what the behavior of the option flags *should* be and how to explain it, etc.  That's just *my* workflow...but usage info needs to be written anyway, so it may as well be written first.

Name - scan-mpf - ????

Synopsis - scan_mpf.py [version] ...

Description: ???

olehermanse:
@mikeweilgart Maybe don't look at the code for now, think more about use and workflow.

From a user perspective I think something like this makes sense:
```
$ python3 policy_upgrade /var/cfengine/masterfiles
Detected masterfiles version 3.10.2
masterfiles/update.cf has been modified!
masterfiles/controls/cf_execd.cf is outdated, matches version 3.10.0!
masterfiles/lib/bundles.cf has been copied/renamed to masterfiles/lib/bundl.cf!
masterfiles/lib/bundles.cf has been deleted!
```

I started focusing on what information we can find about the policy, before starting to work on making changes / upgrading. Most of what is shown above is done or close to done. The script doesn't rely on git history, and that's kind of important (version information will be pre generated and distributed with the tool).

Question now is what happens next. How do you imagine using a tool like this to do the actual upgrade part? This question applies to @nickanderson and anyone else who is interested as well. Like I said, the workflow is most important to me. As I'm not primarily a policy writer, I need input on this. I don't think it's useful to start thinking about the help menu or what the options should be.

One idea:
```
$ python3 policy_upgrade /var/cfengine/masterfiles $HOME/masterfiles_v3.10.3
Detected masterfiles version 3.10.2 - Correct? (y/n) y

[1] /var/cfengine/masterfiles/ - your custom policy, will not be edited
[2] /root/masterfiles_v3.10.3/ - upgraded policy you want to apply your changes to, will be edited(!)
Please read and understand the above info before continuing - OK? (y/n) y

Skipping 129 policy files which have not been modified
Skipping 2 policy files which only have whitespace changes

masterfiles/update.cf has been modified - What to do:
  1 Create diff / "merge conflict" for manual editing
  2 Skip (Your custom policy will not be applied, you can make manual edits later)
  3 Print diff
> 1
Please open /root/masterfiles_v3.10.3/update.cf to resolve the conflict

masterfiles/controls/cf_execd.cf is outdated, matches version 3.10.0 - What to do:
  1 Keep using outdated version - copy it to /root/masterfiles_v3.10.3/controls/cf_execd.cf
  2 Skip (Your outdated policy will not be applied, you can make manual edits later)
> 3 Print diff
[...]

masterfiles/lib/bundles.cf has been copied/renamed to masterfiles/lib/bundl.cf - What to do:
  1 Copy /var/cfengine/masterfiles/lib/bundl.cf -> /root/masterfiles_v3.10.3/lib/bundl.cf
  2 Copy /root/masterfiles_v3.10.3/lib/bundles.cf -> /root/masterfiles_v3.10.3/lib/bundl.cf
  3 Skip it, /var/cfengine/masterfiles/lib/bundl.cf is a mistake or not needed.
> 3

masterfiles/lib/bundles.cf has been deleted - What to do:
  1 Delete /root/masterfiles_v3.10.3/lib/bundles.cf
  2 Skip, keep the new version at /root/masterfiles_v3.10.3/lib/bundles.cf
> 2

Done! You can now go into /root/masterfiles_v3.10.3 and commit to an up-to-date branch there.
```

mikeweilgart:
That's just about right, regarding the workflow.

Regarding the merge conflict resolution—the ideal is to show the conflict tags using the "diff3" style (as in Git) for easier comparison and resolution.

For comparison, the workflow *I* have always used for policy upgrades, although somewhat convoluted in terms of necessary Git commands, makes the actual upgrade and conflict resolution downright easy:

For sake of example, I'll say I'm upgrading masterfiles from 3.7.3 to 3.10.3.  The first thing I do is set up the following situation in Git (I won't detail here how I accomplish this, unless someone asks):

stock-3.7.3 is a tagged version with the stock 3.7.3 masterfiles.  (You don't really need the tag but it makes the following explanation easier.)

custom-3.7.3 is one commit forward from stock-3.7.3, and contains *all* customizations (all added in a single commit).

stock-3.10.3 is also one commit forward from stock-3.7.3, and contains what you expect, all added/changed in a single commit.

Then I just merge custom-3.7.3 and stock-3.10.3.

Since they are both immediate descendants of (have the common parent) stock-3.7.3, there are only three versions to compare and no intermediate points.  Diff3 style of merge conflict display means for any file with a conflict, I can easily see what was in stock-3.7.3, what was changed from there to stock-3.10.3, and what was changed from the stock-3.7.3 to my custom-3.7.3.  (For more about diff3 style, run the command `LESS='+/HOW CONFLICTS ARE PRESENTED' man git-merge`)

The *only* thing unsatisfactory to me about this workflow is the irritation/difficulty involved in setting it up.

@olehermanse, I hope some of that helps?

olehermanse:
@mikeweilgart Thanks, that helps a lot. Your workflow, using git, is very similar to what I imagined. 

I think this will require some reflection and experimentation to get right. I'm not going to rush this, I'll come back to this around / after 3.12 release.

For the record, I'm open to suggestions from anyone else interested, just comment here :)

olehermanse:
See #1322

PR #1060, opened 2018-03-22
Remove trailing whitespace (git doesn't like it!)

atsaloli:
Remove trailing whitespace from inventory/aix.cf as git complains about
it.

In Git core.whitespace configuration settings, `blank-at-eof` treats
blank lines added at the end of file as an error (enabled by default).

I ran into this when using git to update our MPF. I had to override
it with `git apply --whitespace=warn` to get git to actually apply
the patch to inventory/aix.cf


atsaloli:
:+1:

PR #1062, opened 2018-03-23
Added .ropeproject to .gitignore

olehermanse:


nickanderson:
What's this for?

nickanderson:
Python rope a refactoring library

olehermanse:
It is used by Atom-IDE (Facebook and github's effort to bring IDE like functionality to atom), possibly other editors also.

PR #1064, opened 2018-03-31
Add CoreOS paths to lib/paths.cf

atsaloli:


atsaloli:
I downloaded a CoreOS ISO and booted it up in VMWare and verified all the paths.

nickanderson:
Thank you!

PR #1052, opened 2018-02-19
ENT-2797 Do not manage cf-consumer and redis-server after they have been removed

nickanderson:
This change avoids managing cf-consumer and redis-server after they have been removed.

Knowledge of removal based on cfengine_MAJOR_MINOR hard classes defined for
cfredis_in_enterprise and cfconsumer_in_enterprise in def.cf and update_def.cf

PR #1065, opened 2018-04-05
ENT-2797 Do not manage cf-consumer since it is no longer used

nickanderson:


vpodzime:
@nickanderson, isn't this also part of #1052 already?

nickanderson:
@vpodzime  yes it is, but @olehermanse  asked that I move it to a separate PR because he is working on removing cf-consumer first. If we merge this then I will close #1052 and open a new PR for just the redis stuff.

vpodzime:
OK, thanks for the explanation.

PR #1066, opened 2018-04-06
ENT-3482: Refresh hosts materialized view every 5 minutes.

aleksandrychev:


aleksandrychev:
@nickanderson please review

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -526,6 +526,28 @@ bundle agent cfe_internal_refresh_inventory_view
...
 }
 
+bundle agent cfe_internal_refresh_hosts_view
+# @brief Refresh hosts materialized view every 5 minutes
+{
+  meta:
+
+    (policy_server|am_policy_hub).enterprise_edition::
+
+      "tags" slist => { "enterprise_maintenance" };
```
nickanderson: Good use of tagged bundle for automatic running.

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -526,6 +526,28 @@ bundle agent cfe_internal_refresh_inventory_view
...
+
+  commands:
+
+    (policy_server|am_policy_hub).enterprise_edition::
+
+      "$(sys.workdir)/httpd/php/bin/php"
+        args => "$(cfe_internal_hub_vars.docroot)/index.php cli_tasks materialized_hosts_view_refresh",
+        comment => "This refreshes the hosts view. If the hosts view is not refreshed then it will contain stale data.",
+        handle  => "mpf_fresh_hosts_view",
+        classes => kept_successful_command,
```
nickanderson: I think that we should remove this classes => kept_successful_command constraint.

It causes the command to be interpreted as a promise kept, instead of a promise repaired, but it is actually a change ot the system, so I think we should interpret it so. We have removed this from other promises for similar reasons before.

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -526,6 +526,28 @@ bundle agent cfe_internal_refresh_inventory_view
...
+        if => countlinesmatching("materialized_hosts_view_refresh", "$(cfe_internal_hub_vars.docroot)/application/controllers/Cli_tasks.php");
```
nickanderson: I don't think this constraint makes sense.

countlinesmatching() returns an int not a boolean. https://docs.cfengine.com/docs/3.10/reference-functions-countlinesmatching.html

I guess this should probably be something like `if => isgreaterthan( countlinesmatching(), 0 );`

aleksandrychev:
>I think that we should remove this classes => kept_successful_command constraint.
done

>I guess this should probably be something like if => isgreaterthan( countlinesmatching(), 0 );
thanks, works!

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -526,6 +526,27 @@ bundle agent cfe_internal_refresh_inventory_view
...
+
+    (policy_server|am_policy_hub).enterprise_edition::
+
+      "tags" slist => { "enterprise_maintenance" };
+
+  commands:
+
+    (policy_server|am_policy_hub).enterprise_edition::
+
+      "$(sys.workdir)/httpd/php/bin/php"
```
nickanderson: Lets add a stakeholder for future xreferencing.

```
"$(sys.workdir)/httpd/php/bin/php" -> { "ENT-3482" }
```

nickanderson:
One more change, then please squash.

aleksandrychev:
@nickanderson done.

nickanderson:
Thanks!

PR #1067, opened 2018-04-13
CFE-2797 Fix memory collection on AIX

ncharles:


ncharles:
See https://tracker.mender.io/browse/CFE-2797

inventory/any.cf
```diff
@@ -145,6 +145,11 @@ bundle agent cfe_autorun_inventory_memory
       "free" string => ifelse("windows", "0",
                               $(mon.value_mem_free)),
       meta => { "report" };
+
+     enterprise.aix::
+      "total" string => execresult("/usr/bin/lparstat -i | grep 'Online Memory' | awk '{ print $4 }'", "useshell"),
+      meta => { "inventory", "attribute_name=Memory size (MB)" };
+
```
vpodzime: What about the ``"free"`` variable?

ncharles: you are totally right, thank you !

inventory/any.cf
```diff
@@ -136,15 +136,21 @@ bundle agent cfe_autorun_inventory_memory
...
                                $(mon.value_mem_total)),
       meta => { "inventory", "attribute_name=Memory size (MB)" };
 
+    enterprise::
       "free" string => ifelse("windows", "0",
                               $(mon.value_mem_free)),
       meta => { "report" };
+
+    enterprise.aix::
+      "total" string => execresult("/usr/bin/lparstat -i | grep 'Online Memory' | awk '{ print $4 }'", "useshell"),
```
nickanderson: If we use this execresult to collect memory let's get rid of grep.

This should work:

`/usr/bin/lparstat -i | awk '/Online Memory/ {print $4 }'

nickanderson:
Hum, is it not pulling the right data from cf-monitord or are you not running cf-monitord?

nickanderson:
Can you add the info that it is pulling to the ticket please. Would like to see what it thinks the free mem is from cf-monitord.

inventory/any.cf
nickanderson: @ncharles can you update the PR to drop the extra pipeline?

```
$ lparstat -i | grep Online
Online Virtual CPUs                        : 3
Online Memory                              : 2048 MB
```

```
$ /usr/bin/lparstat -i | grep 'Online Memory' | awk '{ print $4 }'
2048
```

Also, it would be nice if we had the information about what cf-monitord was capturing for memory.

```
$(mon.value_mem_free)
```

nickanderson: oh TOTAL not free, i was looking at the wrong line.

nickanderson:
Thank you @ncharles!  closing this pr in favor of #1073 which includes the small cleanup and a changelog keyword.

PR #1068, opened 2018-04-19
ENT-3730 Fix variable used for redis conf file

nickanderson:
This conf file variable was moved into another bundle as part of some
refactoring to enable a more plugable policy update model for users to override with their own.


olehermanse:
Tested, works!

nickanderson:
No backport necessary. ticket updated accordingly.

PR #1095, opened 2018-05-23
CFE-2827 Slackware package module support

victorqrt:


nickanderson:
Thanks for this. Will you also update the docs for the default package modules?

https://github.com/cfengine/documentation/blob/master/reference/promise-types/packages.markdown#package-modules-out-of-the-box

nickanderson:
CI failed on trailing whitespace:

```
modules/packages/slackpkg:162:        ls -l /var/log/packages/ |sed -e 's/.* \([^ ]\+\)-\([_a-w0-9.?]\+\)-\(.*\)-.*/Name=\1\'$'\nVersion=\\2\\nArchitecture=\\3/'  
FAIL: trailing whitespace was found
```

VictorQrt:
Done.
I know, but the idea here is to make the module work on (very) minimal slackware installs...

modules/packages/slackpkg
```diff
@@ -0,0 +1,188 @@
...
+    read_options
+fi
+
+case "${CMD}" in
+    get-package-data)
+        get_package_data
+        ;;
+    list-installed)
+        ls -l /var/log/packages/ |sed -e 's/.* \([^ ]\+\)-\([_a-w0-9.?]\+\)-\(.*\)-.*/Name=\1\'$'\nVersion=\\2\\nArchitecture=\\3/'
+        ;;
```
craigcomstock: I tried the basic nmap absent example you give in the docs PR https://github.com/cfengine/documentation/pull/1823 and get some errors.

```
root@slackami:/var/cfengine/masterfiles# vim services/main.cf
root@slackami:/var/cfengine/masterfiles# cf-promises services/main.cf
root@slackami:/var/cfengine/masterfiles# cf-agent -Klf update.cf
root@slackami:/var/cfengine/masterfiles# cf-agent -KI
   error: Unsupported response from package module: total 13276
   error: Unsupported response from package module: -rw-r--r-- 1 root root   10515
Feb 22  2014 bind-9.9.3_P2-x86_64-1
   error: Unsupported response from package module: -rw-r--r-- 1 root root    1443
Feb 22  2014 cdparanoia-III_10.2-x86_64-1
   error: Unsupported response from package module: -rw-r--r-- 1 root root    1879
Feb 22  2014 dhcp-4.2.5_P1-x86_64-1
    info: Error updating packages cache.
    info: Some error occurred while updating available updates cache.
   error: Can not update cache.
   error: Unsupported response from package module: total 13232
   error: Unsupported response from package module: -rw-r--r-- 1 root root   10515
Feb 22  2014 bind-9.9.3_P2-x86_64-1
   error: Unsupported response from package module: -rw-r--r-- 1 root root    1443
Feb 22  2014 cdparanoia-III_10.2-x86_64-1
   error: Unsupported response from package module: -rw-r--r-- 1 root root    1879
Feb 22  2014 dhcp-4.2.5_P1-x86_64-1
    info: Error updating packages cache.
    info: Can not update installed packages cache after package installation
   error: Error removing package 'nmap'
```

I'm wondering if you could change the version capture to ```([_a-zA-Z0-9.?]+)``` and somehow get rid of the ```total 13232``` output of ls -l maybe with a ```grep -v "^total"``` in between the ls and sed commands? So something like

```ls -l /var/log/packages/ | grep -v "^total" |sed -e 's/.* \([^ ]\+\)-\([_a-zA-Z0-9.?]\+\)-\(.*\)-.*/Name=\1\'$'\nVersion=\\2\\nArchitecture=\\3/'```

That change fixes things up for me pretty well.
+        ;;

File: modules/packages/slackpkg [hunk 2]
```diff
@@ -0,0 +1,188 @@
...
+        case "${CMD}" in
+            repo-install)
+                slackpkg -dialog=off -batch=on -default_answer=yes install ${PKG} 2>/dev/null >/dev/null
+                ;;
+            file-install)
+                installpkg ${PKG} 2>/dev/null >/dev/null
+                ;;
+            remove)
+                slackpkg -dialog=off -batch=on -default_answer=yes remove ${PKG} 2>/dev/null >/dev/null
+                ;;
```
craigcomstock: I was playing around with install/remove of bsd-games either by repo or file and it seems install/remove works fine for repo but only install works for file.

```
"/tmp/bsd-games-2.13-x86_64-12.txz"
#"bsd-games"
#  policy => "present",
  policy => "absent",
  package_module => slackpkg;
```

Can you test this out and see what might be wrong? Or if you can even reproduce my results?

modules/packages/slackpkg
VictorQrt: Fixed this by changing the ls option to -1, this gives less output for sed to filter and simplifies the regex. Also changed the version capture as you recommended. Strangely this issue never occured to me..

VictorQrt:
Amended the commit

modules/packages/slackpkg
craigcomstock: You know I suppose you are right. What you have currently does conform to the docs so let's go with that. Intuitively I think it makes sense for day-to-day use. The case I was trying is more of a testing-centric case than an everyday use case. :+1:

PR #1100, opened 2018-06-12
make package name matching more accurate

snltd:
The regex matching in `get_package_data()` is too loose. For instance, a promise to install the `gmp` package resulted in `gmp-api` being installed. Because:

```
$ pkgin -pP avail | grep ^gmp-                
gmp-6.1.2;Library for arbitrary precision arithmetic
gmp-api-38.0;Headers for developing Gecko Media Plugins
```

We now look for `Name-[0-9]` in the available package list. The regex can't be any tighter than that, as it seems pretty much any character is allowed in the version number. So `File-[0-9\.]*;` will miss packages like `gmpc-tagedit-11.8.16nb8`.

### DIFF
--- a/modules/packages/pkgsrc
+++ b/modules/packages/pkgsrc
@@ -102,7 +102,7 @@ get_package_data () {
         # or version, return nothing.
         # There's possibly a bug here because we're already emitting that the
         # PackageType is repo.
-        parse_pkg_data "$(pkgin -pP avail | grep "^${File}-" | grep "$Version;" | sort -n | tail -1)" | grep Name
+        parse_pkg_data "$(pkgin -pP avail | grep "^${File}-[0-9]" | grep "$Version;" | sort -n | tail -1)" | grep Name
     fi
 }
 


cf-bottom:

craigcomstock:
I took a look at debian and redhat package naming conventions and agree that your change seems reasonable.

https://www.debian.org/doc/manuals/debian-faq/ch-pkg_basics.en.html
https://fedoraproject.org/wiki/Packaging:Naming
https://fedoraproject.org/wiki/Packaging:Versioning

For FreeBSD it seems it will work as well
```The package name is defined as ${PKGNAMEPREFIX}${PORTNAME}${PKGNAMESUFFIX}-${PORTVERSION}.```
 from https://www.freebsd.org/doc/en/books/porters-handbook/makefile-naming.html

It would be great if we could test this change with all of the currently supported package modules: apt_get, freebsd_ports, nimclient, pkg, pkgsrc, yum and zypper

https://github.com/cfengine/masterfiles/tree/master/modules/packages

We do have some unit and acceptance tests already in 

https://github.com/cfengine/masterfiles/tree/master/tests

So I will trigger a build internally where we can test several of these platforms with our test suite.

craigcomstock:
@snltd our internal builds seemed to work well on a variety of platforms: rhel, debian, ubuntu, solaris and a few other exotics.

@olehermanse or @vpodzime could you double check this PR? I know it is a very small change but it seems it could have a big impact so want a second set of eyes to look. Thanks!

craigcomstock:
@snltd thanks for the contribution!

craigcomstock:
I tried to cherry pick to 3.7.x and found that pkgsrc was not present in that branch so didn't seem like a good candidate for back porting. @olehermanse

PR #1105, opened 2018-06-20
Store the epoch of packages in cache db with zypper

amousset:
When a package has an epoch, in order to install the precise version
with zypper, it is required to write the complete version `epoch:version`
and `version` alone is not understood.

But currently the installed package cache does not store the
epoch, and a package with an epoch in the version will never be considered
as present.

This commits adds the `${epoch}:` before the version of the package
in the installed package database, allowing to install and
properly detect the package.


craigcomstock:
Thanks! We will review soon.

PR #1106, opened 2018-06-25
ENT-3883 Fix variables so that cfengine services start on windows

nickanderson:


nickanderson:
Approved by @craigcomstock

PR #1125, opened 2018-07-11
CFE-2860 Specify scope => "namespace" when using persistent classes

nickanderson:
If scope is not set to namespace, then info messages are emitted and the class
is automatically scoped to namespace. This change prevents annoying messages and
retains the same behavior.

(cherry picked from commit fc029033785a82ab0f742741722e335a523d9681)


nickanderson:
The patch passed in master, 3.12, and 3.7. The failure for this pr must be something unrelated.

PR #1131, opened 2018-07-16
ENT-3958 Make recommendations about postgresql.conf

nickanderson:


cfe_internal/recommendations.cf
```diff
@@ -0,0 +1,66 @@
...
+      "conf[shared_buffers]"
+        string => "16GB",
+        if => isgreaterthan( "$(mem_info_data[MemTotal][1])", $(upper) );
+
+      "conf[effective_cache_size]"
+        string => "11GB", # 70% of 16GB
+        if => isgreaterthan( "$(mem_info_data[MemTotal][1])", $(upper) );
+
+      "calculated_shared_buffers_MB"
+        string => format( "%d0",
```
vpodzime: What does the `0` at the end of the format string mean?

vpodzime:
```

File: cfe_internal/recommendations.cf (hunk 1 above)
Comment (nickanderson) [reply cfe_internal/recommendations.cf]: it formats it to an integer with 0 decimals.

**Comment by nickanderson** [comment]
```
R: dcs_passif_output: the output of command '"/home/travis/build/cfengine/masterfiles/tests/acceptance/workdir/__00_basics_def_json_control_common_bundlesequence_end_cf/bin/cf-agent" -Kf /home/travis/build/cfengine/masterfiles/tests/acceptance/workdir/__00_basics_def_json_control_common_bundlesequence_end_cf/inputs/promises.cf' was: '   error: A variable seems to have been used for the name of the method. In this case, the promiser also needs to contain the unique name of the method
   error: A method attempted to use a bundle '$(recommendation_bundles)' that was apparently not defined
R: reporting actuation of bundle x'
```

So, it errored on 3.7.x because of an empty variable and that is what failed the test. It did actually report the expected output. So I just need to tidy it up a little and make it more compatible with 3.7.x.

PR #1136, opened 2018-07-27
CFE-2868: Analyze yum return code before parsing its output

Lex-2008:


modules/packages/yum
```diff
@@ -126,19 +126,19 @@ def list_updates(online):
         # Nothing to do for us here anyway
         return process.returncode
     lastline = ""
-    for line in stdoutdata:
+    for line in stdoutdata.split('\n'):
```
vpodzime: There's `stdoutdata.splitlines()`.

modules/packages/yum
```diff
@@ -126,19 +126,19 @@ def list_updates(online):
...
         # Combine multiline entries into one line. A line without at least three
         # space separated fields gets combined with the next line, if that line
         # starts with a space.
-        if lastline and not line[0].isspace():
+        if lastline and (len(line)==0 or not line[0].isspace()):
```
vpodzime: Spaces around `==`. Also I think this would be better readable as
```
if lastline and len(line) > 0 and not line[0].isspace()
```

Lex-2008: no, in your case this expression will evaluate to False if `len(line) == 0`, while my idea is that empty `line` means that "`line` does not start with space" and the expression should evaluate to True.

Lex-2008:
good point

PR #1137, opened 2018-08-03
log file should only be on hub, causing errors on windows

estenberg:
See https://tracker.mender.io/browse/ENT-3946

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -300,13 +300,13 @@ bundle common def
         "$(sys.workdir)/outputs/dc-scripts.log",
         "$(sys.workdir)/state/promise_log.jsonl",
         "$(sys.workdir)/state/classes.jsonl",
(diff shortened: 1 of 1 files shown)


estenberg:
Should probably be backported to stable branches if approved.

craigcomstock:
This does seem like a good and obvious change. Also Nick's recommendation from the JIRA ticket would be good.

> Craig Comstock in lib/bunldes.cf see bundle agent logrotate.
>
> Try this:
> 
> {
>   files:
>       "$(log_files)"
>       comment     => "Rotate file if above specified size",
>       rename      => rotate("$(rotate_levels)"),
>       file_select => bigger_than("$(max_size)"),
>       if => fileexists( $(log_files) );
> }
> 
> I added the if => fileexists attribute.
>

PR #1144, opened 2018-08-10
ENT-3946 Only consider files that exist for rotation

nickanderson:


cf-bottom:

PR #1148, opened 2018-08-10
CFE-2855 apt_get package module includes held packages when listing updates

nickanderson:
We ignore held packages (`--ignore-hold`) so that all package updates available
are listed. This makes package update listing compatible with debian 8 and
higher `apt list --upgradeable`.


cf-bottom:

PR #1154, opened 2018-08-14
Redirect output from 'geninstall' on AIX

vpodzime:
Somehow dropping the '-q' option and redirecting the output seems
to fix the issues with self-upgrade not properly (re-)starting
services on AIX.

Ticket: ENT-3972


cf-bottom:

nickanderson:
Superceeded by #1155  which has the correct commit and PR prefix for automatic linking in Jira.

PR #1155, opened 2018-08-14
ENT-3972 Fix unattended self upgrade on AIX

nickanderson:
This change resolves an issue where the package fails to finish upgrading on
AIX.


cf-bottom:

PR #1157, opened 2018-08-14
ENT-3972 Fix unattended self upgrade on AIX for 3.10.x

nickanderson:
For unknown reasons the package fails to complete a self upgrade if the
output is not re-directed.


nickanderson:
Rebased

PR #1158, opened 2018-08-14
Sort indices in insert_ini_section()

nickanderson:
The order of the key-value pairs doesn't matter in INI files, but
they are easier to read if it they are sorted by key.

Also make sure indices are sorted in the insert_ini_section()
test. Otherwise they may be in arbitrary order.

Use reverse(sort()) to make sure the sections in the resulting
file are alphabetically ordered (in combination with how
insert_ini_section() works).

(cherry picked from commit 8f8966f3460cae6cbacfd7feddaf9ff231d82209)


cf-bottom:

PR #1159, opened 2018-08-15
Explanation why stdout redirect fixes package operations on AIX

vpodzime:
This is quite a tricky thing to figure out so let's note it down
before we forget about it.

Ticket: ENT-3972


vpodzime:
See #1155

PR #1174, opened 2018-09-05
Added sles hardclass

dstam:
This is done because parsing the /etc/os-release results in the harclasses sles and
`sles_<version>`. The suse and `suse_<version>` are only available on SLES 12 SP 3 or
lower because they still have the /etc/SuSE-release. This has been removed in SLES 15.

I have added the new hardclass instead of replacing it. So it's compatible with CFEngine 3.10.x LTS and 3.12.x LTS


cf-bottom:

PR #1176, opened 2018-09-11
CFE-2894: Travis fails to install cfengine on masterfiles

Lex-2008:


nickanderson:
Nice

Lex-2008:
I see some errors in Travis, but I believe they're not related to this PR. @nickanderson : ok to merge?

Also, note that CF_VERSION=master actually installs same CFEngine version as CF_VERSION=3.12.x...

PR #1182, opened 2018-09-11
CFE-2897 Consider sles when considering suse

nickanderson:
Added sles hardclass This is done because parsing the /etc/os-release results in the harclasses sles and sles_<version>. The suse and suse_<version> are only available on SLES 12 SP 3 or lower because they still have the /etc/SuSE-release. This has been removed in SLES 15.


nickanderson:
Thanks, @dstam. Looks good to me. Will cherry-pick to 3.12.x

PR #1183, opened 2018-09-11
CFE-2897 Consider sles when considering suse

nickanderson:


cf-bottom:

PR #1188, opened 2018-09-21
ENT-4807 Cleanup old report data relative to the most recent changetimestamp

nickanderson:


nickanderson:
@cf-bottom jenkins please

PR #1191, opened 2018-09-24
ENT-4092 Prevent standalone self upgrade from triggering un-necessarily

nickanderson:


cf-bottom:

PR #1195, opened 2018-09-25
ENT-4116 Remove un-necessary agent run during self upgrade

nickanderson:
The cf-agent run that is executed after upgrade is not necessary, and in some
cases can contribute to the agent upgrade failing.

Changelog: Title
(cherry picked from commit faaf46e82ede24f5b8a173f5bd021fa365e85119)


cf-bottom:

PR #1196, opened 2018-09-25
ENT-4117 Fix standalone self upgrade when path contains spaces

nickanderson:
Noteably on windows the path to workdir commonly contains spaces. The execution fails if the path is not quoted or the spaces are not escaped.

Changelog: Title


cf-bottom:

PR #1197, opened 2018-09-25
ENT-4117 Fix standalone self upgrade when path contains spaces

nickanderson:


cf-bottom:

PR #1211, opened 2018-10-17
CFE-2922 Avoid trying to read /proc/meminfo when it doesnt exist

nickanderson:
Because of pre-evaluation the agent attempts to parse the file and fill the variable. Guarding it will avoid generating un-necessary errors where the policy is not explicitly active.


cf-bottom:

PR #1212, opened 2018-10-17
CFE-2922 Avoid trying to read /proc/meminfo when it doesn't exist

nickanderson:
On hosts that this policy is not explicitly active on we see errors trying to
read the file because of pre-evaluation.

Changelog: Title
(cherry picked from commit df4d1ea59692834b28266b6d26035552fb7f5e7f)


cf-bottom:

PR #1213, opened 2018-10-17
CFE-2932 Detect systemd service enablement for non native services

nickanderson:
Because non-native services have UnitFileState=bad we need an alternate way to detect if the service is enabled or not.


cf-bottom:

olehermanse:
@nickanderson This PR has the wrong ticket number in title.

PR #1216, opened 2018-10-18
CFE-2932 Detect systemd service enablement for non native services 3.7.x

nickanderson:
Because non-native services have UnitFileState=bad we need an alternate way to
detect if the service is enabled or not. Since systemctl is-enabled
logs the fact that the service is not native to stderr, the probe must
be run in a shell and the output must be redirected to /dev/null or the
agent will output the information.

Changelog: Title
(cherry picked from commit a3b4260b81dae8a48b3dee55589616d6d47a82d9)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
         apt:
           sources:
             - sourceline: "deb https://cfengine-package-repos.s3.amazonaws.com/pub/apt/packages stable main"
-            - key_url: https://cfengine-package-repos.s3.amazonaws.com/pub/gpg.key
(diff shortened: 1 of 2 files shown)


cf-bottom:

olehermanse:
@nickanderson This PR has the wrong ticket number in title.

nickanderson:
Dammer, i think it was supposed to be CFE-2923 not CFE-2932

nickanderson:
At least I got the number correct in the branch name! PRs to other branches renamed as well. Good spot.

PR #1225, opened 2018-10-19
CFE-2927 Allow dmidefs inventory to be overridden via augments

nickanderson:
This change allows uses to override the default dmidecode attributes that are
inventoried via augments. Being able to extend the inventoried attributes allows
for more flexibility without having to maintain modifications to the vendored
policy.


cf-bottom:

PR #1233, opened 2018-10-24
ENT-4158 Inventory Setuid Files

nickanderson:
When the agent observes a file that is setuid it is logged to
`$(sys.workdir)/cfagent.$(hostname -f).log`.

This change introduces inventory for these observed files. For each file in the
log that is setuid inventory of that file is tagged with `setuid Files`.
Additionally separate inventory is done for each root owned setuid file as `root
owned setuid Files`. This policy also handles clearing of files that were once
setuid but are no longer found or are no longer setuid from the log since there
is no pre-existing automatic processes to clear entries from the log.

Changelog: Title
(cherry picked from commit 06027d875638e1597c7dfb8337548d66f4375354)

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -167,6 +167,70 @@ bundle agent cfe_autorun_inventory_memory
 
 }
 
+bundle agent cfe_autorun_inventory_setuid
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #1235, opened 2018-10-24
Fix license utilization performance

mikeweilgart:
License utilization logging unavoidably involves some performance hit due to the function calls hostsseen() and hostswithclass().  However, what *is* avoidable is taking this performance hit on *every* run rather than only taking the hit every 12 hours (which is how often the logging is actually done).  This pull request fixes the extraneous function calls.

Some cleanup of class expressions in this bundle seemed in order and were done at the same time; see commit messages for rationale.

### DIFF
--- a/cfe_internal/enterprise/CFE_hub_specific.cf
+++ b/cfe_internal/enterprise/CFE_hub_specific.cf
@@ -374,7 +374,7 @@ bundle agent log_cfengine_enterprise_license_utilization
 # @brief Log the number of hosts seen within the last 24 hours and the number of
 # hosts reported to have the "cfengine" class. Note any hosts that has been
 # successfully collected from is expected to have the "cfengine" class. This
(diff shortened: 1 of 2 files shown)


cf-bottom:

nickanderson:
Also thanks for each separate commit, it made reviewing this much easier.

mikeweilgart:
> Also thanks for each separate commit, it made reviewing this much easier.

Thanks!

In that case, why squash them?  :)

nickanderson:
It's fine to leave them as seperate commits. But then please rename your branch to something starting with ENT-4190 and in the last commit add a line

 Changelog: Improve performance of enterprise license utilization logging

mikeweilgart:
@nickanderson okay, done - #1248 has the renamed branch and the added commit message line, for you to merge.  :)  I'm closing this PR accordingly.

PR #1236, opened 2018-10-25
ENT-152 Inventory memory on Windows and HPUX

nickanderson:


cf-bottom:

nickanderson:
@craigcomstock if you can double check/test this it would be great

inventory/any.cf
```diff
@@ -141,22 +141,43 @@ bundle agent cfe_autorun_inventory_memory
...
+@if minimum_version(3.11)
+      # The `with` attribute is necessary for this to work in a single promise.
+    enterprise.windows::
+      "total"
+        meta => { "inventory", "attribute_name=Memory size (MB)" },
+        string => format( "%d", eval("$(with)/1024", "math", "infix" )),
+        if => not( isvariable( "total" ) ),
+        with => nth( string_split( execresult("wmic OS get TotalVisibleMemorySize /format:list", useshell ),
+                                   "=", 2), 1);
+      "totalPhysical"
```
craigcomstock: A comment on the first one would be nice. Like:

```
# wmic returns "TotalVisibleMemorySize=10760224" so split on = and take the second item (0-based with nth())
```

inventory/any.cf
```diff
@@ -141,22 +141,43 @@ bundle agent cfe_autorun_inventory_memory
...
-                               $(mon.value_mem_total)),
-      meta => { "inventory", "attribute_name=Memory size (MB)" };
-
-    enterprise::
-      "free" string => ifelse("windows", "0",
-                              $(mon.value_mem_free)),
-      meta => { "report" };
-
+@if minimum_version(3.11)
+      # The `with` attribute is necessary for this to work in a single promise.
```
craigcomstock: Shouldn't there be an @else for this minimum_version guard? What happens in our 3.10 LTS for this? Shouldn't it return 0 for total and free as before?

craigcomstock:
> @craigcomstock if you can double check/test this it would be great

Will do!

inventory/any.cf
nickanderson: ACK

inventory/any.cf
nickanderson: AFAIK we do not have an @else . The important thing is that the parser will complain on any version older than 3.11 if it sees the with the attribute in use.

reporting 0 is lame.

craigcomstock: Ok. No @else. @nickanderson I tested with 3.10 and 3.12 agents and this policy seems fine. I think the difference is not too important.

PR #1243, opened 2018-10-29
ENT-4192 Add policy to enable policy analyzer

nickanderson:


nickanderson:
Tested with success by @aleksandrychev

PR #1248, opened 2018-11-01
Ent 4190 hub license usage logging performance

mikeweilgart:
This is exactly the same as PR #1235 but with the branch renamed and the changelog line added to the last commit message as per Nick Anderson's request.


cf-bottom:

PR #1251, opened 2018-11-02
CFE-2905 An improved autorun methodology

nickanderson:
Until now, when the class `services_autorun` was defined, files in `services/autorun` ending in `.cf` were added to inputs, and any bundles found tagged with `autorun` were executed in lexical order.

This change brings a new methodology that better allows for polices needing more than a single asset.

Now, when `services_autorun` is defined, the services directory will be searched one level deep for directories named `policy` that contains a policy file matching the current entries name. Those matching files will be added to inputs. Bundle execution is controlled by a control.yaml or control.json file in the root of the subdirectory which contains keys matching the expected entries name, and subkeys for the namespace and bundle name to execute. These bundles are executed in lexical order of the directories name.

For example, given the following files:

- `services/beta/control.json`
- `services/beta/policy/example.cf`
- `services/alpha/control.json`
- `services/alpha/policy/promises.cf`

When the default policy is executed (`/var/cfengine/inputs/promises.cf`) with `services_autorun` defined, `services/alpha/policy/promises.cf` will be added to inputs.

Given the following `services/alpha/control.json`:

```
{ 
  "promises.cf": { 
    "namespace": "default",
    "bundle": "alpha_main"
  }
}
```

The bundle `alpha_main` in the `default` namespace will be automatically executed.

Changelog: Title

### DIFF
--- a/lib/autorun.cf
+++ b/lib/autorun.cf
@@ -7,15 +7,56 @@ bundle agent autorun
 {
   vars:
     services_autorun::
-      "bundles" slist => bundlesmatching(".*", "autorun");
+@if minimum_version(3.11)
+      "services_prefixed_classes"
+        slist => classesmatching( "services_.*" );
+      # For backwards compatibility during upgrades, this must be guarded
+      # because the with attribute is used.
 
-      "sorted_bundles"
-        slist => sort("bundles", "lex"),
+      "services_autorun_input[$(with)]" -> { "CFE-2905" }
+        string => "$(sys.workdir)/inputs/services/$(with)/$(sys.policy_entry_basename)",
+        with => nth(string_split( "$(services_prefixed_classes)", "_", 2), 1),
+        if => fileexists( "$(sys.workdir)/inputs/services/$(with)/$(sys.policy_entry_basename)" );
+
+      # Both yaml and json formatted control files are supported.
+
+      "data_$(bundles)"
+        data => readyaml( "$(sys.workdir)/inputs/services/$(bundles)/control.yaml"),
+        if => fileexists( "$(sys.workdir)/inputs/services/$(bundles)/control.yaml" );
+
+      "data_$(bundles)"
+        data => readjson( "$(sys.workdir)/inputs/services/$(bundles)/control.json"),
+        if => fileexists( "$(sys.workdir)/inputs/services/$(bundles)/control.json");
+
+      # This is each directory inside services that has a matching policy entry
+      # e.g. services/example/promises.cf then the list would have an element
+      # matching example.
+      "services_autorun_policies"
+        slist => getindices( services_autorun_input );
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
@tzz good or bad?

tzz:
This looks like a nice way to deliver bundles as directories. But it moves the control into a data language one level deeper than users know to expect. Maybe the control data could live in `def.json`?

nickanderson:
My thought was the bundle author would know best how to specify the entry for that set. Instead of burdening the person who is trying to use the policy directory

basvandervlies:
@tzz and @nickanderson the surfsara_lib already does dynamic inputs only for `promises.cf`, but you must add the service  to a `def.json` file:
```
"sara_services_enabled": [
    "postfix"
]
```
This will only load/parse/execute `postfix.cf` from all surfsara services. The other are not loaded at all. I use the autorun feature to enable loading of the `surfsara_lib`  in the MPF.    we can easily extend the promises.cf with services by this method. 

 Maybe we can integrate the libs..  

If I read it correctly. What Nick proposes you can also extend the `update.cf` process. I will use it to load the `surfsara` lib.

nickanderson:
indeed @basvandervlies, you do see my thoughts for the future. I haven't yet resolved in my head the specifics around policy sharing to support the functionality from other entries, but yes, the idea is that we could add the support for any policy entry, making it easy to share a bit of policy that extends the update policy (update.cf), or the main policy (promises.cf), or the self-upgrade policy (standalone-self-upgrade.cf) or any other custom entry point.

basvandervlies:
thanks @nickanderson for the update.

basvandervlies:
@nickanderson  I am more for a "standard" method of invoking services and using the meta tags just like the `cf_surfsara_lib`.  Each service has its own meta tag `template_$(service)`. People only have to add this meta tag keyword to the services and it run.  Maybe you allow to override entry point by defining the `bundle|entry_point` parameter.

This will automatically include `nick.cf` from a standard directory , `surfsara_lib` it is `services/surfsara/nick.cf` and run it. You can define more parameters if you want, eg `def.json`
```
`services_enabled`: [
    "nick"
]
"nick": {
  "bundle":  "<override_entry_point>"
  ....
}
```

You can have the same approach for `update.cf`:
````
"update_services" : [
  "nick"
]
```
Same behaviour as above. With the multiple `json` file  it can be in a separate json file:
 * services.json
 * update.json
 * ...

nickanderson:
I am not sure that I fully follow, @basvandervlies.

maybe we should move the discussion to the ticket .... https://tracker.mender.io/browse/CFE-2905

nickanderson:
Closing, not going to pursue this.

PR #1260, opened 2018-11-12
ENT-4235 Make runalerts quiet during normal operation 3.10.x

nickanderson:
These frequent log messages add up in large environments, we only want to log
information when there is a problem.


cf-bottom:

nickanderson:
@aleksandrychev please review, note this is slightly different from master, but the same as the pr to 3.12.x

PR #1264, opened 2018-11-13
ENT-4283 collect_window in body server control can now be set from au…

nickanderson:


nickanderson:
TODO fix history in docs, this isnt going to make it into those upcoming releases.

controls/def.cf
```diff
@@ -408,6 +408,10 @@ bundle common def
         string => "5",
         unless => isvariable(control_server_call_collect_interval);
 
+      "control_server_collect_windows" -> { "ENT-4102" }
```
vpodzime: window**s**

nickanderson: BOOM, thanks!

PR #1268, opened 2018-11-16
Fix an example of setting control_hub_hub_schedule to be a valid JSON

vpodzime:
Otherwise it fails to parse.


cf-bottom:

nickanderson:
YES!

PR #1270, opened 2018-11-19
ENT-4255 Add debian 9 to the self upgrade package map

nickanderson:
This change causes the policy hub to try and download the debian 7 package for
the debian 9 hosts and for debian 9 hosts to look for the debian 7 package.

Changelog: None
(cherry picked from commit 0d4131d6b537d4c57550255d717ed0c9f2dbae6c)


cf-bottom:

PR #1271, opened 2018-11-19
[3.12.x] Fix an example of setting control_hub_hub_schedule to be a valid JSON

vpodzime:
Otherwise it fails to parse.

(cherry picked from commit f60fb1535af53ebc26d5b219c139ca8cd2419002)


vpodzime:
Applied cleanly, merging.

PR #1280, opened 2018-11-27
ENT-4273 Restart runalerts if modified

nickanderson:
This change causes the runalerts.php process to be killed if the script has been altered. There already exists a promise to start the service if appropriate.

If the service is not restarted, then the changes to the runalerts script will not be in effect, and Mission Portals ability to notify on alert conditions may be impaired.


cf-bottom:

PR #1281, opened 2018-11-30
CFE-2953 - fix isvariable syntax error in update_def.cf

beto7591:
This error prevents pulling policy updates from from def.mpf_update_policy_master_location:

if => isvariable( $(def.mpf_update_policy_master_location) );

should be:

if => isvariable( "def.mpf_update_policy_master_location" );


cf-bottom:

nickanderson:
Merging. Failure on 3.12.x branch is related to lack of nightly package including the cf-check binary and is completely un-related

PR #1285, opened 2018-12-04
ENT-4273 Restart runalerts if modified

nickanderson:
This change causes the runalerts.php process to be killed if the script has been
altered. There already exists a promise to start the service if appropriate.

If the service is not restarted, then the changes to the runalerts script will
not be in effect, and Mission Portals ability to notify on alert conditions may
be impaired.

Changelog: Title
(cherry picked from commit 9130dae2f02be7794aa158c873bde3d38c5190c8)


cf-bottom:

PR #1287, opened 2018-12-04
ENT-4283 collect_window in body server control can now be set from augments

nickanderson:


cf-bottom:

PR #1291, opened 2018-12-12
ENT-4317 Fix standalone_self_upgrade.cf not triggering because of stale data

nickanderson:


cf-bottom:

PR #1293, opened 2018-12-14
ENT-4333: Prevent performance overhead on hubs that don't enable license utilization logging 3.13.x

nickanderson:


cf-bottom:

PR #1294, opened 2018-12-14
ENT-4333: Prevent performance overhead on hubs that don't enable license utilization logging 3.12.x

nickanderson:


cf-bottom:

PR #1298, opened 2018-12-20
ENT-4331 Handle cleanup of status table after it's been renamed

nickanderson:
Merge together:
https://github.com/cfengine/nova/pull/1329

### DIFF
--- a/lib/cfe_internal_hub.cf
+++ b/lib/cfe_internal_hub.cf
@@ -202,8 +202,21 @@ bundle agent cfe_internal_database_cleanup_consumer_status (row_count)
 {
 
   vars:
+
+    any::
+
+      "status_table_name" -> { "ENT-4331" }
+        string => "__status";
+
+    cfengine_3_7|cfengine_3_10|cfengine_3_12_0|cfengine_3_12_1::
+
+      "status_table_name" -> { "ENT-4331" }
+        string => "status";
+
+    any::
+
       "remove_query"
-      string => "DELETE FROM status WHERE ts IN (SELECT ts FROM status ORDER BY ts DESC OFFSET $(row_count));";
+        string => "DELETE FROM $(status_table_name) WHERE ts IN (SELECT ts FROM $(status_table_name) ORDER BY ts DESC OFFSET $(row_count));";
 
   commands:
       "$(sys.bindir)/psql cfdb -c \"$(remove_query)\""


lib/cfe_internal_hub.cf
```diff
@@ -202,8 +202,14 @@ bundle agent cfe_internal_database_cleanup_consumer_status (row_count)
 {
 
   vars:
+
+      "status_table_name" -> { "ENT-4331" }
+        string => ifelse( eval( "$(sys.cf_version) < 3.12.1", class, infix ), "__status",
+                          "status"),
```
vpodzime: Either a don't understand how `eval` and `ifelse` work or this is wrong. The name of the table is `status` on 3.12.1 and older, `__status` on 3.12.2 and newer.

nickanderson: Oh i put it in backwards.
DMANit

nickanderson:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1876/)

https://ci.cfengine.com/job/pr-pipeline/1876/

lib/cfe_internal_hub.cf
```diff
@@ -204,8 +204,8 @@ bundle agent cfe_internal_database_cleanup_consumer_status (row_count)
   vars:
 
       "status_table_name" -> { "ENT-4331" }
-        string => ifelse( eval( "$(sys.cf_version) < 3.12.1", class, infix ), "__status",
-                          "status"),
+        string => ifelse( eval( "$(sys.cf_version) < 3.12.2", class, infix ), "status",
+                          "__status"),
```
vpodzime: Would be better to use `cf_version_major/minor/patch`, otherwise the comparison fails on pre-release builds because variables there look like this:
```
default:sys.cf_version                   3.12.2a.8f887ce                                              inventory,source=agent,attribute_name=CFEngine version
default:sys.cf_version_major             3                                                            source=agent                            
default:sys.cf_version_minor             12                                                           source=agent                            
default:sys.cf_version_patch             2                                                            source=agent
```

vpodzime:
Unfortunately, the `eval ( "3.12.2 < 3.12.2" )` doesn't seem to work as expected. Also, the documentation of `eval()` says:
> comparison <, <=, >, >= operators added in 3.10.0
so I'm not sure if they are available in 3.7.8.

lib/cfe_internal_hub.cf
nickanderson: I could not resolve the full matrix of comparisons trying to compare the parts separately in my head.
Maybe you have some simple logic that I have overlooked, but I did not get very far down the road before the expressions was not ledible.

nickanderson:
OK @vpodzime, I punted.

Now using version based classes. It's ugly, but nothing fancy here.

vpodzime:
> Looks pretty neat to me. Let me test if it works properly.

Works great! @nickanderson please merge this and forward-port it to master.

nickanderson:
And squash these down to 1?

vpodzime:
> And squash these down to 1?

Ideally, yes.

nickanderson:
Squashed, and its already in master.

https://github.com/cfengine/masterfiles/blob/master/lib/cfe_internal_hub.cf#L206-L216

lib/cfe_internal_hub.cf
```diff
@@ -202,8 +202,21 @@ bundle agent cfe_internal_database_cleanup_consumer_status (row_count)
...
+
+    any::
+
+      "status_table_name" -> { "ENT-4331" }
+        string => "__status";
+
+    cfengine_3_7|cfengine_3_10|cfengine_3_12_0|cfengine_3_12_1::
+
+      "status_table_name" -> { "ENT-4331" }
+        string => "status";
```
olehermanse: @nickanderson So you are relying on the ordering of vars promises here to overwrite. Is this something we do in other places? Should we document that it's okay, here: https://docs.cfengine.com/docs/3.12/reference-promise-types-vars.html ?

nickanderson: I thought I responded to this before. Yes, I am relying on the order of vars and over-writing. It's documented on the page you reference under the policy attribute. Happy to improve the docs if you have a suggestion as to specifically where should include something else.

nickanderson: The behaviour is not really different from other promise types.  Normal Ordering https://docs.cfengine.com/docs/3.12/reference-language-concepts-normal-ordering.html#agent-evaluation-step notes:

The order of promises within one of the above types follows their top-down ordering within the bundle itself. The order may be overridden by making a promise depend on a class that is set by another promise, or by using the depends_on attribute in the promise.

So you may want to sprinkle your doc pr in a couple places

https://docs.cfengine.com/docs/3.12/reference-language-concepts-variables.html
https://docs.cfengine.com/docs/3.12/reference-language-concepts-normal-ordering.html
https://docs.cfengine.com/docs/3.12/reference-promise-types-vars.html

olehermanse: https://github.com/cfengine/documentation/pull/2064

PR #1303, opened 2018-12-26
Improve efficiency and error handling of user specified policy update bundle

nickanderson:
Changelog: Title

Fix MPF reporting both "found" and "missing" regarding the user-defined
update policy bundle.

First, note that
https://docs.cfengine.com/docs/3.10/reference-functions-some.html says:

> It's convenient to set a class to not => some(".*", mylist) in order to
> check if mylist is empty. Since some() returns as soon as possible, that
> is better than using length() or every() or none() which must traverse
> the entire list.

So I switched to some().

Second, let's stop relying on negative knowledge (absense of "found"
class) and instead rely on positive knowledge only ("found" class
for if found; and "missing" class for if missing).

(cherry picked from commit 26eb50bd827e88041fac2c90b026f781e79d9055)

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -30,7 +30,12 @@ bundle agent cfe_internal_update_policy
       # (otherwise we may get an error about undefined bundle)
 
       "have_found_user_specified_update_bundle"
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #1307, opened 2019-01-02
ENT-4365: Change cleanup consumer status SQL query

aleksandrychev:
Changelog: Title
Use a new cleanup_historical_data SQL function that keeps a specified number of items for a host

PR #1308, opened 2019-01-09
ENT-4365: Change cleanup consumer status SQL query

nickanderson:
Use a new cleanup_historical_data SQL function that keeps a specified number of items for a host

Superceeds #1307


cf-bottom:

PR #1310, opened 2019-01-10

nickanderson:


cf-bottom:

lib/cfe_internal_hub.cf
```diff
@@ -209,17 +209,19 @@ bundle agent cfe_internal_database_cleanup_consumer_status (row_count)
         expression => returnszero( "$(sys.bindir)/psql cfdb -c \"SELECT 'cleanup_historical_data'::regproc;\" > /dev/null 2>&1", useshell ),
         if => isexecutable( "$(sys.bindir)/psql" );
 
+      "has_sql_table___status" -> { "ENT-4331" }
+        expression => returnszero( "$(sys.bindir)/psql cfdb -c \"SELECT 'public.__schema'::regclass;\" > /dev/null 2>&1", useshell ),
```
craigcomstock: I would change this variable to cfdb_schema since that's what it is and you grep it later on.

lib/cfe_internal_hub.cf
```diff
@@ -209,17 +209,19 @@ bundle agent cfe_internal_database_cleanup_consumer_status (row_count)
...
-    cfengine_3_7|cfengine_3_10|cfengine_3_12_0|cfengine_3_12_1::
+      # The status table changed it's name for 3.12.2 and 3.10.6, this handles
+      # using the proper table name in the cleanup query
 
       "status_table_name" -> { "ENT-4331" }
-        string => "status";
+        string => ifelse( "has_sql_table___status", "__status", "status");
 
     any::
       # This is the fallback query to use in case the cleanup_historical_data
```
craigcomstock: Another thing I noticed when reviewing the back-port PR (https://github.com/cfengine/masterfiles/pull/1311) is that the fallback query still uses 50000 hard-coded instead of $(row_count). Is that OK?

lib/cfe_internal_hub.cf
vpodzime: Yeah, this is really confusing.

nickanderson: fixed, changed to __status

nickanderson: Also fixed in other PRs.

nickanderson:
@vpodzime @craigcomstock great catch on the table name in the probe. It's fixed now, please re-review.

PR #1311, opened 2019-01-10

nickanderson:


cf-bottom:

craigcomstock:
Shouldn't the commit message include the ticket number?

nickanderson:
There are so many commits for this little section of code, there are already changelog entries, and the association is alreayd there with the branch so i just opted to not include the ticket in the title.

PR #1313, opened 2019-01-11
CFE-2974: Separate kill signals from restart class to avoid warning with minor cleanups

nickanderson:


cf-bottom:

PR #1317, opened 2019-01-25
CFE-2993: Avoid calling yum at every agent run

amousset:
As it was part of a common bundle, it was executed everytime even when
not using old package promises. Furthermore, this simple yum command lead
to contact the server in some cases. As the option was introduced in RHEL 6.1,
use the RHEL version instead of dynamic detection.

### DIFF
--- a/lib/packages.cf
+++ b/lib/packages.cf
@@ -262,18 +262,11 @@ bundle common rpm_knowledge
 
 bundle common redhat_no_locking_knowledge {
   vars:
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
Also found reference to yum-3.2.28

vpodzime:
> I like it.

Then please merge (and backport?) this.

nickanderson:
Thank you!

PR #1319, opened 2019-02-05
CFE-2993 Reduce cost of knowing when setopt is available in yum

nickanderson:


cf-bottom:

PR #1322, opened 2019-02-25
ENT-3554: Tooling to upgrade MPF files (WIP)

olehermanse:
See #1058

### DIFF
--- a/contrib/policy_upgrade/generate_hashes.py
+++ b/contrib/policy_upgrade/generate_hashes.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+import json
+import hashlib
+
+from collections import OrderedDict
+
+
+def json_string(data):
+    return json.dumps(data, indent=2)
+
+
+def json_print(data):
+    print(json_string(data))
+
+
+def json_save(data, path):
+    string = json_string(data)
+    with open(path, "w") as f:
+        f.write(string)
+
+
+def json_load(path):
+    with open(path, "r") as f:
+        data = f.read()
+        return json.loads(data, object_pairs_hook=OrderedDict)
+
+
+def file_hash(path):
+    if not os.path.isfile(path):
+        raise ValueError
+
+    hasher = hashlib.sha256()
+    block_size = 4096
+    with open(path, 'rb') as f:
+        buf = f.read(block_size)
+        while len(buf) > 0:
+            hasher.update(buf)
+            buf = f.read(block_size)
+    return hasher.hexdigest()
+
+
+def get_file_paths(root_path):
+    file_paths = []
+    for subdir, dirs, files in os.walk(root_path):
+        for f in files:
+            path = os.path.join(subdir, f)
+            if ".git/" in path:
+                continue
+            file_paths.append(path)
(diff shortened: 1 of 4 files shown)


contrib/policy_upgrade/generate_hashes.py
```diff
@@ -0,0 +1,102 @@
...
+            buf = f.read(block_size)
+    return hasher.hexdigest()
+
+
+def get_file_paths(root_path):
+    file_paths = []
+    for subdir, dirs, files in os.walk(root_path):
+        for f in files:
+            path = os.path.join(subdir, f)
+            if ".git/" in path:
```
mikeweilgart: This doesn't look right.  How will it handle a directory called `something.git`?

contrib/policy_upgrade/generate_hashes.py
```diff
@@ -0,0 +1,102 @@
...
+def generate_db(root_path, version):
+    db = OrderedDict()
+    file_paths = get_file_paths(root_path)
+
+    for path in file_paths:
+        digest = file_hash(path)
+        if digest not in db:
+            db[digest] = []
+        if path not in db:
+            db[path] = []
```
mikeweilgart: This looks like it will break if there are files in the directory whose names are hashes...but on second look, I guess the paths will always contain slashes, and the hashes never well; is that the intention?

contrib/policy_upgrade/scan_mpf.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def json_print(data):
+    print(json_string(data))
+
+
+def json_save(data, path):
+    string = json_string(data)
+    with open(path, "w") as f:
+        f.write(string)
```
mikeweilgart: This is a lot of redundant function definitions.  All of these same functions are also defined in generate_hashes.py.  Violates SPOT (single point of truth).

olehermanse: @mikeweilgart this is a WIP / draft. A code review is not that helpful at this time.

mikeweilgart: Okay, that makes sense.  :)  Do you have any outline on the overall design you're planning to implement here?

(In the collaborative workflow I use, we write a design and then get a design review before we start implementation, so if there is anything like that I would be very interested to read the design doc.)

olehermanse: @mikeweilgart There are some JIRA tickets already:

https://tracker.mender.io/browse/CFE-3121
https://tracker.mender.io/browse/CFE-3122
https://tracker.mender.io/browse/CFE-2991

However, they are mostly just placeholders, for now. We do have your comments from the previous PR:

https://github.com/cfengine/masterfiles/pull/1058

I will pick this up after 3.15 is released. I'll make sure to ping you whenever we have a prototype / design / proposal where it makes sense to ask for feedback.

olehermanse:
Closing this due to inactivity. The planned work is still tracked in JIRA:

https://tracker.mender.io/browse/CFE-3121
https://tracker.mender.io/browse/CFE-3122
https://tracker.mender.io/browse/CFE-2991

PR #1330, opened 2019-03-05
Services do not detect state on FreeBSD

xaxo05:
Services do not detect state on FreeBSD due to wrong path to init scripts when using the default sysvservice facility. Replace /etc/rc.d/$(service) with /usr/local/etc/rc.d/$(service) or /etc/rc.d/$(service) if the first does not exist.

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -170,12 +170,14 @@ bundle agent standard_services(service,state)
       "call_systemctl" string => "$(paths.systemctl) --no-ask-password --global --system";
       "systemd_properties" string => "-pLoadState,CanStop,UnitFileState,ActiveState,LoadState,CanStart,CanReload";
       "c_service" string => canonify("$(service)");
+
     freebsd::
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
@BasementTrix can you please review?

nickanderson:
Or maybe @skreuzer ?

skreuzer:
@nickanderson 3rd party software installed via the ports system gets installed to /usr/local by default on FreeBSD. First checking to see if the rc script to start and stop the daemon is in /usr/local and defaulting to /etc/rc.d if it does not seems reasonable

nickanderson:
Thanks for the input @skreuzer. @BasementTrix said that he was also going to take a look, so I will wait a bit longer to give him more opportunity to review.

nickanderson:
Sorry for the delay here, I didn't intend for this to sit so long.

So that we have proper changelog, I am taking these in a different PR: https://github.com/cfengine/masterfiles/pull/1531

PR #1331, opened 2019-03-06
ENT-4041 ubuntu-18

Lex-2008:
Merge together:
https://github.com/cfengine/buildscripts/pull/462
https://github.com/cfengine/nova/pull/1371

### DIFF
--- a/tests/acceptance/17_packages/01_init/unsafe/timed/001-prepare-repositories.cf
+++ b/tests/acceptance/17_packages/01_init/unsafe/timed/001-prepare-repositories.cf
@@ -122,15 +122,20 @@ bundle agent signing_keys
         expression => "any",
         scope => "namespace";
 
-  files:
+  vars:
     debian::
-      "$(G.testdir)/."
-        create => "true";
+      "gpg" string => "gpg --homedir .gnupg-temp --batch --no-tty";
 
   commands:
     debian::
-      "gpg --homedir $(G.testdir) --no-default-keyring --keyring $(p.resources)/gpg/pubring.gpg --export \
-       | gpg --homedir $(G.testdir) --no-default-keyring --primary-keyring /etc/apt/trusted.gpg --import"
+      # re-sign repo
+      "mkdir -m 700 .gnupg-temp; \
+       $(gpg) --import $(p.resources)/gpg/secring.gpg; \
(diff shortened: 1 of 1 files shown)


cf-bottom:

tests/acceptance/17_packages/01_init/unsafe/timed/001-prepare-repositories.cf
```diff
@@ -122,15 +122,20 @@ bundle agent signing_keys
...
+      # re-sign repo
+      "mkdir -m 700 .gnupg-temp; \
+       $(gpg) --import $(p.resources)/gpg/secring.gpg; \
+       rm /test-repos/deb_repo1/dists/package1/Release.gpg /test-repos/deb_repo2/dists/package2/Release.gpg; \
+       $(gpg) --detach-sign --armor --output /test-repos/deb_repo1/dists/package1/Release.gpg /test-repos/deb_repo1/dists/package1/Release; \
+       $(gpg) --detach-sign --armor --output /test-repos/deb_repo2/dists/package2/Release.gpg /test-repos/deb_repo2/dists/package2/Release; \
+       rm -rf .gnupg-temp; \
+       apt-key add $(p.resources)/gpg/pubring.gpg"
         contain => useshell,
         classes => if_successful("signing_keys_ok");
```
nickanderson: Beware: This may not do what you wish. There are MANY commands being run in that one promise, which return code will the agent interpret for the result? (I don't know) Also, because of how promises work, and they can have multiple outcomes, Kept|Repaired does not necessarily mean that any part of the promise was NOTKEPT. I tend to reccomend using the results classes body which defines a class suffixed with the result for each result the promise produces, that way you can later check based on a full expression like  `(signing_keys_kept|signing_keys_repaired).!signing_keys_notkept::` (Signing keys kept OR repaired, and NOT notkept).

Lex-2008: Right, I'll replace `;` with `&&` and see if all commands return 0, as expected.

vpodzime: Hmm, I wonder if we could actually drop the backslashes.

PR #1335, opened 2019-03-18
ENT-4494 Improvements to federation policy

nickanderson:


cf-bottom:

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -11,12 +15,12 @@ bundle agent federation_config
   classes:
     enterprise_edition.(policy_server|am_policy_hub)::
       "cfengine_internal_federation_enabled"
-        expression => fileexists("$(path)"),
+        expression => fileexists( $(path) ),
```
craigcomstock: @nickanderson is there a style guide for policy somewhere that says to have spaces here? Mostly just curious and want to read/bookmark said style guide. :)

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -81,69 +85,66 @@ bundle agent federation_manage_cfdrop
     "$(create_files)"
       create => "true";
     "$(home)/."
-      depth_search => recurse_with_base("inf"),
-      file_select => all,
-      perms => mo("600", "$(user)");
+      depth_search => default:recurse_with_base("inf"),
+      file_select => default:all,
```
craigcomstock: default: is needed because of namespacing this bundle? It doesn't degrade/search-up to default if not found in the current namespace? ;) #design #todo?

cfe_internal/enterprise/federation/federation.cf
nickanderson: https://docs.cfengine.com/docs/3.12/guide-writing-and-serving-policy-policy-style.html

Looks like this style suggestion is missing from the guide. Avoid quoting variables if you can is advice I received from developers long ago because the parser can pick up on errors more easily.

cfe_internal/enterprise/federation/federation.cf
nickanderson: Yes, if you're in a namespace, then it looks in your namespace for the body. If you want to use a body from another namespace, you must do so explicitly.

cfe_internal/enterprise/federation/federation.cf
olehermanse: I still don't quite understand where I have to use quotes and not.

nickanderson: secret knowledge

nickanderson:
Squashed. Anyone think it needs squashed further?

olehermanse:
LGTM

PR #1339, opened 2019-03-19
CFE-3022: redhat_pure is no longer defined on Fedora hosts

nickanderson:
Changelog: Title

The agent defines the redhat class on fedora hosts. Fedora isn't really a
derivative of redhat, it's the other way around. But Fedora is definitely not
pure redhat.

(cherry picked from commit b72953241bd91c3f2226379a3fd267d3a33e8fd7)


cf-bottom:

PR #1340, opened 2019-03-20
Add scripts and templates for Federated Reporting

vpodzime:
The configurable parts are only in the config.sh.mustache and
log.sh.mustache files, the other files just use the configuration
rendered by the policy with those two templates.

Ticket: ENT-4473
Changelog: Added scripts and templates for Federated Reporting

### DIFF
--- a/templates/federated_reporting/parallel.sh
+++ b/templates/federated_reporting/parallel.sh
@@ -1,0 +1,127 @@
+# A shell "library" providing a function to run a given command on multiple
+# arguments in parallel.
+#
+# For best results, make sure 'parallel' is installed.
+#
+
+if type parallel >/dev/null; then
+  HAVE_PARALLEL=1
+else
+  HAVE_PARALLEL=0
+fi
+
+if type xargs >/dev/null; then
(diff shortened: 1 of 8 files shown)


cf-bottom:

templates/federated_reporting/parallel.sh
```diff
@@ -0,0 +1,127 @@
...
+      job_slots="$(expr $job_slots + 1)"
+    fi
+    "$1" "$item" &
+    job_slots="$(expr $job_slots - 1)"
+  done < $input_arg
+  wait
+  return $failure
+}
+
+run_in_parallel() {
```
nickanderson: SHoudl this be called try_run_in_parallel since it falls back to a serial for loop in the end? really nitpicky, hard to find things to complain about.

olehermanse: the for loop is using backgrounding `&` so it's still parallell. :)

nickanderson: OK fine

PR #1350, opened 2019-03-22
ENT-4317 Fix standalone_self_upgrade not triggering because of stale data (3.10.x)

nickanderson:
The desired version state data is written by standalone_self_upgrade.cf, which
is triggered when trigger_upgrade is defined and the running agent version does
not match the desired version.

The desired version defaults to the version of the MPF when it was built and can
be overridden via augments.

Once the agent was at the desired version, the standalone_self_upgrade.cf policy
would never trigger, because the running version would match the desired version
data. Since it would not run, it would not update the desired version data.

This change purges the desired version state data when it is older than the
standalone_self_upgrade.cf. This way, when the policy changes there will not be
stale data preventing upgrade.

Changelog: Title
(cherry picked from commit c3343cfb0cc3921c7641974c0b00abf61b0acfe2)

### DIFF
--- a/update.cf
+++ b/update.cf
@@ -56,9 +56,38 @@ bundle common cfengine_update_controls
 }
 
 bundle agent cfengine_internal_standalone_self_upgrade
+# @brief Manage desired version state and execution of policy to reach the target version.
+{
+  methods:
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #1353, opened 2019-03-25
Handle yum exit_on_lock with centos

nickanderson:


nickanderson:
@mikeweilgart thanks for pointing this out, for your review.

lib/packages.cf
```diff
@@ -271,9 +271,9 @@ bundle common rpm_knowledge
 bundle common redhat_no_locking_knowledge {
   vars:
     # Option was introduced in rhel 6.1
-    redhat.!(redhat_3|redhat_4|redhat_5|redhat_6_0)::
+    (redhat|centos).!(redhat_3|redhat_4|redhat_5|redhat_6_0|centos_3|centos_4|centos_5|centos_6_0)::
```
mikeweilgart: The `centos` class isn't needed - the line should be:

```
redhat.!(redhat_3|redhat_4|redhat_5|redhat_6_0|centos_3|centos_4|centos_5|centos_6_0)::
```

mikeweilgart: (It won't break anything to have the `centos` class there, though.)

nickanderson: Right, I added centos there so that in the event redhat is not defined on centos, it still does the right thing in this very specific case. Probably many other things won't work as expected, but this would.

PR #1354, opened 2019-03-26
ENT-4551 Ensure that asynchronous query API semaphores are writable

nickanderson:
Changelog: Title

This change ensures that the files used by the asynchronous query API are
writable by the web server so that the status can be correctly updated. Without
being writable scheduled reports may hang and never complete.


cf-bottom:

PR #1357, opened 2019-03-27
ENT-4494: Render federation scripts

nickanderson:


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -133,18 +133,51 @@ bundle agent federation_manage_files
...
+        perms => default:mog( "600", "root", "root" );
+
+      "$(cfengine_enterprise_federation:config.bin_dir)/parallel.sh"
+        copy_from => local_dcp( "$(this.promise_dirname)/../../../templates/federated_reporting/parallel.sh" ),
         perms => default:mog( "600", "root", "root" );
+
+    am_feeder::
+      "$(cfengine_enterprise_federation:config.bin_dir)/dump.sh"
+        copy_from => local_dcp( "$(this.promise_dirname)/../../../templates/dump.sh" ),
+        perms => default:mog( "700", "root", "root" );
```
craigcomstock: templates/dump.sh -> templates/federated_reporting/dump.sh

maybe make a var for the templates/federated_reporting path? a lot of stuff in all these could get typos in them.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -133,18 +133,51 @@ bundle agent federation_manage_files
...
+
+    am_transporter::
+      "$(cfengine_enterprise_federation:config.bin_dir)/transport.sh"
+        copy_from => local_dcp( "$(this.promise_dirname)/../../../templates/transport.sh" ),
+        perms => default:mog( "700", "root", "root" );
```
craigcomstock: templates/ -> templates/federated_reporting

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -133,18 +133,51 @@ bundle agent federation_manage_files
...
+
     am_superhub::
       "$(cfengine_enterprise_federation:config.bin_dir)/import.sh"
-        create => "true",
+        copy_from => local_dcp( "$(this.promise_dirname)/../../../templates/import.sh" ),
+        perms => default:mog( "700", "root", "root" );
```
craigcomstock: templates/ -> templates/federated_reporting

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -133,18 +133,51 @@ bundle agent federation_manage_files
...
+
+      "$(cfengine_enterprise_federation:config.bin_dir)/import_file.sh"
+        copy_from => local_dcp( "$(this.promise_dirname)/../../../templates/import_file.sh" ),
+        perms => default:mog( "700", "root", "root" );
```
craigcomstock: templates/ -> templates/federated_reporting

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -133,18 +133,51 @@ bundle agent federation_manage_files
...
+
+      "$(cfengine_enterprise_federation:config.bin_dir)/filter.sed"
+        copy_from => local_dcp( "$(this.promise_dirname)/../../../templates/filter.sed" ),
         perms => default:mog( "600", "root", "root" );
+
```
craigcomstock: templates/ -> templates/federated_reporting

craigcomstock: @nickanderson @vpodzime This doesn't seem quite right. The files copied over by `cf-agent -KIf update.cf` don't include `*.sed` I'm pretty sure. Reading update_policy.cf and update_def.cf which has "input_name_patterns" which doesn't include `*.sed`.
Also the config.sh and import_file.sh scripts seem to expect multiple sed filter files in /opt/cfengine/federation/superhub/import/filters directory but I don't see anything that would copy `*.sed` to that location in the policy.

cfe_internal/enterprise/federation/federation.cf
nickanderson: ack

cfe_internal/enterprise/federation/federation.cf
nickanderson: ack on path

will consider DRY

cfe_internal/enterprise/federation/federation.cf
nickanderson: ack

cfe_internal/enterprise/federation/federation.cf
nickanderson: ack

cfe_internal/enterprise/federation/federation.cf
nickanderson: ack

nickanderson: You're right. add to input name patterns or rename to .txt and copy to .sed?

craigcomstock: I am working on a fix for the import/filters path creation and copying *.sed over to there.

craigcomstock: I don't think *.sed makes sense for i2 so think just one filter.sed is fine for now. Easier in policy anyway.

nickanderson: Yes. Also, if a user wants to install sed files, they can add their own promises in their own bundles and files.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -133,18 +133,53 @@ bundle agent federation_manage_files
...
-    am_transporter::
-      "$(cfengine_enterprise_federation:config.bin_dir)/transport.sh"
+
+      # TODO: Instrument augments
+      "$(cfengine_enterprise_federation:config.bin_dir)/log.sh"
         create => "true",
+        template_method => "mustache",
+        edit_template => "$(this.promise_dirname)/../../../templates/federated_reporting/log.sh.mustache",
+        template_data => '{ "logger": "log", "logger_args": "", "timestamp_cmd": "date --iso-8601=seconds"}',
+        perms => default:mog( "600", "root", "root" );
```
craigcomstock: configuring "logger": "log" will cause an infinite loop in log.sh :) Since log() is the function that it defines and then it calls ${CFE_LOGGER}.

```
function log() {
  # $1 -- message to log
  ts="$($CFE_TIMESTAMP_CMD)"
  if [ -n "$ts" ]; then
    "${CFE_LOGGER}" "[$ts] $1"
  else
    "${CFE_LOGGER}" "$1"
  fi
}
```

So likely we want to leave this blank. Remove the "logger" data entry. Might as well remove timestamp_cmd also since what you have there is the default in log.sh.

nickanderson:
Rebased after changes to transport script to support pull

cfe_internal/enterprise/federation/federation.cf
nickanderson: Yeah ok i won't provide ANY template data, that was just an artifact from me starting to look at what should be supported from def.json, but there are so many things that i think we should ignore it for now. Will redact template_data momentarily

PR #1359, opened 2019-03-27
ENT-4551 Ensure that asynchronous query API semaphores are writable (3.10.x)

nickanderson:
Changelog: Title

This change ensures that the files used by the asynchronous query API are
writable by the web server, so that the status can be correctly updated. Without
being writable scheduled reports may hang and never complete.

(cherry picked from commit 446b54e46f9dd804a8b5329eacb361009b568e1b)


cf-bottom:

PR #1361, opened 2019-04-08
Add support for pull to FR/transport.sh

vpodzime:
Making the superhub pull the dump files from the feeder hubs is a
better strategy than pushing the files from the feeder hubs to
the superhub. The superhub knows better when it has a capacity to
process the files and also, there's only one SSH private key that
has to be kept safe -- the superhub's key.

However, in some situations the push approach may be necessary so
let's not just delete the functionality. The transport.sh script
can easily support both modes.

Ticket: ENT-4499
Changelog: None

### DIFF
--- a/templates/federated_reporting/pull_dumps_from.sh
+++ b/templates/federated_reporting/pull_dumps_from.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# A script to pull dumps from a given feeder hub to the $PWD/$feeder folder.
+#
+# $1 -- feeder hub hostname/IP to pull the dumps from
+#
+
+set -e
+
+# make sure a failure in any part of a pipe sequence is a failure
+set -o pipefail
+
+source "$(dirname "$0")/config.sh"
+
+# check that we have all the variables we need
+true "${CFE_FR_TRANSPORT_DIR?undefined}"
+true "${CFE_FR_SUPERHUB_DROP_DIR?undefined}"
+true "${CFE_FR_TRANSPORTER?undefined}"
+true "${CFE_FR_TRANSPORTER_ARGS?undefined}"
+true "${CFE_FR_SSH?undefined}"
+true "${CFE_FR_SSH_ARGS?undefined}"
+true "${CFE_FR_COMPRESSOR_EXT?undefined}"
+true "${CFE_FR_FEEDER_USERNAME?undefined}"
+
+if [ $# != 1 ]; then
+  log "Invalid number of arguments ($#) given to $0, exiting!"
+  exit 1
+fi
+
+feeder="$1"
+
+mkdir -p "$feeder"
+
+# move the files so that they don't get overwritten/deleted during the transport
+"$CFE_FR_SSH" $CFE_FR_SSH_ARGS "$CFE_FR_FEEDER_USERNAME@$feeder" "mkdir $CFE_FR_TRANSPORT_DIR/$$.transporting"
+"$CFE_FR_SSH" $CFE_FR_SSH_ARGS "$CFE_FR_FEEDER_USERNAME@$feeder" "mv $CFE_FR_TRANSPORT_DIR/*.sql.$CFE_FR_COMPRESSOR_EXT $CFE_FR_TRANSPORT_DIR/$$.transporting/"
+
+failed=0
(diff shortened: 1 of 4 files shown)


cf-bottom:

templates/federated_reporting/transport.sh
```diff
@@ -21,28 +33,108 @@ if ! type "$CFE_FR_TRANSPORTER" >/dev/null; then
...
+      log "No files to transport."
+      exit 0
+    }
+
+  log "Transporting files: $dump_files"
+  some_failed=0
+  for dump_file in $dump_files; do
+    failed=0
+    mv "$dump_file" "$dump_file.transporting"
+    "$CFE_FR_TRANSPORTER" "$CFE_FR_TRANSPORTER_ARGS" "$dump_file.transporting" "$CFE_FR_SUPERHUB_LOGIN:$CFE_FR_SUPERHUB_DROP_DIR/$(basename "$dump_file")" &&
```
craigcomstock: in case of failure should we not `mv` the files but rather `cp` them and then delete them if not failed?

templates/federated_reporting/transport.sh
```diff
@@ -21,28 +33,108 @@ if ! type "$CFE_FR_TRANSPORTER" >/dev/null; then
...
+  fi
+}
+
+function pull() {
+  # $@ -- feeder hubs to pull the dumps from
+  feeder_lines="$(printf "%s\n" "$@")"
+  log "Pulling dumps from: $feeder_lines"
+
+  chmod u+x "$(dirname "$0")/pull_dumps_from.sh"
+
```
craigcomstock: shouldn't perms like this be managed by policy? seems like this is a good safety measure but I wonder if it is noise.

templates/federated_reporting/pull_dumps_from.sh
```diff
@@ -0,0 +1,48 @@
...
+true "${CFE_FR_TRANSPORTER?undefined}"
+true "${CFE_FR_TRANSPORTER_ARGS?undefined}"
+true "${CFE_FR_SSH?undefined}"
+true "${CFE_FR_SSH_ARGS?undefined}"
+true "${CFE_FR_COMPRESSOR_EXT?undefined}"
+true "${CFE_FR_FEEDER_USERNAME?undefined}"
+
+if [ $# != 1 ]; then
+  log "Invalid number of arguments ($#) given to $0, exitting!"
+  exit 1
```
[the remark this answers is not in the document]
vpodzime: Good catch, fixing.

templates/federated_reporting/transport.sh
vpodzime: So this is a generic question about what should happen when something fails. I talked to @olehermanse about this and we agreed that primarily we should make sure things are not piling up. So what I think should actually be done here is that the file should be removed after the transport attempt in any case. But let's not change that in this commit or PR because that's just moving the "old" push code, not modifying it.

templates/federated_reporting/transport.sh
vpodzime: I was wondering about the same. But then I thought this won't hurt. And it happened to me while testing that the file was not executable (copied over by `cf-remote`).

templates/federated_reporting/transport.sh
```diff
@@ -1,4 +1,15 @@
...
+#
+# Transport dump files from the feeder hubs to the superhub.
+#
+# Can be run as:
+#   transport.sh
+#         On a feeder hub, pushes dump files to the superhub.
+#   transport.sh push
+#         The same as with no arguments.
+#   transport.sh pull FEEDER_HUB [FEEDER_HUB2...FEEDER_HUBn]
+#         On the superhub, pull dumps from the given feeder hubs (in parallel).
```
olehermanse: @vpodzime @nickanderson I would avoid using the word dump here.

nickanderson: - s/dumps/exported data/
- s/dumps/excretions/
- s/dumps/discharge/
?

vpodzime: Not deeply concerned, to be honest, this is not really supposed to be run manually by people who don't know what a dump is. Besides, `pg_dump` is the tool that creates those files.

PR #1362, opened 2019-04-08
Use glob-matching to promise permissions on installation logs

vpodzime:
The logs may be in the /var/log/CFEngineHub-Install.log file or
/var/log/CFEngine-Install.log or both (one from old installation,
another from a new installation/upgrade).

Ticket: ENT-4564

### DIFF
--- a/cfe_internal/enterprise/CFE_knowledge.cf
+++ b/cfe_internal/enterprise/CFE_knowledge.cf
@@ -30,7 +30,11 @@ bundle agent cfe_internal_setup_knowledge
         or => {
                 "enterprise_3_7_3", "enterprise_3_7_4", "enterprise_3_7_5",
                 "enterprise_3_10_0"
-              };
+        };
+
+  vars:
+      "install_logs" -> {"ENT-4564"}
+        slist => findfiles("/var/log/CFEngine*Install.log");
 
   files:
 
@@ -43,7 +47,7 @@ bundle agent cfe_internal_setup_knowledge
 
     any::
 
-      "/var/log/CFEngineHub-Install.log" -> { "ENT-4506" }
+      "$(install_logs)" -> { "ENT-4506" }
         perms => mog("0600", "root", "root" );
 
       "$(cfe_internal_hub_vars.docroot)"


cf-bottom:

cfe_internal/enterprise/CFE_knowledge.cf
```diff
@@ -30,7 +30,11 @@ bundle agent cfe_internal_setup_knowledge
         or => {
                 "enterprise_3_7_3", "enterprise_3_7_4", "enterprise_3_7_5",
                 "enterprise_3_10_0"
-              };
+        };
+
+  vars:
+      "install_logs" -> {"ENT-4564"}
+        slist => findfiles("/var/log/CFEngine*Install.log");
```
nickanderson: Alternatively, instead of having an intermediary variable, you could use a regex in the promiser.

cfe_internal/enterprise/CFE_knowledge.cf
```diff
@@ -43,7 +47,7 @@ bundle agent cfe_internal_setup_knowledge
 
     any::
 
-      "/var/log/CFEngineHub-Install.log" -> { "ENT-4506" }
+      "$(install_logs)" -> { "ENT-4506" }
```
nickanderson: What about:

```
"/var/log/CFEngine.*\.log"
  perms => mog( "0600", "root", "root" );
```

cfe_internal/enterprise/CFE_knowledge.cf
vpodzime: I don't know, I like that it explicitly acts on files that exist. And globs are easier to read and more come in combination with file paths. YMMV.

nickanderson: Indeed, that's the advantage of the intermediary variable. You are documenting a specific intent.
Cool, let's keep it the way it is. However, this should be a different path on windows.

vpodzime: >  Cool, let's keep it the way it is. However, this should be a different path on windows.

AFAICT, I'm not changing anything here regarding windows. Also, I don't think we have such an installation log on windows in the first place. So do you think we should just move this promise under a `!windows::` guard?

nickanderson:
Thanks, superseded by https://github.com/cfengine/masterfiles/pull/1366 which includes your commit.

PR #1367, opened 2019-04-11
ENT-4576: remote config.sh args, add login value, configure authorized_keys 

craigcomstock:


craigcomstock:
Will work on the two TODOs tomorrow if someone else doesn't take them up. :) ssh_* on superhub and feeder hubs.

nickanderson:
I hope you don't mind, I pushed directly into your branch a promise to extract login.

How many ssh_publkeys do we expect will be needed? Just one like login, or search for all and render unified set?

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -88,6 +102,12 @@ bundle agent transport_user
       depth_search => default:recurse_with_base("inf"),
       file_select => default:all,
       perms => default:mog( "600", $(user), "root" );
+    "$(home)/.ssh/authorized_keys"
+      create => "true",
+      edit_template_string => "{{#-top-}}{{{.}}}$(const.n){{/-top-}}",
+      template_data => @(pubkeys),
```
nickanderson: `pubkeys` is defined in `bundle agent config`, so I believe it needs to either be passed to this `bundle agent transport_user` as a parameter, or it needs to be referenced using it's fully qualified name `config.pubkeys` or `cfengine_enterprise_federation:config.pubkeys`.

craigcomstock:
@nickanderson could you give this a look and merge? I'm going to work on known_hosts a bit as a separate ticket/PR.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -88,9 +119,22 @@ bundle agent transport_user
       depth_search => default:recurse_with_base("inf"),
       file_select => default:all,
       perms => default:mog( "600", $(user), "root" );
+    "$(home)/.ssh/authorized_keys"
```
nickanderson: $(ssh_auth_keys) here instead of $(home)/.ssh/authorized_keys

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -88,9 +119,22 @@ bundle agent transport_user
...
+      create => "true",
+      edit_template_string => "{{#-top-}}{{{.}}}$(const.n){{/-top-}}",
+      template_data => @(cfengine_enterprise_federation:config.pubkeys),
+      template_method => "inline_mustache";
+
   commands:
+    # Generate a modern ssh keypair (2019-04)
+    "/usr/bin/ssh-keygen"
+      handle => "ssh_keys_configured",
+      args => "-N '' -o -a 150 -t $(ssh_key_type) -f $(ssh_priv_key)",
```
nickanderson: Don't we need to drop `-o`? or was that not problematic for one env?

craigcomstock: Wasn't a problem in my testing with ubuntu 16. But you are right, depending on version it may be a problem. Is -o required for the ssh_key_type you chose?

cfe_internal/enterprise/federation/federation.cf
craigcomstock: ack.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -187,12 +232,19 @@ bundle agent federation_manage_files
...
     am_superhub::
       "/var/cfengine/state/pg/data/postgresql.conf"
-        edit_line => default:insert_lines("shared_buffers=1GB");
-      "/var/cfengine/state/pg/data/postgresql.conf"
-        edit_line => default:insert_lines("max_locks_per_transaction=4000");
+        edit_line => default:set_line_based( "$(this.bundle).c",
+                                     "=",
+                                     "\s*=\s*",
+                                     ".*",
+                                     "");
```
craigcomstock: @nickanderson it might be better if I don't have this postgres change in here since you have it in your PR: https://github.com/cfengine/masterfiles/pull/1369

cfe_internal/enterprise/federation/federation.cf
nickanderson:
```
     -o	     Causes ssh-keygen to save private keys using the new OpenSSH for-
	     mat rather	than the more compatible PEM format.  The new format
	     has increased resistance to brute-force password cracking but is
	     not supported by versions of OpenSSH prior	to 6.5.	 Ed25519 keys
	     always use	the new	private	key format.
```

So, no, sicne i am using ed25519  `-o` shouldn't be requred

cfe_internal/enterprise/federation/federation.cf
nickanderson: It doesn't matter, i will rebase

PR #1369, opened 2019-04-18
Tweaks for federation policy

nickanderson:


cf-bottom:

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -31,10 +31,10 @@ bundle agent config
         comment => "Has a remote hub with superhub role, should act as feeder",
         scope => "namespace";
 
+      # Knowing if your a superhub is based on explicit setting of role in path
+      # (federation-config.json)
```
craigcomstock: s/your/you're/

nickanderson: ACK, thanks. DONE.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -89,8 +103,16 @@ bundle agent transport_user
       file_select => default:all,
       perms => default:mog( "600", $(user), "root" );
   commands:
+
+      # Generate a modern ssh keypair (2019-04)
+      "/usr/bin/ssh-keygen"
+        args => "-N '' -o -a 150 -t $(ssh_key_type) -f $(ssh_priv_key)",
+        contain => setuid( $(user) ),
```
craigcomstock: odd, my debian ssh-keygen (ssh-V = OpenSSH_7.9p1, OpenSSL 1.1.1b 26 Feb 2019) doesn't have -o option, but ubuntu xenial 16.04 has OpenSSL 1.0.2g DOES have the option.

I don't see little o -o option at https://man.openbsd.org/ssh-keygen.

I think according to the 1.0.2g man page -o is not needed if you are using Ed25519 type?

>      -o      Causes ssh-keygen to save private keys using the new OpenSSH format rather than the more compati‐
             ble PEM format.  The new format has increased resistance to brute-force password cracking but is
             not supported by versions of OpenSSH prior to 6.5.  Ed25519 keys always use the new private key
             format.

Couldn't really find a mention of "format" that seemed similar in the newer man page. FWIW.

nickanderson: We can strip out the -o if it's not needed. I knew that by going with ed25519 i was asking or some kind of issue.

nickanderson: not that the -o is needed for that, just one of the things found while i was doing some checks.

craigcomstock:
I think I've pulled everything from here into https://github.com/cfengine/masterfiles/pull/1370 so maybe close this PR?

PR #1370, opened 2019-04-22
Ent 4587 - ssh fingerprints and run transport.sh as cfdrop

craigcomstock:


craigcomstock:
@nickanderson can you review this policy now before I squash commits? I have tested this policy and was able to get dumps from feeder to hub and into superhub cfdb database. :)

craigcomstock:
squashing after all, the various commits had no value. :)

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -231,26 +250,43 @@ bundle agent postgres_config
 {
   vars:
     am_superhub::
-      "c[shared_buffers]" string => "1GB";
-      "c[max_locks_per_transaction]" string => "4000";
-
+      "c[shared_buffers]"
```
nickanderson: Do we have any issues related to these config value settings? Seems some promisees might ge good for reference.

This is an ideal spot to hook in for automatically configuring postgres based on the information we have about how it's functioning/hardware available.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -113,14 +121,22 @@ bundle agent transport_user
     "$(create_files)"
       create => "true";
     "$(home)/."
+      depends_on => { "ssh_auth_keys_configured", "ssh_known_hosts_configured" },
       depth_search => default:recurse_with_base("inf"),
```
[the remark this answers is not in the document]
craigcomstock: Hmm. I'm pretty sure in testing that permissions weren't being applied "quick enough" and with these deps things happen in "one shot".

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -139,11 +155,17 @@ bundle agent transport_user
...
+      string => "$(cfengine_enterprise_federation:config.transport_user)";
+    "login" data => parsejson('{"login":"$(cfengine_enterprise_federation:config.login)"}');
+    "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
+
   files:
     enterprise_edition.(policy_server|am_policy_hub)::
       "$(cfengine_enterprise_federation:config.federation_dir)/."
         create => "true",
-        perms => default:mog( "701", "root", "root" );
+        perms => default:mog( "761", "root", "$(transport_user)" );
```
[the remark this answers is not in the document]
craigcomstock: transport_user needs "mv" permissions (directory rw-) in order to move the dump file over before doing rsync.

cfe_internal/enterprise/federation/federation.cf
craigcomstock: created one: https://tracker.mender.io/browse/ENT-4614

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -260,7 +260,7 @@ bundle agent postgres_config
                                      "\s*=\s*",
                                      ".*",
                                      ""),
-        classes => results( "bundle", "postgresql_conf" );
+        classes => default:results( "bundle", "postgresql_conf" );
```
nickanderson: Good.

cfe_internal/enterprise/federation/federation.cf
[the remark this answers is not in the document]
nickanderson: why other 1?

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -139,11 +148,19 @@ bundle agent transport_user
...
+  vars:
+    "transport_user"
+      string => "$(cfengine_enterprise_federation:config.transport_user)";
+    "login" data => parsejson('{"login":"$(cfengine_enterprise_federation:config.login)"}');
+    "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
+
   files:
     enterprise_edition.(policy_server|am_policy_hub)::
+      # $(transport_user) needs permissions from / up to it's homedir
+      # cfapache needs to get into this dir, so o+x is specified here
```
nickanderson: alrighty

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -154,11 +171,12 @@ bundle agent federation_manage_files
...
       "$(cfengine_enterprise_federation:config.bin_dir)/."
         create => "true",
-        perms => default:mog( "600", "root", "root" );
+        perms => default:mog( "660", "root", "$(transport_user)" );
     am_superhub::
       "$(cfengine_enterprise_federation:config.federation_dir)/superhub/."
         create => "true",
-        perms => default:mog( "701", "root", "root" );
+# TODO why did this have 701??? change and test.
+        perms => default:mog( "770", "root", "$(transport_user)" );
```
nickanderson: It's fine to have this as 77, it's explicitly what you want, just remember that by default, any time you set read bit, you get execute bit. So, above for bin_dir, 660, will actually be 770. This may seem super weird, but it's super duper useful when you apply for permissions recursively. With one promise, you can usually get what you want on both files and directory instead of having to do recurseive dir promise, then seperately a recursive file promise (file_selct)

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -231,26 +247,43 @@ bundle agent postgres_config
...
                                      "\s*=\s*",
                                      ".*",
-                                     "");
+                                     ""),
+        classes => default:results( "bundle", "postgresql_conf" );
+
+  commands:
+    am_superhub.postgresql_conf_repaired::
+      "$(sys.bindir)/pg_ctl"
+        arglist => {
```
nickanderson: very nicely documented.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -260,10 +293,17 @@ bundle agent data_transport
...
     am_transporter::
-      # TODO: pass correct arguments once script is done
-      "bash"
+      "/bin/bash"
         arglist => {"$(cfengine_enterprise_federation:config.bin_dir)/transport.sh"},
-        contain => default:in_shell;
+        contain => contain_transport_user;
+}
+
+body contain contain_transport_user
```
nickanderson: Getting fancy now I see with custom bodies!

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -291,29 +331,58 @@ bundle agent entry
...
       "CFEngine Enterprise Federation Postgres Configuration"
+        handle => "postgres_config",
         usebundle => postgres_config;
       "CFEngine Enterprise Federation Feeder Data Export"
         usebundle => exported_data;
       "CFEngine Enterprise Federation Feeder Data Transport"
         usebundle => data_transport;
       "CFEngine Enterprise Federation Feeder Data Import"
         usebundle => imported_data;
+      "Update Status"
```
nickanderson: I might call this Configuration Status instead of "Update Status"

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -291,29 +331,58 @@ bundle agent entry
...
+      };
+
+  files:
+    superhub_setup_status_complete::
+      "$(cfengine_enterprise_federation:config.federation_dir)/cfapache/setup-status.json"
+        create => "true",
+        template_method => "inline_mustache",
+        edit_template_string => "{{%-top-}}",
+        template_data => '{ "configured": true, "transport_ssh_public_key": "$(with)" }',
+        with => readfile( "$(cfengine_enterprise_federation:transport_user.ssh_pub_key)" ),
```
nickanderson: whoa, leveraging `with` too I see. I think that this will need to be placed inside of an if minimum version macro because with isn't valid on 3.10.x, so I would expect this to fail the backward compatibility syntax check.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -231,26 +247,43 @@ bundle agent postgres_config
...
+      "$(sys.bindir)/pg_ctl"
+        arglist => {
+                     "restart",
+                     "-w", # Wait for startup or shutdown to complete, -W (do not wait) is the default for start and restart modes
+                     "-D",
+                     "$(sys.statedir)/pg/data",
+                     "-l",
+                     "/var/log/postgresql.log",
+                     "-m",
+                     "fast" # All active transactions rolled back and clients forcibly disconnected
```
vpodzime: Shouldn't this do `smart` instead?

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -154,11 +171,12 @@ bundle agent federation_manage_files
     enabled::
       "$(cfengine_enterprise_federation:config.bin_dir)/."
         create => "true",
-        perms => default:mog( "600", "root", "root" );
+        perms => default:mog( "660", "root", "$(transport_user)" );
     am_superhub::
       "$(cfengine_enterprise_federation:config.federation_dir)/superhub/."
         create => "true",
-        perms => default:mog( "701", "root", "root" );
+# TODO why did this have 701??? change and test.
```
olehermanse: Looks like this comment is now outdated.

cfe_internal/enterprise/federation/federation.cf
olehermanse: @nickanderson yes,but IMO this should be explicit - `rxdirs` should default to false ;)

nickanderson: Maybe, but for as long as I have used cfengine it has defaulted to the "do what I mean" behavior. It would be very disruptive to change that default now. We can introduce new bodies where rxdirs is not on by default, default:mog_explicit or some other better name.

craigcomstock:
will wait for travis green then merge.

craigcomstock:
Will we cherry pick this back to 3.12.x to enable superhub back there? It seems doubtful but thought I'd leave this open for now.

PR #1372, opened 2019-04-23
ENT-4500: per-hub partitioned tables on superhub

craigcomstock:
Changelog: title


craigcomstock:
I suppose this sql script could live in nova/db/ but that's a fine point. Also needed will be something to run this script based on "schema needs to change" from policy point of view. Probably superhub enabled and setup-status.json not yet written or something.

PR #1374, opened 2019-04-26
Refresh inventory after import on superhub

vpodzime:
Refreshing inventory on a superhub with tens of thousands of
hosts can easily take a couple minutes and thus should not be
part of every agent run. It's also not necessary to refresh the
inventory on a superhub too often because reporting data only
changes when successfully imported from the feeder hubs.

Let's just refresh the inventory after a successfull data import
on the superhub.

Ticket: ENT-4527
Changelog: None

### DIFF
--- a/cfe_internal/enterprise/CFE_hub_specific.cf
+++ b/cfe_internal/enterprise/CFE_hub_specific.cf
@@ -14,6 +14,11 @@
 #
 ##################################################################
 
+body file control
+{
+        inputs => { "$(sys.inputdir)/cfe_internal/enterprise/federation/federation.cf" };
+}
+
 bundle common cfe_internal_hub_vars
 # @brief Set hub specific variables
 {
@@ -459,6 +464,18 @@ bundle agent cfe_internal_enterprise_maintenance
         ifvarclass => isgreaterthan( $(enterprise_maintenance_bundle_count), 0 );
 }
 
+bundle agent cfe_internal_refresh_inventory_args
+{
+  vars:
+    cfengine_enterprise_federation:am_superhub::
+      "args" -> {"ENT-4527"}
+        string => "inventory_refresh_by_hostkey $(sys.key_digest)",
+        comment => "Only refresh inventory for the superhub, other hosts get refreshed on data import";
+
+    !cfengine_enterprise_federation:am_superhub::
+      "args" string => "inventory_refresh";
+}
+
 bundle agent cfe_internal_refresh_inventory_view
 # @brief Refresh materialized view every 5 minutes
 {
@@ -468,17 +485,21 @@ bundle agent cfe_internal_refresh_inventory_view
 
       "tags" slist => { "enterprise_maintenance" };
 
+  methods:
+      # we need to know if we are running on a superhub
+      "superhub_config" usebundle => "cfengine_enterprise_federation:config";
+      "inventory_refresh_args" usebundle => "cfe_internal_refresh_inventory_args";
+
   commands:
 
     (policy_server|am_policy_hub).enterprise_edition::
 
       "$(sys.workdir)/httpd/php/bin/php"
-        args => "$(cfe_internal_hub_vars.docroot)/index.php cli_tasks inventory_refresh",
+        args => "$(cfe_internal_hub_vars.docroot)/index.php cli_tasks $(cfe_internal_refresh_inventory_args.args)",
         contain => silent,
         comment => "This refreshes the inventory view. If the inventory view is not refreshed then it will contain stale data.",
         handle  => "mpf_fresh_inventory_view",
         if => isdir( "$(cfe_internal_hub_vars.docroot)/api/modules/inventory" );
-
 }
 
 bundle agent cfe_internal_refresh_hosts_view
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -1,6 +1,6 @@
 body file control
 {
-        inputs => { "$(sys.libdir)/stdlib.cf" };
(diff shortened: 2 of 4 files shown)


vpodzime:
This is an untested WIP. I only submitted it to get feedback on the approach and get some advice about how to reuse variables and classes between bundles.

vpodzime:
> This is an untested WIP. I only submitted it to get feedback on the approach and get some advice about how to reuse variables and classes between bundles.

Still untested, but I believe it should be in an almost-final shape now.

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -468,17 +474,29 @@ bundle agent cfe_internal_refresh_inventory_view
...
+      "inventory_refresh_args" -> {"ENT-4527"}
+        string => "inventory_refresh_by_hostkey $(sys.key_digest)",
+        comment => "Only refresh inventory for the superhub, other hosts get refreshed on data import";
+
+    !cfengine_enterprise_federation:am_superhub::
+      "inventory_refresh_args" string => "inventory_refresh";
+
+  methods:
+      # we need to know if we are running on a superhub
+      "superhub_config" usebundle => cfengine_enterprise_federation:config;
```
nickanderson: Is the full 3 pass convergence required to know `am_superhub`?

The variables defined above, based on this bundle converging may not know about am_superhub until the next pass.

nickanderson: When I run into these cases where normal order and 3 passes don't jive well, I usually break them into seperate bundles. So i might have a bundle agent inventory_refresh_args where those vars are defined, and have that called before superhub_config

vpodzime: You mean **after** `superhub_config`?

nickanderson: yes, sorry, something like:

```
  bundle agent cfe_internal_refresh_inventory_view
  {
    meta:
      (policy_server|am_policy_hub).enterprise_edition::
  
        "tags" slist => { "enterprise_maintenance" };

    methods:
      # Do we want to run this policy always?
      "Superhub Configuration"
        comment => "The superhub configuration policy defines classes and
                    variables needed to determine the proper way to refresh
                    inventory. We give it a full 3 pass evaluation to ensure
                    they are resolved before refreshing inventory.",
        usebundle => cfengine_enterprise_federation:config;

      "Refresh Inventory"
        comment => "Because people would like to have up to date information
                    about the state of hosts in their infrastructure.",
        usebundle => cfe_internal_refresh_inventory_view_actually;
  }
  bundle agent cfe_internal_refresh_inventory_view_actually
  {
    vars:
      cfengine_enterprise_federation:am_superhub::
        "inventory_refresh_args" -> {"ENT-4527"}
          string => "inventory_refresh_by_hostkey $(sys.key_digest)",
          comment => "Only refresh inventory for the superhub, other hosts get
                      refreshed on data import";

      !cfengine_enterprise_federation:am_superhub::
        "inventory_refresh_args" string => "inventory_refresh";

    commands:

      (policy_server|am_policy_hub).enterprise_edition::
  
        "$(sys.workdir)/httpd/php/bin/php"
          args => "$(cfe_internal_hub_vars.docroot)/index.php cli_tasks $(inventory_refresh_args)",
          contain => silent,
          comment => "This refreshes the inventory view. If the inventory view
                      is not refreshed then it will contain stale data.",
          handle  => "mpf_fresh_inventory_view",
          if => isdir( "$(cfe_internal_hub_vars.docroot)/api/modules/inventory" );
  }
```

vpodzime: I chose a bit different strategy.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -1,6 +1,6 @@
 body file control
 {
-        inputs => { "$(sys.libdir)/stdlib.cf" };
+        inputs => { "$(sys.libdir)/stdlib.cf", "cfe_internal/enterprise/CFE_hub_specific.cf" };
```
nickanderson: With this input path, what happens if you try to run the federation policy directly leveraging bundle agent __main__?

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -14,6 +14,12 @@
 #
 ##################################################################
 
+body file control
+{
+        inputs => { "cfe_internal/enterprise/federation/federation.cf" };
+}
+
```
nickanderson: I think one of these blank lines can be redacted.

cfe_internal/enterprise/CFE_hub_specific.cf
nickanderson: I saw, pretty similar, same effect and the intention is clear.

cfe_internal/enterprise/CFE_hub_specific.cf
vpodzime: Yup

cfe_internal/enterprise/federation/federation.cf
vpodzime: Good catch. I added `$(sys.inputdir)` there.

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2617/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2617/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2617/

vpodzime:
I checked that a non-FR still refreshes inventory just fine. FR setup seems to be doing the right thing too.

PR #1378, opened 2019-05-03
ENT-4601 Accounted for database needs in state permissions

nickanderson:
The database user must be able to access the parent directory of where the
database is installed or it won't function.

Changelog: None


cf-bottom:

PR #1383, opened 2019-05-07
CFE-3049: Condition use of curl on binary existing and being executable (3.10.x)

nickanderson:


cf-bottom:

PR #1390, opened 2019-05-07
ENT-4601: Account for database needs in state permissions (3.12)

nickanderson:
statedir promises have been moved from bundle agent cfe_internal_setup_knowledge
which was only run on a hub to a bundle that executes on clients as well so that
we can be sure that client permissions are restrictive as well.

Changelog: None
Ticket: ENT-4601
(cherry picked from commit fe6d1fc39c6b179b84658ae87c037e9c51a620b6)


cf-bottom:

olehermanse:
Debian 6 failure due to errors in gpg and java. Seems unrelated.

PR #1401, opened 2019-05-21
ENT-4611: when superhub enabled ensure partition schema is correct

craigcomstock:
merge with 
https://github.com/cfengine/nova/pull/1421

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -213,6 +213,10 @@ bundle agent federation_manage_files
         copy_from => default:local_dcp( "$(this.promise_dirname)/../../../templates/federated_reporting/parallel.sh" ),
         perms => default:mog( "640", "root", "$(transport_user)" );
 
+      "$(cfengine_enterprise_federation:config.bin_dir)/psql_wrapper.sh"
+        copy_from => default:local_dcp( "$(this.promise_dirname)/../../../templates/federated_reporting/psql_wrapper.sh" ),
+        perms => default:mog( "700", "root", "root" );
+
     am_feeder::
       "$(cfengine_enterprise_federation:config.bin_dir)/dump.sh"
         copy_from => default:local_dcp( "$(this.promise_dirname)/../../../templates/federated_reporting/dump.sh" ),
@@ -316,6 +320,26 @@ bundle agent imported_data
         contain => default:in_shell;
 }
 
+bundle agent superhub_schema
+# @brief Run SQL script to ensure schema is migrated to superhub partitioned tables architecture
+{
+  commands:
+    am_superhub::
+      "$(cfengine_enterprise_federation:config.bin_dir)/psql_wrapper.sh"
+        arglist => {
+                     "cfdb",
+                     "'select superhub_schema();'",
+                   },
+        classes => superhub_schema_exit_codes;
+}
+
+body classes superhub_schema_exit_codes
+{
+  kept_returncodes => { "0" };
+  repaired_returncodes => { "1" };
+  failed_returncodes => { "2" };
+}
+
 bundle agent entry
 # @brief Conditionally runs all federated reporting bundles
 {
@@ -350,6 +374,8 @@ bundle agent entry
         usebundle => data_transport;
       "CFEngine Enterprise Federation Feeder Data Import"
         usebundle => imported_data;
+      "CFEngine Enterprise Federation Schema Migration"
+        usebundle => superhub_schema;
       "Configuration Status"
         usebundle => setup_status;
   reports:
--- a/templates/federated_reporting/psql_wrapper.sh
+++ b/templates/federated_reporting/psql_wrapper.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# run an arbitrary psql command as cfpostgres user
+# expectation is that a line will be output with exit_code=<n>
+# this exit code <n> will then be used to exit
+# This design enables cfengine policy to run a commands promise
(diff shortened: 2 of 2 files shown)


templates/federated_reporting/psql_wrapper.sh
```diff
@@ -0,0 +1,14 @@
...
+# run an arbitrary psql command, either -f or -c
+# expectation is that a line will be output with exit_code=<n>
+# this exit code <n> will then be used to exit
+# This design enables cfengine policy to run a commands promise
+# that can be kept, repaired or failed depending on an exit
+# code from psql since psql can't get different codes from scripts.
+
+set -e
+
+EXIT_CODE=$(psql -qtAX "$@" 2>&1 | awk -F= '{print $2}')
```
craigcomstock: what if something else fails here? like psql command path not correct? Maybe add some more error handling and/or adjust failed_returncodes in the policy.

nickanderson: can you use long options? /me is dumb and doenst know exactly what all qtAX does.
what tuples (so no headers), quiet or is that no headers?, A? X does horizontal tables wich is more friendly for line processing. What if the awk fails? Are we sure to exit badly if the query doesnt return what you expect?

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -316,6 +320,27 @@ bundle agent imported_data
         contain => default:in_shell;
 }
 
+bundle agent superhub_schema
+# @brief Run SQL script to migrate to superhub partitioned table schema
```
nickanderson: Run the SQL to ensure the table has been migrated (small wording change to indicate the script is always run, but change is not always done.

templates/federated_reporting/psql_wrapper.sh
vpodzime: Yeah, please use long options and split this to multiple lines. Or at least use `pipefail`.

templates/federated_reporting/psql_wrapper.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [ $# -ne 2 ]
+then
+    echo "Usage: $0 [DBNAME] [SQL STRING]"
+    exit 2
+fi
+
+TMP=$(mktemp)
+TMP2=$(mktemp)
+OUT=$(psql --quiet --tuples-only --no-align --no-psqlrc "$1" -c "$2" 2> $TMP ; echo $? > $TMP2)
```
nickanderson: thank you

templates/federated_reporting/psql_wrapper.sh
craigcomstock: Cool. I split it with my latest change.

craigcomstock:
@vpodzime: squashed, will wait for travis before merging or if you have any other comments.

templates/federated_reporting/psql_wrapper.sh
vpodzime: Please double check that the `$?` is only expanded after the `psql`. I think it might be better to just assign it to the `RETURN_CODE` variable after the assignment to `OUT`.

templates/federated_reporting/psql_wrapper.sh
```diff
@@ -0,0 +1,32 @@
...
+ERR=$(<$TMP)
+
+EXIT_CODE=$(echo $TMP | awk -F= '{ if ( /exit_code/ ) print $2}')
+RETURN_CODE=$(<$TMP2)
+
+rm $TMP $TMP2
+
+if (( $RETURN_CODE != 0 )); then
```
vpodzime: Please stick to one way of shell syntax and code style. The two `if` statements in this file look totally different while they are doing the same.

templates/federated_reporting/psql_wrapper.sh
vpodzime: The exit code of the subshell will be the exit code of its last command.

craigcomstock: fixed

templates/federated_reporting/psql_wrapper.sh
craigcomstock: fixed

templates/federated_reporting/psql_wrapper.sh
```diff
@@ -0,0 +1,31 @@
...
+ERR=$(<$TMP)
+
+EXIT_CODE=$(echo $TMP | awk -F= '{ if ( /exit_code/ ) print $2}')
+
+rm $TMP $TMP2
+
+if [ $RETURN_CODE -ne 0 ]; then
+    echo "$OUT"
+    echo "$ERR"
+    exit 2; # failed
```
vpodzime: Shouldn't this just let it go further to the `exit $EXIT_CODE` below?

craigcomstock: I should add a comment about what exit codes mean from this script and how they are used from policy.

psql return codes are not sufficient to return to policy. Returning 2 for any psql failure is the correct thing to do.

psql:
```
EXIT STATUS
       psql returns 0 to the shell if it finished normally, 1 if a fatal error of its own
       occurs (e.g. out of memory, file not found), 2 if the connection to the server went
       bad and the session was not interactive, and 3 if an error occurred in a script and
       the variable ON_ERROR_STOP was set.
```

But I need to always return exit codes as follows:
```
body classes superhub_schema_exit_codes
{
  kept_returncodes => { "0" };
  repaired_returncodes => { "1" };
  failed_returncodes => { "2" };
}
```

PR #1402, opened 2019-05-22
Always set files_single_copy from augments if available

vpodzime:
Not only when the mpf_control_agent_default_repository class is
defined.


nickanderson:
YES

nickanderson:
I will merge and backport.

PR #1404, opened 2019-05-22
Always set files_single_copy from augments if available (3.10.x)

nickanderson:
Not only when the mpf_control_agent_default_repository class is
defined.

(cherry picked from commit b8f0f0ddc29a2c46538c238dd853df2568723d4d)


cf-bottom:

PR #1407, opened 2019-05-24
Set default access promises for directories to only share if directory exists (3.10.x)

nickanderson:
In particular, /var/cfengine/templates does not exist by default. Running
cf-serverd with log level info emits messages noting the path does not exist.

info: Failed to canonicalise filename '/var/cfengine/templates/' (realpath: No such file or directory)
info: Path does not exist, it's added as-is in access rules: /var/cfengine/templates/

Instead of sending a user chasing to see if something is wrong, we now just
avoid providing access to paths that are not directories when we expect them to
be.

Ticket: CFE-3060
Changelog: Title
(cherry picked from commit 353169594ff6d1377350d56029fe60dd88677f40)

### DIFF
--- a/controls/cf_serverd.cf
+++ b/controls/cf_serverd.cf
@@ -85,33 +85,39 @@ bundle server access_rules()
       handle => "server_access_grant_access_policy",
       shortcut => "masterfiles",
       comment => "Grant access to the policy updates",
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #1408, opened 2019-05-27
Policy fixes for federated reporting i2/i3

olehermanse:
See: https://github.com/cfengine/system-testing/pull/213

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -383,7 +383,7 @@ bundle agent setup_status
       "$(cfengine_enterprise_federation:config.federation_dir)/cfapache/setup-status.json"
         create => "true",
         template_method => "inline_mustache",
-        edit_template_string => "{{%-top-}}",
+        edit_template_string => "{{%-top-}}$(const.n)",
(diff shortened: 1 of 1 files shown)


olehermanse:
@craigcomstock @nickanderson Feel free to review this. Will wait for API changes before I can do more here and eventually get everything tested and merged. (So it's not done, hence the WIP label).

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -361,6 +366,8 @@ bundle agent entry
 bundle agent setup_status
 {
 @if minimum_version(3.11)
```
nickanderson: @olehermanse I see that `bundle agent setup_status` is called for any case where federated reporting is `enabled`.  If we want to support 3.10 feeders this needs to be re-worked to not use the `with` attribute (introduced in 3.11).

Do you want me to open a separate PR or push to your fork or ...?

nickanderson: Well, here is a PR against your PR.
https://github.com/olehermanse/masterfiles/pull/1

olehermanse: Thanks!

PR #1409, opened 2019-05-28
Pull support for Federated Reporting dump transport

vpodzime:
Support for pulling SQL dumps from the feeders by superhub. Just like we, CFE people, like it.

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -33,17 +33,29 @@ bundle agent config
                        strcmp( "enabled", "$(data[remote_hubs][$(remotes)][target_state])" ),
                        # To ensure the remote we are pushing to actually needs the data (is a superhub)
                        strcmp( "superhub", "$(data[remote_hubs][$(remotes)][role])" ));
+
+    am_superhub::
+      # Public keys of enabled pushing feeders need to be trusted (on a superhub)
       "pubkey[$(remotes)]" string => "$(data[remote_hubs][$(remotes)][transport][ssh_pubkey])",
-            if => and(
-                       # To ensure we are using a remote hub that's actually enabled
-                       strcmp( "enabled", "$(data[remote_hubs][$(remotes)][target_state])" ),
-                       # To ensure the remote is a feeder that we need to add to superhub's authorized_keys
-                       strcmp( "feeder", "$(data[remote_hubs][$(remotes)][role])" ));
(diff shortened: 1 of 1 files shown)


vpodzime:
This is missing the required changes in the SSH keys setup and trust establishing. But with that, I'm waiting for @olehermanse to finish #1408 to avoid extra work due to merge conflicts.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -54,20 +54,34 @@ bundle agent config
...
+
+      "am_puller"
+        and => {"am_superhub",
+                strcmp("enabled", "$(data[remote_hubs][$(remotes)][target_state])"),
+                strcmp("pull_over_rsync", "$(data[remote_hubs][$(remotes)][transport][mode])")},
+        comment => "Superhub with some enabled remote hub with pull as transport method, should run pull transport",
         scope => "namespace";
 
+      "am_transporter"
+        or => {"am_pusher", "am_puller"};
```
nickanderson: am_pusher and am_transporter are both namespace scoped classes that should be accessible from other bundles, shouldn't am_transporter be namespace scoped also?

vpodzime: Yes, just like it was before my change. Good catch!

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -33,13 +33,23 @@ bundle agent config
                        strcmp( "enabled", "$(data[remote_hubs][$(remotes)][target_state])" ),
                        # To ensure the remote we are pushing to actually needs the data (is a superhub)
                        strcmp( "superhub", "$(data[remote_hubs][$(remotes)][role])" ));
+
+      # Public keys of enabled pushing feeders need to be trusted (on a superhub)
+      # XXX: would it be okay to guard this with 'am_superhub::'???
```
nickanderson: Hummm.
worst case (but I know that the agent looks deeper), following normal order, enabled is not set during first pass of vars, enabled gets defined during first pass of classes, second pass of vars data is loaded, second pass of classes am_superhub is defined, third pass of vars pubkey array if guarded with am_superhub is generated.

Yeah, I think that it would be fine to guard that with am_superhub, you should still resolve everything within the three passes.

In fact I think if data => readjson( path ) were moved outside of the enabled guard, and guarded with if => fileexists() then we might be able to reduce the number of passes required to reach full convergence.

Either way, yes, I think we can guard the generation of the pubkey array.

PR #1414, opened 2019-06-04
ENT-4704: Added ssh server fingerprint to FR setup-status.json

olehermanse:


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -434,7 +434,8 @@ bundle agent setup_status
     "ssh_pub_key"
       string => readfile( "$(cfengine_enterprise_federation:transport_user.ssh_pub_key)" ),
       if => fileexists( "$(cfengine_enterprise_federation:transport_user.ssh_pub_key)" );
-
+    "ssh_server_fingerprint"
+      string => execresult("ssh-keyscan -t ecdsa -H localhost 2>/dev/null", useshell);
```
nickanderson: This gives you the data you want?

```
> $ ssh-keyscan -t ecdsa -H localhost                                                [±3.7 ●]
# localhost:22 SSH-2.0-OpenSSH_7.9p1 Ubuntu-10
|1|Oknn52tOM+sZ/rwApDXJ0v93gUY=|sIPQWBaapfjmDgHSlQVqrYqR0gI= ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBB7BkQAUSfVHfbsswmCcK43ES3rvprk+s33c4XTslhWOdlEzUf/zfQpHX/psjFswkyQSil9TtI7rXfqGhYcsQGU=
```

What about readfile( '/etc/ssh/ssh_host_ecdsa_key.pub' )?

```
> $ sudo cat /etc/ssh/ssh_host_ecdsa_key.pub
ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBB7BkQAUSfVHfbsswmCcK43ES3rvprk+s33c4XTslhWOdlEzUf/zfQpHX/psjFswkyQSil9TtI7rXfqGhYcsQGU= root@nickanderson-ThinkPad-W550s
```

olehermanse: So, this command was already in our instructions. I just automated it in policy. @craigcomstock added the command originally (to instructions).

We could do it by reading a file, however there are quite a few things to consider:

1. The format in your output from `cat` is different from `ssh-keyscan`. Whatever is using the data would need to be adjusted.
2. `/etc/ssh/ssh_host_ecdsa_key.pub` could be in a different location depending on sshd config.
3. sshd config might also have different locations (?)

It seemed much easier / safer to just add what we were already doing, but yes, it might be slightly slower, performance wise.

nickanderson: That's fine. I think it's worthwhile to drop a hash comment in about why using ssh-keyscan instead of reading from the filesystem, primarily reliability/consistency not having to try and parse the sshd config for host keys that might be in non-standard locations.

PR #1420, opened 2019-06-06
Use the feeder's key digest in names of its dump files

vpodzime:
The host key is a good unique identifier of the feeder and will
help us to identify the partition/schema to insert the data into.

Ticket: ENT-4613
Changelog: None

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -192,6 +192,7 @@ bundle agent federation_manage_files
       string => "$(cfengine_enterprise_federation:config.transport_user)";
     "login" data => parsejson('{"login":"$(cfengine_enterprise_federation:config.login)"}');
     "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
+    "feeder" data => parsejson('{"feeder": "$(sys.key_digest)"}');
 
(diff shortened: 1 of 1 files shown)


cf-bottom:

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -192,6 +192,7 @@ bundle agent federation_manage_files
       string => "$(cfengine_enterprise_federation:config.transport_user)";
     "login" data => parsejson('{"login":"$(cfengine_enterprise_federation:config.login)"}');
     "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
+    "feeder" data => parsejson('{"feeder": "$(sys.key_digest)"}');
```
craigcomstock: @aleksandrychev @vpodzime If we use hostkey then it seems like we should require a hostkey value in the UI that goes to the API to manage remote hubs. Is this practical?

vpodzime: Well, we need **some** unique identifier one way or another, right?

aleksandrychev: @craigcomstock yes, but currently we don't require a name on the UI, I use API to get it automatically, I can rewrite logic to get hostkey as well.

PR #1421, opened 2019-06-06
Sync determine_version.py from core

vpodzime:
It has a couple improvements over the old version originally
copied over here.

Ticket: CFE-3039
Changelog: None


cf-bottom:

PR #1423, opened 2019-06-07
ENT-4613 Import database dumps into feeder schema

craigcomstock:
includes ENT-4611/ENT-4612 for now, will clean that up.

merge with https://github.com/cfengine/nova/pull/1441

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -401,7 +401,7 @@ bundle agent superhub_schema
       "$(cfengine_enterprise_federation:config.bin_dir)/psql_wrapper.sh"
         arglist => {
                      "cfdb",
-                     "'select superhub_schema();'",
+                     `"select superhub_schema('$(sys.key_digest)');"`,
                    },
         classes => psql_wrapper_exit_codes;
 }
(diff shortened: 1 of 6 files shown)


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -399,12 +402,31 @@ bundle agent superhub_schema
       "$(cfengine_enterprise_federation:config.bin_dir)/psql_wrapper.sh"
         arglist => {
                      "cfdb",
-                     "'select superhub_schema();'",
+                     "\"select superhub_schema('$(sys.key_digest)');\"",
```
nickanderson: Personally, I hate escaping quotes. In cfengine you can use backtickts, single quotes or double quotes as quoting characters. So, here where your value needs both single and double quotes, use backticks on the outside and you can avoid having to escape the inner double quotes.

craigcomstock: funny, you just told me that recently and I fixed the other one. Probably had this one "in the queue" and didn't think of fixing it also at the time. :thx:

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -281,6 +281,7 @@ bundle agent federation_manage_files
         copy_from => default:local_dcp( "$(this.promise_dirname)/../../../templates/federated_reporting/pull_dumps_from.sh" ),
         perms => default:mog( "500", "$(transport_user)", "root" );
 
+# TODO run import.sh as "su - cfpostgres" if possible, or inside of import.sh call import_file.sh as cfpostgres user
```
craigcomstock: oops, will remove this TODO. already handled this.

PR #1424, opened 2019-06-07
[3.12.x] Do not run DB maintenance tasks on a passive HA hub

vpodzime:
They all just fail because the DB is read-only on a passive HA
hub. Changes are done on the active one and replicated.

Ticket: ENT-4706
Changelog: Title
(cherry picked from commit 704e262c54ff818d3bd622a48fe70d65d9bc5bf2)


cf-bottom:

PR #1429, opened 2019-06-11
ENT-4526: handle __promiselog* tables in import

craigcomstock:
merge with https://github.com/cfengine/nova/pull/1443

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -310,6 +310,9 @@ bundle agent postgres_config
       "c[max_locks_per_transaction]"
         string => "4000",
         comment => "Changing this setting requires restarting the database.";
+      "c[log_lock_waits]"
+        string => "on",
+        comment => "Changing this setting requires restarting the database.";
   files:
     am_superhub::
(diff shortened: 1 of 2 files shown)


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -310,6 +310,9 @@ bundle agent postgres_config
       "c[max_locks_per_transaction]"
         string => "4000",
         comment => "Changing this setting requires restarting the database.";
+      "c[log_lock_waits]"
```
nickanderson: you think this is something that we will want ON all the time, or is it more for debugging?

craigcomstock: it is really more for debugging, I'm torn really... without it I feel like we get deadlock logging which looks "bad" but maybe isn't because I suspect that postgresql logs locks that have to wait for more than 1 second but without log_lock_waits I think it won't log the "acquired" log message that indicates that things are basically "OK", just a bit slow for a moment when the system is busy.

craigcomstock: I'm going to merge this and add a blocker to decide if we want this.

PR #1431, opened 2019-06-12
ENT-4749: Run pg_ctl as cfpostgres user and not root

craigcomstock:
Changelog: none


cf-bottom:

PR #1432, opened 2019-06-12
ENT-4745: cleaned up output of import_file.sh

craigcomstock:
Changelog: None

### DIFF
--- a/templates/federated_reporting/import_file.sh
+++ b/templates/federated_reporting/import_file.sh
@@ -3,7 +3,6 @@
 #
 # $1 -- compressed SQL file to import
 set -e
-set -x
 
 source "$(dirname "$0")/config.sh"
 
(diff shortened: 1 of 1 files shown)


templates/federated_reporting/import_file.sh
```diff
@@ -22,8 +21,7 @@ hostkey=$(basename "$file" | cut -d. -f1)
 
 mv "$file" "$file.importing"
 
-  # temp step to see the result of sed filter
-  "$CFE_FR_EXTRACTOR" $CFE_FR_EXTRACTOR_ARGS "$file.importing" | sed $CFE_FR_SED_ARGS $(printf ' -f %s' $sed_scripts) > $file.after-sed
+"$CFE_FR_EXTRACTOR" $CFE_FR_EXTRACTOR_ARGS "$file.importing" | sed $CFE_FR_SED_ARGS $(printf ' -f %s' $sed_scripts) > $file.after-sed
```
craigcomstock: I struggle with this a bit. I really want the .after-sed file which shows what we actually imported but if the files are large then this will eat up disk space. Some ideas for debugability? Maybe add a DEBUG flag that could be set in def.json in case there is trouble? As it is only the LATEST dump will be left for the .after-sed and .log files so it won't grow "forever" just like the .failed files are right now.

nickanderson: CFEngine can cull old files

craigcomstock: I will rework this to put everything back "inline" and not generate this .after-sed file to disk. We can debug problems by looking at the .log files and using the .failed dump file along with sed filters to get back to what I had in .after-sed which is the actual SQL run during the import.

craigcomstock: fixed. tested in aws. Looks good to me.

PR #1433, opened 2019-06-13
ENT-4743: address deadlocking issues in serial import

craigcomstock:
merge with https://github.com/cfengine/nova/pull/1444


craigcomstock:
I tested with the package from ENT-4772 + these changes. Biggest problem which is a "todo" is to prevent multiple cf-agent runs from trying to import at the same time. Which could definitely happen if the import takes some time.

PR #1436, opened 2019-06-17
ENT-4773: Fixed statedir group ownership on aix

nickanderson:


cf-bottom:

lib/files.cf
```diff
@@ -1674,6 +1674,9 @@ body perms system_owned(mode)
 
     solaris::
       groups => { "sys" };
+
```
vpodzime: Missing **s** in *system* in the commit message.

nickanderson: ACK

PR #1440, opened 2019-06-20
Do not require specific SSH key type for FR

vpodzime:
Systems may have various restrictions for SSH key types they
use. Let's just leave it to the system configuration to
decide. One of the problematic cases here is RHEL 6 which has an
older version of SSH that doesn't support the specific key type
we were trying to use.

Because we no longer specify the key type, we need to specify the
key file name and then make sure it is being used (SSH only uses
the default-named keys like id_rsa, id_dsa,...).

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -113,11 +113,12 @@ bundle agent transport_user
         string => "$(cfengine_enterprise_federation:config.transport_user)";
       "home"
         string => "$(cfengine_enterprise_federation:config.transport_home)";
-      "ssh_key_type" string => "ed25519";
-      "ssh_priv_key" string => "$(home)/.ssh/id_$(ssh_key_type)";
-      "ssh_pub_key" string => "$(home)/.ssh/id_$(ssh_key_type).pub";
+      "ssh_key_name" string => "id_FR";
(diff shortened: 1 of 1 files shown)


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -165,12 +170,16 @@ bundle agent transport_user
       edit_template_string => "{{#-top-}}{{{.}}}$(const.n){{/-top-}}",
       template_data => @(cfengine_enterprise_federation:config.fingerprints),
       template_method => "inline_mustache";
+    "$(ssh_config)"
+      create => "true",
+      handle => "ssh_config_configured",
```
nickanderson: I would probably call this ssh_configuration_identityfile since that is the only thing it's managing. 

Maybe edit_defaults => empty as well? If there is something in the file then IdentityFile might not get applied by default right?

vpodzime: > I would probably call this ssh_configuration_identityfile since that is the only thing it's managing.

Well, it's still a config file. We may add more things there in the future.

vpodzime: > Maybe edit_defaults => empty as well? If there is something in the file then IdentityFile might not get applied by default right?

Yes, but then we want things to work that way (e.g. a different key for a specific host).

PR #1441, opened 2019-06-21
Separate data import from schema setup and attaching in FR

vpodzime:
ENT-4742: fix deadlock issues with parallel import

The data import is what takes a long time and what we need to
parallelize, setting up the schemas and attaching them can be
done serially before/after the data import.

Merge together: https://github.com/cfengine/nova/pull/1450

### DIFF
--- a/templates/federated_reporting/import.sh
+++ b/templates/federated_reporting/import.sh
@@ -57,7 +57,11 @@ if [ "$failed" = "0" ]; then
   log "Setting up schemas for import: DONE"
 else
   log "Setting up schemas for import: FAILED"
-  # XXX: this needs to revert
+  # remove any newly created schemas (revert the changes)
+  for file in $dump_files; do
+    hostkey=$(basename "$file" | cut -d. -f1)
+    "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb -c "SELECT drop_feeder_schema('$hostkey');" || true
+  done
   exit 1
 fi
 
@@ -71,26 +75,34 @@ if [ "$failed" = "0" ]; then
   log "Importing files: DONE"
 else
   log "Importing files: FAILED"
-  # XXX: this needs to revert (ideally just for the specific failed hosts/dumps)
   for file in "$CFE_FR_SUPERHUB_IMPORT_DIR/*.sql.$CFE_FR_COMPRESSOR_EXT.failed"; do
     log "Failed to import file '${file%%.failed}'"
-    rm -f "$file"
+
+    # revert any changes by dropping the particular feeder's import schema (the
+    # original/in-use/previous schema is left intact)
+    hostkey=$(basename "$file" | cut -d. -f1)
+    "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb -c "SELECT drop_feeder_schema('$hostkey');" || true
   done
-  exit 1
 fi
 
+failed=0
 log "Attaching schemas"
 for file in $dump_files; do
-  hostkey=$(basename "$file" | cut -d. -f1)
-  "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
-                      -c "SET SCHEMA 'public'; SELECT attach_feeder_schema('$hostkey', ARRAY[$table_whitelist]);" \
-    > schema_attach.log 2>&1 || failed=1
+  if [ ! -f "${file}.failed" ]; then
+    hostkey=$(basename "$file" | cut -d. -f1)
+    "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
+                        -c "SET SCHEMA 'public'; SELECT attach_feeder_schema('$hostkey', ARRAY[$table_whitelist]);" \
+      > schema_attach.log 2>&1 || failed=1
+  else
+    rm -f "${file}.failed"
+  fi
 done
 if [ "$failed" = "0" ]; then
   log "Attaching schemas: DONE"
 else
+  # attach_feeder_schema() makes sure the feeder's import schema is removed in
+  # case of failure
   log "Attaching schemas: FAILED"
-  # XXX: anything we can do here to make things ready for the next round of import?
   exit 1
 fi
 
--- a/templates/federated_reporting/import.sh
+++ b/templates/federated_reporting/import.sh
@@ -43,36 +43,65 @@ dump_files="$(ls -1 "$CFE_FR_SUPERHUB_IMPORT_DIR/"*".sql.$CFE_FR_COMPRESSOR_EXT"
     exit 0
   }
 
+table_whitelist=$(printf "'%s'," $CFE_FR_TABLES | sed -e 's/,$//')
+
+failed=0
+log "Setting up schemas for import"
+for file in $dump_files; do
+  hostkey=$(basename "$file" | cut -d. -f1)
+  "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
+                      -c "SELECT ensure_feeder_schema('$hostkey', ARRAY[$table_whitelist]);" \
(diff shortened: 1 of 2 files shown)


cf-bottom:

templates/federated_reporting/import.sh
```diff
@@ -43,36 +43,65 @@ dump_files="$(ls -1 "$CFE_FR_SUPERHUB_IMPORT_DIR/"*".sql.$CFE_FR_COMPRESSOR_EXT"
...
-chmod u+x "$(dirname "$0")/import_file.sh"
+table_whitelist=$(printf "'%s'," $CFE_FR_TABLES | sed -e 's/,$//')
 
-log "Importing files: $dump_files"
 failed=0
-# for now, import in serial to avoid deadlocks (ENT-4742)
+log "Setting up schemas for import"
 for file in $dump_files; do
-  "$(dirname "$0")/import_file.sh" $file || failed=1
+  hostkey=$(basename "$file" | cut -d. -f1)
```
craigcomstock: you use this more than once, why not define it above for all?

templates/federated_reporting/import.sh
```diff
@@ -43,36 +43,65 @@ dump_files="$(ls -1 "$CFE_FR_SUPERHUB_IMPORT_DIR/"*".sql.$CFE_FR_COMPRESSOR_EXT"
...
+  "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
+                      -c "SELECT ensure_feeder_schema('$hostkey', ARRAY[$table_whitelist]);" \
+    > schema_setup.log 2>&1 || failed=1
 done
+if [ "$failed" = "0" ]; then
+  log "Setting up schemas for import: DONE"
+else
+  log "Setting up schemas for import: FAILED"
+  # XXX: this needs to revert
```
craigcomstock: The ensure_feeder_schema() function acts as a transaction. Any changes that are made in that function will be rolled back if there is an error. So no need to revert.

templates/federated_reporting/import.sh
```diff
@@ -43,36 +43,65 @@ dump_files="$(ls -1 "$CFE_FR_SUPERHUB_IMPORT_DIR/"*".sql.$CFE_FR_COMPRESSOR_EXT"
...
+  exit 1
```
craigcomstock: Do we want to exit here or just skip that feeder/hostkey in later steps?

templates/federated_reporting/import_file.sh
```diff
@@ -28,15 +28,12 @@ mv "$file" "$file.importing"
   cat<<EOF
 \set ON_ERROR_STOP 1
 BEGIN;
-SELECT ensure_feeder_schema('$hostkey', ARRAY[$table_whitelist]);
```
craigcomstock: the call to ensure_feeder_schema() could stay here without any problem. There is nothing there that creates a lock other than the momentary ALTER SCHEMA RENAME TO which should happen quite fast.

templates/federated_reporting/import.sh
```diff
@@ -43,36 +43,65 @@ dump_files="$(ls -1 "$CFE_FR_SUPERHUB_IMPORT_DIR/"*".sql.$CFE_FR_COMPRESSOR_EXT"
...
 
-log "Importing files: $dump_files"
 failed=0
-# for now, import in serial to avoid deadlocks (ENT-4742)
+log "Setting up schemas for import"
 for file in $dump_files; do
-  "$(dirname "$0")/import_file.sh" $file || failed=1
+  hostkey=$(basename "$file" | cut -d. -f1)
+  "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
+                      -c "SELECT ensure_feeder_schema('$hostkey', ARRAY[$table_whitelist]);" \
```
craigcomstock: we could probably keep this call to ensure_feeder_schema() in import_file.sh

templates/federated_reporting/import.sh
```diff
@@ -43,36 +43,65 @@ dump_files="$(ls -1 "$CFE_FR_SUPERHUB_IMPORT_DIR/"*".sql.$CFE_FR_COMPRESSOR_EXT"
...
-if [ "$failed" != "0" ]; then
+log "Importing files: $dump_files"
+echo "$dump_files" | run_in_parallel "$(dirname "$0")/import_file.sh" - $CFE_FR_IMPORT_NJOBS ||
+  failed=1
+if [ "$failed" = "0" ]; then
+  log "Importing files: DONE"
+else
   log "Importing files: FAILED"
-  #find "$CFE_FR_SUPERHUB_IMPORT_DIR"
+  # XXX: this needs to revert (ideally just for the specific failed hosts/dumps)
```
craigcomstock: I think we should include the "revert if sql.gz import failed" in import_file.sh that way we don't have to try and keep track or something in this script.

templates/federated_reporting/import.sh
```diff
@@ -43,36 +43,65 @@ dump_files="$(ls -1 "$CFE_FR_SUPERHUB_IMPORT_DIR/"*".sql.$CFE_FR_COMPRESSOR_EXT"
...
-    inv_refresh_failed=0
-    $CFE_FR_INVENTORY_REFRESH_CMD || inv_refresh_failed=1
-    if [ "$inv_refresh_failed" != "0" ]; then
-      log "Refreshing inventory: FAILED"
-      exit 1
-    else
-      log "Refreshing inventory: DONE"
-    fi
+  log "Attaching schemas: FAILED"
+  # XXX: anything we can do here to make things ready for the next round of import?
```
craigcomstock: the call to attach_feeder_schema() should be fairly bullet proof to making things ready for next time. It has an exception handler around the entire function which then reverts the schema name change from hub_n_old to hub_n. It does that work carefully as well to avoid other exceptions, such as using verify_schema(new_schema) and verify_schema(old_schema) before trying to rollback the schema changes.

templates/federated_reporting/import_file.sh
```diff
@@ -28,15 +28,12 @@ mv "$file" "$file.importing"
...
+SELECT switch_to_feeder_schema('$hostkey'); -- so that the import below imports into it
 EOF
 
   "$CFE_FR_EXTRACTOR" $CFE_FR_EXTRACTOR_ARGS "$file.importing" | sed $CFE_FR_SED_ARGS $(printf ' -f %s' $sed_scripts)
 
-cat<<EOF
-
-SET SCHEMA 'public';
-SELECT attach_feeder_schema('$hostkey', ARRAY[$table_whitelist]);
+  cat<<EOF
```
craigcomstock: Let's put the "handle import failure and rollback hub_n_old schema name change" code in the ` || { ... } ` block below around line 42.

templates/federated_reporting/import.sh
craigcomstock: nevermind, depends on the loop and dump_file/$file. ack.

templates/federated_reporting/import.sh
vpodzime: My idea is that the schema manipulation is critical and any failure indicates a potentially big problem with high risk. That's why I think it should terminate the import. If it works next time, great. If it fails again, there's something wrong that needs manual intervention.

templates/federated_reporting/import_file.sh
vpodzime: > which should happen quite fast

This is exactly how deadlocks are usually born. I suggest that we do all the schema changes serially. As you say, they should be quite fast so there's no reason to do them in parallel risking issues. If we later find out that we need to optimize this, parallelization may be one of the options.

templates/federated_reporting/import.sh
vpodzime: See my comment on this above.

templates/federated_reporting/import.sh
vpodzime: Again, I think all schema manipulations should be done serially and thus not in *import_file.sh*.

templates/federated_reporting/import_file.sh
vpodzime: No, as explained above.

PR #1446, opened 2019-06-24
FR, reduced default logging level for import, provided debug option

craigcomstock:
merge with https://github.com/cfengine/nova/pull/1457

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -212,6 +212,11 @@ bundle agent federation_manage_files
     "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
     "feeder" data => parsejson('{"feeder": "$(sys.key_digest)"}');
 
+    "DEBUG|DEBUG_$(this.namespace)_$(this.bundle)"::
+      "debug" data => parsejson('{ "classes": { "DEBUG":true}}');
+    "!(DEBUG|DEBUG_$(this.namespace)_$(this.bundle))"::
+      "debug" data => parsejson('{ "classes": { "DEBUG":false}}');
+
   methods:
       "internal_vars" usebundle => "default:cfe_internal_hub_vars",
         comment => "We need the path to the inventory refresh PHP script before we use it.";
@@ -262,7 +267,7 @@ bundle agent federation_manage_files
         create => "true",
         template_method => "mustache",
         edit_template => "$(this.promise_dirname)/../../../templates/federated_reporting/config.sh.mustache",
-        template_data => mergedata(@(login), @(feeder_username), @(feeder),
+        template_data => mergedata(@(login), @(feeder_username), @(feeder), @(debug),
                                    @(cfengine_enterprise_federation:inventory_refresh_cmd.cmd)),
         perms => default:mog( "640", "root", "$(transport_user)" );
 
--- a/templates/federated_reporting/import_file.sh
+++ b/templates/federated_reporting/import_file.sh
@@ -13,6 +13,10 @@ true "${CFE_FR_EXTRACTOR?undefined}"
 true "${CFE_FR_EXTRACTOR_ARGS?undefined}"
 true "${CFE_FR_DB_USER?undefined}"
 true "${CFE_FR_TABLES?undefined}"
+true "${CFE_FR_DEBUG?undefined}"
+true "${CFE_FR_COMPRESSOR?undefined}"
+true "${CFE_FR_COMPRESSOR_ARGS?undefined}"
+true "${CFE_FR_COMPRESSOR_EXT?undefined}"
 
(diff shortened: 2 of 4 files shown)


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -206,6 +206,11 @@ bundle agent federation_manage_files
     "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
     "feeder" data => parsejson('{"feeder": "$(sys.key_digest)"}');
 
+    "DEBUG|DEBUG_config"::
+      "debug" data => parsejson('{"debug":true}');
+    "!(DEBUG|DEBUG_config)"::
```
nickanderson: No need for the quotes `"` in the class expression unless you are using variables.

So I would do this:

```
DEBUG|DEBUG_config::
      "debug" data => parsejson('{"debug":true}');
!(DEBUG|DEBUG_config)::
```

OR I would do this:

```
"DEBUG|DEBUG_$(this.bundle)"::
      "debug" data => parsejson('{"debug":true}');
"!(DEBUG|DEBUG_$(this.bundle))"::
```

Also, since this federation policy is in a namespace, I would use the namespace in the guard as well.

```
"DEBUG|DEBUG_$(this.namespace)_$(this.bundle)"::
      "debug" data => parsejson('{"debug":true}');
"!(DEBUG|DEBUG_$(this.namespace)_$(this.bundle))"::
```

or



```
DEBUG|DEBUG_cfengine_enterprise_federation_config::
      "debug" data => parsejson('{"debug":true}');
!(DEBUG|DEBUG_cfengine_enterprise_federation_config)::
```

nickanderson: I don't have much of an opinion on the key used for debug in the passed json data.

My first instinct would be to mirror the classes structure to try and make the contexts more familiar, but really, I don't think this matters much at all.

If you wanted to do that, it would be something like this:

```
DEBUG|DEBUG_cfengine_enterprise_federation_config::
      "debug" data => parsejson('{ "classes": { "DEBUG":true}}');
!(DEBUG|DEBUG_cfengine_enterprise_federation_config)::
```
And of course the mustache would need to be updated as well to use `classes.DEBUG` instead of just `debug`.

templates/federated_reporting/import_file.sh
```diff
@@ -22,9 +26,21 @@ hostkey=$(basename "$file" | cut -d. -f1)
...
+    psql_debug_option="--echo-all"
+fi
+
 mv "$file" "$file.importing"
 
 {
+  if $CFE_FR_DEBUG; then
+      echo "\set client_min_messages = notice"
+  else
+      echo "\set client_min_messages = warning"
```
nickanderson: nice

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -212,6 +212,11 @@ bundle agent federation_manage_files
     "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
     "feeder" data => parsejson('{"feeder": "$(sys.key_digest)"}');
 
+    "DEBUG|DEBUG_$(this.namespace)_$(this.bundle)"::
+      "debug" data => parsejson('{ "classes": { "DEBUG":true}}');
+    "!(DEBUG|DEBUG_$(this.namespace)_$(this.bundle))"::
+      "debug" data => parsejson('{ "classes": { "DEBUG":false}}');
```
vpodzime: I don't like this `classes.DEBUG`, let's separate the bash stuff from the policy and keep the bash stuff as simple as possible. IOW, just `{"debug": "true/false"}`, also having it as a string, not a JSON boolean.

templates/federated_reporting/import_file.sh
```diff
@@ -22,9 +26,21 @@ hostkey=$(basename "$file" | cut -d. -f1)
 
 table_whitelist=$(printf "'%s'," $CFE_FR_TABLES | sed -e 's/,$//')
 
+psql_debug_option="--quiet"
+
+if $CFE_FR_DEBUG; then
```
vpodzime: Would be better and safer to just compare strings here.

cfe_internal/enterprise/federation/federation.cf
craigcomstock: yeah, I don't really care, was just rolling with what @nickanderson 's suggestion :)

templates/federated_reporting/import_file.sh
craigcomstock: in terms of command injection or something? makes sense in any case.

vpodzime: Yeah, there's no reason to execute it as a command here, I think.

craigcomstock:
Don't think we need this anymore.

PR #1447, opened 2019-06-25
Make sure /var/cfengine/bin/rsync is used on both sides in FR

vpodzime:
Otherwise /var/cfengine/bin/rsync tries to start 'rsync' on the
remote side and '/var/cfengine/bin' is not in $PATH on the remote
side.


cf-bottom:

PR #1448, opened 2019-06-25
Dump data on feeders every 20 minutes not in every agent run

vpodzime:
Together with the rest of the ETL process this should result in
data being refreshed on the superhub every ~30 minutes.

Also note that we don't have to change the interval on the
superhub because there we want to always try to get fresh dumps
and import them. Of course, normal promise locking rules apply so
an import will only start once the previous one is done.

Ticket: ENT-4806
Changelog: none

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -12,6 +12,12 @@ bundle agent config
       "federation_dir" string => "/opt/cfengine/federation";
       "bin_dir" string => "$(federation_dir)/bin";
       "path" string => "$(federation_dir)/cfapache/federation-config.json";
+
+      # TODO: Instrument augments
+      "dump_interval" -> {"ENT-4806"}
+        int => "20",
+        comment => "Dump data on the feeders every 20 minutes";
+
   classes:
     enterprise_edition.(policy_server|am_policy_hub)::
       "enabled"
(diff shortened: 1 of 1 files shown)


cf-bottom:

craigcomstock:
@cf-bottom jenkins build please

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -12,6 +12,12 @@ bundle agent config
       "federation_dir" string => "/opt/cfengine/federation";
       "bin_dir" string => "$(federation_dir)/bin";
       "path" string => "$(federation_dir)/cfapache/federation-config.json";
+
+      # TODO: Instrument augments
+      "dump_interval" -> {"ENT-4806"}
+        int => 20,
```
nickanderson: I am surprised that it isn't complaining about missing quotes here.

```
int => "20", 
```

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -278,7 +284,8 @@ bundle agent federation_manage_files
     am_feeder::
       "$(cfengine_enterprise_federation:config.bin_dir)/dump.sh"
         copy_from => default:local_dcp( "$(this.promise_dirname)/../../../templates/federated_reporting/dump.sh" ),
-        perms => default:mog( "700", "root", "root" );
+        perms => default:mog( "700", "root", "root" ),
+        action => if_elapsed("$(cfengine_enterprise_federation:config.dump_interval");
```
nickanderson: I don't think that you need to quote the variable , also it looks like you are missing a closing paren `)`  for `if_elapsed`.

```
         action => if_elapsed( $(cfengine_enterprise_federation:config.dump_interval) );
```

PR #1452, opened 2019-06-26
FR blocker fixes

vpodzime:
Two small fixes for potentially big issues with the ETL process in FR.


cf-bottom:

PR #1455, opened 2019-06-28
Fix failure propagation in parallel.sh when using background jobs

vpodzime:
When waiting for the jobs to finish, we need to wait for one at a
time, check its exit code and record it in case of
failure. Waiting for all jobs at once just gives us an exit code
of a random one of them. So a failure can potentially be
ignored. 'xargs' and 'parallel' both handle this for us.


cf-bottom:

PR #1464, opened 2019-07-09
CFE-3094: yum package_module gets updates available from online repos if local cache fails (3.12.x)

nickanderson:


cf-bottom:

PR #1468, opened 2019-07-11
ENT-4753: Added monitoring for postgresql lock acquisition times.

craigcomstock:
To aid in determining performance of postgresql and specifically
Federated Reporting superhubs.

Changelog: Title

### DIFF
--- a/lib/cfe_internal_hub.cf
+++ b/lib/cfe_internal_hub.cf
@@ -361,3 +361,25 @@ bundle agent cfe_internal_postgresql_vacuum
       comment => "Run vacuum db command (database: cfdb)",
       handle => "cfe_internal_postgresql_maintenance_commands_run_vacuumdb";
 }
+
+bundle monitor measure_psql
+{
+  measurements:
+    policy_server|am_policy_hub::
+      "/var/log/postgresql.log"
+        if => fileexists( "/var/log/postgresql.log"),
+        handle => "psql_lock_wait_before_acquisition",
+        stream_type => "file",
+        data_type => "real",
+        history_type => "weekly",
+        units => "ms",
+        match_value => acquired_lock;
+}
+
+body match_value acquired_lock
+{
+  select_line_matching => ".*acquired.*Lock.*";
+  extraction_regex => ".*after (\d+)";
+  track_growing_file => "true";
+  select_multiline_policy => "average"; # average, sum, first, last
+}


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -410,6 +410,27 @@ body contain contain_transport_user
     useshell => "true";
 }
 
+bundle monitor measure_psql
```
nickanderson: So this monitor bundle is in the federation namespace? Ok, I haven't ever namespaced a measurement before, glad to know it works in some way.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -410,6 +410,27 @@ body contain contain_transport_user
...
+{
+  measurements:
+    "/var/log/postgresql.log"
+      handle => "psql_lock_wait_before_acquisition",
+      stream_type => "file",
+      data_type => "real",
+      history_type => "weekly",
+      units => "ms",
+      match_value => acquired_Lock,
+      comment => "Are comments significant?";
```
nickanderson: Comments are not significant with respect to how mission portal graphs stuff. That was just junk from my testing. Maybe change the comment or yank it?

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -410,6 +410,27 @@ body contain contain_transport_user
     useshell => "true";
 }
 
+bundle monitor measure_psql
+{
+  measurements:
```
nickanderson: Oh, i nearly forgot. This hsould be guarded so that it does not try to activate on any other host. Monitor bundles like server bundles don't need to be in the bundlesequence, so they need to be carefully guarded so as not to apply on all hosts.

nickanderson: So id do am_policy_hub| policy_server and maybe if => fileexists( "/var/log/postgresql.log" )

craigcomstock: I tried this with `am_superhub::` but that didn't seem to work, maybe due to the policy being evaluated differently in cf-monitord?

cfe_internal/enterprise/federation/federation.cf
craigcomstock: yank, will do.

cfe_internal/enterprise/federation/federation.cf
craigcomstock: not sure.

cfe_internal/enterprise/federation/federation.cf
nickanderson: Yeah, as we discussed in chat, I think because am_superhub is defined from an agent bundle AND it's not a persistent class so monitor bundles won't know about it.

craigcomstock:
@nickanderson Ok. Cleaned up. Tested with/without the file existing and coming back, worked.

craigcomstock:
@nickanderson this look good to you now? Want to commit?

PR #1472, opened 2019-07-24
ENT-4874: Changed FR policy to honor target_state properly

craigcomstock:
Also added a unit test at tests/unit/test_federation_policy

Changelog: title

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -96,19 +96,31 @@ bundle agent config
         or => {"am_pusher", "am_puller"},
         scope => "namespace";
 
+      "am_paused"
+        expression => strcmp("paused", "$(data[target_state])"),
+        scope => "namespace";
+
+  # Note: in order to see these debugs you must either define the default DEBUG class
+  # or the namespace prefixed class like:
+  # cf-agent -KI -DDEBUG
+  # or
+  # cf-agent -KI -Dcfengine_enterprise_federation:DEBUG_config
+
   reports:
-    enabled.(DEBUG|DEBUG_config)::
+    enabled.(default:DEBUG|DEBUG_config)::
       "Federation enabled!";
-    am_superhub.(DEBUG|DEBUG_config)::
+    am_superhub.(default:DEBUG|DEBUG_config)::
       "I'm a superhub!";
-    am_feeder.(DEBUG|DEBUG_config)::
+    am_feeder.(default:DEBUG|DEBUG_config)::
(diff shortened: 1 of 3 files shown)


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -109,6 +116,8 @@ bundle agent config
       "I'm pulling dumps!";
     am_transporter.(DEBUG|DEBUG_config)::
       "I'm a transporter!";
+    am_paused.(DEBUG|DEBUG_config)::
```
nickanderson: Should we look for default:DEBUG?

nickanderson: So then it would be like (default:DEBUG|DEBUG_config)

And you could run `cf-agent --define DEBUG` OR `cf-agent --define cfengine_enterprise_federation:DEBUG_config`

tests/unit/test_federation_policy
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
nickanderson: Well, this is interesting. Instead of doing this single unit test, shouldn't it run all unit tests? I can probably find other cases of tests that would be easier to do from bash like this than how some of the policy acceptance tests are done

cfe_internal/enterprise/federation/federation.cf
craigcomstock: oh. yes! I agree!

tests/unit/test_federation_policy
craigcomstock: I added test_federation_policy to Makefile.am so would expect it to get run: https://github.com/cfengine/masterfiles/pull/1472/files#diff-156d4744b2414baf39a261e75355322eR8

cfe_internal/enterprise/federation/federation.cf
craigcomstock: @nickanderson changed and added a longer comment as explanation. I suppose we might have a top-level place where we recommend the use of a DEBUG class? Because it doesn't directly get defined if you do `cf-agent --debug`.

nickanderson: its an mpfism https://docs.cfengine.com/docs/3.12/guide-writing-and-serving-policy-policy-style.html#policy-reports

tests/unit/test_federation_policy
nickanderson: right, i saw that, it just made me think how we need general bash tests, no framework really exists in our stuff for bash tests today

PR #1479, opened 2019-08-08
ENT-4906: added ssh_home_t type to cftransport .ssh dir

craigcomstock:
On Centos6 the agent was unable to generate a key with
ssh-keygen due to mismatched contexts on cftransport .ssh
dir.

Changelog: title

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -153,12 +153,17 @@ bundle agent transport_user
           "$(ssh_known_hosts)",
           "$(ssh_config)"
         };
+      "ssh_home_cil_file" string => "/tmp/cftransport_ssh_home_t.cil";
+
   classes:
     enabled::
       "selinux_enabled"
-        expression => strcmp("Enforcing", execresult("getenforce", useshell));
+        expression => strcmp("Enforcing", execresult("$(paths.getenforce)", noshell));
       "missing_ssh_context"
         expression => not( or(
+                               regcmp(".*[\s:]ssh_home_t[\s:].*",
+                                      execresult("ls -Z $(home) | grep .ssh",
+                                                 useshell)),
                                regcmp(".*[\s:]ssh_home_t[\s:].*",
                                       execresult("ls -Z $(ssh_auth_keys)",
                                                  useshell)),
@@ -200,7 +205,18 @@ bundle agent transport_user
       handle => "ssh_config_configured",
       edit_line => default:insert_lines("IdentityFile $(ssh_priv_key)");
 
+    selinux_enabled.missing_ssh_context::
+      "$(ssh_home_cil_file)"
+        create => "true",
+        edit_defaults => default:empty,
+        edit_line => default:insert_lines('(filecon "$(home)/.ssh" any (system_u object_r ssh_home_t ((s0)(s0))))');
+
   commands:
+    selinux_enabled.missing_ssh_context::
+      # the filename becomes the name for the selinux policy (as listed by semodule -l)
+      "$(paths.semodule) -i $(ssh_home_cil_file)";
+      "$(paths.restorecon) -R -F  $(home)/.ssh/";
+
     # Generate a modern ssh keypair (2019-04)
     "/usr/bin/ssh-keygen"
       handle => "ssh_keys_configured",
@@ -208,9 +224,6 @@ bundle agent transport_user
       if => and( isdir( "$(home)/.ssh" ),
                  not( fileexists( "$(ssh_priv_key)" )));
 
-    selinux_enabled.missing_ssh_context::
-      "restorecon -R -F  $(home)/.ssh/"
-        contain => default:in_shell;
 }
 
 bundle agent inventory_refresh_cmd
--- a/lib/paths.cf
+++ b/lib/paths.cf
@@ -345,6 +345,7 @@ bundle common paths
       "path[ethtool]"       string => "/usr/sbin/ethtool";
       "path[find]"          string => "/usr/bin/find";
       "path[free]"          string => "/usr/bin/free";
+      "path[getenforce]"    string => "/usr/sbin/getenforce";
       "path[grep]"          string => "/bin/grep";
       "path[hostname]"      string => "/bin/hostname";
       "path[init]"          string => "/sbin/init";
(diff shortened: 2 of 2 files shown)


cf-bottom:

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -202,6 +202,8 @@ bundle agent transport_user
                  not( fileexists( "$(ssh_priv_key)" )));
 
     selinux_enabled.missing_ssh_context::
+      "semanage fcontext -a -t ssh_home_t $(home)/.ssh/"
+        contain => default:in_shell;
```
vpodzime: AFAICT, this will run **after** *ssh-keygen* above.

craigcomstock: I will move it before since order inside of a promise type is based on occurrence in the file right? Otherwise multiple passes would take care of it.

craigcomstock:
> In general this is quite weird because:
> a) We have never seen before. (in tests)
> b) `/opt/cfengine/federated-reporting/cftransport` **is** the home dir of the `cftransport` user and thus as `~cftransport/.ssh` the folder should have the right label.

I don't think it is weird because we have always run policy as root so our user had the unconfined_u context. The .ssh dir ends up with system_u:object_r:usr_t:s0 in the case where it failed for me and when policy runs by itself is it running as something other than unconfined_u because of running from systemd/cf-exec right? I will modify the policy slightly, get a build, and re-test. :+1:

craigcomstock:
Another problem here was that there is no rule for .ssh in home dirs that are not of the pattern /home/*/.ssh so when we just did restorecon before cftransport's home dir wasn't being changed to ssh_home_t as we might have expected it to be because it's home is in /opt/cfengine/federated-reporting/cftransport instead of /home/cftransport.

craigcomstock:
@cf-bottom jenkins if you please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2646/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2646/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2646/

craigcomstock:
travis failed with a bunch of ```----------------------------------------------------------------------

./promises/promises.cf 

----------------------------------------------------------------------

/home/travis/build/cfengine/masterfiles/../core/tests/acceptance/testall: 1: eval: fakeroot: not found

Return code is 127.

  ==> FAIL```

cfe_internal/enterprise/federation/federation.cf
craigcomstock: Now I'm not sure. I tested again and it seems ssh-keygen works fine. `semanage` is an add-on package we would have to install with `yum -y install policycoreutils-python` which doesn't seem reasonable at all. I'll test again on centos6 with the latest nightly and possibly close this PR. :+1:

vpodzime:
> The .ssh dir ends up with system_u:object_r:usr_t:s0 in the case where it failed for me and when policy runs by itself is it running as something other than unconfined_u because of running from systemd/cf-exec right?

The label of the file is determined by the *fcontext* rules, it doesn't matter what created the file, AFAIK.

vpodzime:
> Another problem here was that there is no rule for .ssh in home dirs that are not of the pattern /home/*/.ssh

Oh, I thought this was handled in some more clever way like `useradd` adding the rule or something like that.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2664/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2664/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2664/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2665/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2665/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2665/

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -195,16 +195,17 @@ bundle agent transport_user
       edit_line => default:insert_lines("IdentityFile $(ssh_priv_key)");
 
   commands:
+    selinux_enabled.missing_ssh_context::
```
vpodzime: `missing_ssh_context` is based on contexts of the **files** in the directory, not on the context of the directory itself, AFAICT.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -195,16 +195,17 @@ bundle agent transport_user
...
+      "/usr/bin/chcon -v --type ssh_home_t $(home)/.ssh/"
```
vpodzime: Would be better to add a new *fcontext* rule because this will be reverted by relabeling otherwise.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -195,16 +195,17 @@ bundle agent transport_user
...
+        contain => default:in_shell;
```
vpodzime: I don't think this has to run in a shell.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2667/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2667/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2667/

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -193,18 +200,23 @@ bundle agent transport_user
       create => "true",
       handle => "ssh_config_configured",
       edit_line => default:insert_lines("IdentityFile $(ssh_priv_key)");
+    "$(ssh_home_cil_file)"
+      create => "true",
+      edit_line => default:insert_lines('(filecon "$(home)/.ssh" any (system_u object_r ssh_home_t ((s0)(s0))))');
```
vpodzime: Why not just `semanage fcontext -a`?

PR #1480, opened 2019-08-09
Fix the noise from attempts to parse lldpctl's stderr as json

mikeweilgart:
See https://tracker.mender.io/browse/CFE-3103 for background.

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -829,7 +829,7 @@ bundle agent cfe_autorun_inventory_LLDP
 
   vars:
     !disable_inventory_LLDP.lldpctl_exec_exists::
-      "info" data => parsejson(execresult($(inventory_control.lldpctl_json), "noshell"));
+      "info" data => parsejson(execresult("$(inventory_control.lldpctl_json) 2>/dev/null", "useshell"));
 }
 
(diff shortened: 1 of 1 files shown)


cf-bottom:

mikeweilgart:
Actually, I will add that *nothing* in policy uses this information, and it is not tagged for collection.  I recommend turning it off by default.  It's just overhead and it can't be used for anything without policy being written for that.  If someone is writing policy for that, they will know it, and can explicitly add some class to enable this code.

nickanderson:
Does the methodology described in CFE-3109 not work for you?
https://tracker.mender.io/browse/CFE-3109

mikeweilgart:
I don't understand why this should be enabled by default.  Literally nothing in policy uses it, and it's not tagged for collection by the Hub, so if a command is run in the forest but nothing depends on it and the output isn't collected, then why not just leave it disabled by default to avoid the performance overhead of falling trees?  (*We've* disabled it now in our code using def.json to define the disabling class, but I don't think that should be necessary.  The default should be for it NOT to run.)

nickanderson:
Superceeded by https://github.com/cfengine/masterfiles/pull/1493

PR #1484, opened 2019-08-14
Make sure superhub trusts all SSH keys provided by its feeders

vpodzime:
SSH generates host keys of multiple types in modern systems. We
need to make sure all host keys from each feeder are trusted by
the superhub because only the mutual negotiation between the
superhub and each feeder determines which type of the key is used
(and thus needs to be trusted).

Ticket: ENT-4917
Changelog: Made keys of all types from feeder hubs trusted on a superhub

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -57,11 +57,11 @@ bundle agent config
     am_superhub|am_feeder::
       "pubkeys" slist => getvalues( pubkey );
 
-      "fingerprint[$(remotes)]"
-        string => "$(data[remote_hubs][$(remotes)][transport][ssh_host]) $(data[remote_hubs][$(remotes)][transport][ssh_fingerprint])",
+      "fingerprint[$(data[remote_hubs][$(remotes)][transport][ssh_host])]"
+        slist => string_split("$(data[remote_hubs][$(remotes)][transport][ssh_fingerprint])", "$(const.n)", "inf"),
         # To ensure we are using a remote hub that's enabled
         if => strcmp( "on", "$(data[remote_hubs][$(remotes)][target_state])" );
-      "fingerprints" slist => getvalues( fingerprint );
+      "fingerprints" slist => maparray("$(this.k) $(this.v)", fingerprint);
       "feeder[$(remotes)]" string => "$(data[remote_hubs][$(remotes)][hostkey])",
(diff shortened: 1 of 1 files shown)


vpodzime:
Hmm, there is a new error, this time on the feeders:
> 52.19.202.140	   error: No template data for the promise '/opt/cfengine/federation/cftransport/.ssh/known_hosts'

vpodzime:
> Hmm, there is a new error, this time on the feeders:
> 
> > 52.19.202.140	   error: No template data for the promise '/opt/cfengine/federation/cftransport/.ssh/known_hosts'

Fixed.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -57,11 +57,11 @@ bundle agent config
     am_superhub|am_feeder::
       "pubkeys" slist => getvalues( pubkey );
 
-      "fingerprint[$(remotes)]"
-        string => "$(data[remote_hubs][$(remotes)][transport][ssh_host]) $(data[remote_hubs][$(remotes)][transport][ssh_fingerprint])",
+      "fingerprint[$(data[remote_hubs][$(remotes)][transport][ssh_host])]"
+        slist => string_split("$(data[remote_hubs][$(remotes)][transport][ssh_fingerprint])", "$(const.n)", "inf"),
```
nickanderson: I'm a bit surprised there aren't issues here with lists inside of classic arrays. That was problematic in recent history. To avoid that, I have used tagging and then find/collect variables with variablesmatching

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -57,11 +57,11 @@ bundle agent config
...
         # To ensure we are using a remote hub that's enabled
         if => strcmp( "on", "$(data[remote_hubs][$(remotes)][target_state])" );
-      "fingerprints" slist => getvalues( fingerprint );
+      "fingerprints" slist => maparray("$(this.k) $(this.v)", fingerprint);
```
nickanderson: Nice

cfe_internal/enterprise/federation/federation.cf
nickanderson: Did you test this against old versions?

vpodzime: No, but I don't care about anything older than 3.12.2.

PR #1486, opened 2019-08-23
ENT-4937: Upgrade MPF from 3.12.x to 3.14.x fails on cf-promises

craigcomstock:
Need to specify path relative to this.promise_dirname instead of
sys.inputdir.

Changelog: None


nickanderson:
LGTM!

craigcomstock:
If folks need to upgrade MPF for federation let's have them use 3.15.0.

PR #1493, opened 2019-09-03
CFE-3109 Suppressed stderr output from lldpctl when using nonstandard path

nickanderson:
This change suppresses output on stderr during lldpctl inventory when using
lldpctl from an expected path. When lldpctl is inventoried by the executable
identified by def.lldpctl_json (a user supplied value, likely defined via
augments) we do not execute with shell.

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -829,7 +829,7 @@ bundle agent cfe_autorun_inventory_LLDP
 
   vars:
     !disable_inventory_LLDP.lldpctl_exec_exists::
-      "info" data => parsejson(execresult($(inventory_control.lldpctl_json), "noshell"));
+      "info" data => parsejson(execresult("$(inventory_control.lldpctl_json) 2>/dev/null", "useshell"));
 }
 
 bundle agent cfe_autorun_inventory_packages
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -829,7 +829,19 @@ bundle agent cfe_autorun_inventory_LLDP
 
   vars:
     !disable_inventory_LLDP.lldpctl_exec_exists::
-      "info" data => parsejson(execresult("$(inventory_control.lldpctl_json) 2>/dev/null", "useshell"));
+      # TODO When CFE-3108 is DONE, migrate to capturing only stdout
+      "info" -> { "CFE-3109", "CFE-3108" }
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
Superceeds: https://github.com/cfengine/masterfiles/pull/1480

mikeweilgart:
Thanks!  I don't understand how it's "for safety" that stderr is read in (and assumed to be JSON) if the user has customized the path.  Also, that means that when CFE-3108 is fixed, the behavior of the policy on custom values will be changed.

I submit that there is never a case where it would make sense to assume that the stderr of a user-defined lldpctl path will be valid json.

This PR is still an improvement to the status quo, since at least the default policy without modifications won't make noise.

nickanderson:
The data fed into the agent is less likely to have any kind of peer review unlike a custom script being added.

And when cfe-3108 is done we can transition to just throwing away stderr all the time for this promise.

mikeweilgart:
> And when cfe-3108 is done we can transition to just throwing away stderr all the time for this promise.

That's my point.  The behavior of the code in this PR is that stderr will NOT be thrown away for a custom script that's added.  But the plan is to *change* that behavior once CFE-3108 is complete, so that the stderr WILL be thrown away, even for a custom script that's added.

I believe it would be cleaner to just throw away the stderr right now, so that the planned change once CFE-3108 is complete will be a refactor only, and will not change the behavior.

It would also make the code simpler right now, since it would be only one promise instead of two.

vpodzime:
> I don't understand how it's "for safety" that stderr is read in (and assumed to be JSON) if the user has customized the path.

It's for safety that an arbitrary string given in augments is not run in a shell. And if a customized path is used, it's very easy to use a script that discards *stderr* from `lldpctl`.

mikeweilgart:
> It's for safety that an arbitrary string given in augments is not run in a shell. And if a customized path is used, it's very easy to use a script that discards _stderr_ from `lldpctl`.

I still don't understand the attack model or possible mistake that this guards against.  If malice is intended, it is just as simple to have the arbitrary string in augments name a script, which in turn can contain arbitrary strings which will then be run in a shell.  And if you mean it's for safety in the sense of preventing mistakes, could you please give an example?

I can only think of examples where the user would be *expecting* the command to be interpreted by a shell, and so would end up surprised when e.g. redirection didn't work as intended.

PR #1495, opened 2019-09-04
CFE-3109/3.10.x: Suppressed stderr output from lldpctl when using path defined by def.lldpctl_json

nickanderson:


cf-bottom:

PR #1500, opened 2019-09-11
ENT-4906: fix use of _stdlib_path_exists_<command> in FR transport_user bundle

craigcomstock:
Changelog: title


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -230,6 +229,11 @@ bundle agent transport_user
         if => and( isdir( "$(home)/.ssh" ),
                    not( fileexists( "$(ssh_priv_key)" )));
 
+  reports:
```
craigcomstock: On further consideration it seems like we might fail some promises if these commands are not present. So that a user is more likely to see the problem?

PR #1501, opened 2019-09-13
ENT-4944/3.12.x: Fixed synchronization of important configuration files from active to passive hub

nickanderson:


cf-bottom:

PR #1502, opened 2019-09-13
CFE-3150/master Fixed re-spawning of cf-execd or cf-monitord after remediating duplicate concurrent processes

nickanderson:
As long as refresh_processes in body agent control is null (which it is in the
MPF by default) cf-agent refreshes the process table each time a bundle is
entered. Re-spawning failed to work because the promises to spawn cf-execd or
cf-monitord were based on direct inspection of the process table, and in the
case of remediation being necessary, there were more than the expected number of
processes running.

This change separates those promises to re-spawn daemons after remediation to a
separate bundle so that the promises will be inspecting a fresh version of the
process table.

### DIFF
--- a/cfe_internal/core/limit_robot_agents.cf
+++ b/cfe_internal/core/limit_robot_agents.cf
@@ -6,6 +6,26 @@
 ###############################################################################
 
 bundle agent cfe_internal_limit_robot_agents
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #1507, opened 2019-09-17
CFE-3153/master Fixed pkgsrc in case where multiple Prefix paths are returned for pkg_install

nickanderson:


nickanderson:
@joseyluis I ran into issues trying to validate your removal of `-u 0` on NetBSD 8. I don't know how correct this is, but only taking the first path from Prefix fixed it for me.

@bahamat you may also have an opinion on this nieve fix.

joseyluis:
The same happens to me with PKG_PATH. (NetBSD and Solaris also).
I wanted to solve it with other pull, but if you have fixed it, it is better.

When I test this I fixed the PKG_PATH assignment.

I have other pending fixes for the pkgsrc module: latest doesn't work, etc.

nickanderson:
@joseyluis where is PKG_PATH? What was your fix?

joseyluis:
I think your patch is ok.

bahamat:
@nickanderson Let me check with someone who knows it better than I do. I'd be more inclined to do `sort -u` to eliminate duplicates. I'm concerned about if anything would need to be done if there are multiple but different.

nickanderson:
> @nickanderson Let me check with someone who knows it better than I do. I'd be more inclined to do `sort -u` to eliminate duplicates. I'm concerned about if anything would need to be done if there are multiple but different.

With `sort -u` i guess it would fail if there were still more than one entry, maybe that would be the desired behavior.

bahamat:
I think it's self evident that if `sort -u` returns a single directory that is OK.

We're only using PKG_ROOT to find `pkg_install.conf`. And we're only checking the `PKG_ROOT` of the `pkg_install` package. If it were to happen that `pkg_install` somehow has two roots, and we're unable to locate `pkg_install.conf`, or perhaps pick the wrong one could cause severe harm to the system. So there's a very narrow set of conditions where the behavior would change. Namely, `pkg_install` would need to have some 

That being the case, if there's more than one I think we should error out. This might be inconvenient for a user, but when the options are "safe" vs "maybe installed malware on your system", I would bias toward safe. When/if that condition ever comes up we can figure out then what needs to be done to safely handle the situation.

nickanderson:
Agreed, thanks @bahamat !

PR #1517, opened 2019-09-20
Travis: Added valgrind bootstrap test from core

olehermanse:


travis-scripts/valgrind.sh
```diff
@@ -0,0 +1,218 @@
...
+cd ../
+
+if [ ! -d core ]; then
+    echo "Cloning core (master)"
+    git clone --recursive https://github.com/cfengine/core.git
+fi
+
+if [ ! -d masterfiles ]; then
+    echo "Cloning masterfiles (master)"
+    git clone --recursive https://github.com/cfengine/masterfiles.git
```
nickanderson: cloning whole repo, not shallow?

olehermanse: `--recursive` includes submodules. This part is only included because the script also runs in core travis.

PR #1516, opened 2019-09-20
Revert "Fixed agent disabling on systemd systems"

olehermanse:
Reverts cfengine/masterfiles#1511

PR #1520, opened 2019-09-21
Fix disabling agents on systemd managed hosts

nickanderson:
Reverts cfengine/masterfiles#1516


cf-bottom:

PR #1556, opened 2019-10-10
ENT-5005 Migrate from DO to AWS

Lex-2008:


cf-bottom:

Lex-2008:
> It doesn't seem correct to change from p.arch to hard coded amd64. What was the value of p.arch?

it's the other way around :-)

If we look at [named_multi_pkg_present.cf][] - we'll see that in `test` it installs packages by name:

```
bundle agent test
{
    vars:
        "package_name" slist => { "$(p.name[3])","$(p.name[1])","$(p.name[2])"};
    packages:
        debian::
            "$(package_name)"
```
where `p.name[…]` is `test-package-{alpha,beta,charlie}` defined in [packages-info.cf.sub][].

and in `check` it tests that *64-bit* version of the package is installed:
```
bundle agent check
{
    classes:
        debian::
            "has_pkg_$(test.package_name)" expression => returnszero("dpkg -l | egrep ' $(test.package_name)(:amd64)? '", "useshell");
```

Which is clearly wrong on *32-bit* OSes. This PR fixes it.

[named_multi_pkg_present.cf]: https://github.com/cfengine/masterfiles/blob/master/tests/acceptance/17_packages/11_old/unsafe/named_multi_pkg_present.cf
[packages-info.cf.sub]: https://github.com/cfengine/masterfiles/blob/master/tests/acceptance/17_packages/packages-info.cf.sub#L11

PR #1558, opened 2019-10-21
Only report on skipped upgrade if trigger_upgrade is set

nicholasa78:
While prepping for a client upgrade, we found this to be too noisy. Before an upgrade attempt, at_desired_version was never set so this never triggered. Once a host upgraded and the standalone_self_upgrade created the json file that provided a defnition for this variable, we see it every time we run update.cf in inform mode.

As a workaround, we changed the code to only provide the report if the trigger_upgrade class is set AND at_desired_version

Mike Weilgart and Aleksey Tsalolikhin asked me to submit this change as an upstream pull request

### DIFF
--- a/update.cf
+++ b/update.cf
@@ -135,7 +135,7 @@ bundle agent cfengine_internal_standalone_self_upgrade_execution
 
   reports:
 
-    inform_mode|verbose_mode|DEBUG|DEBUG_cfengine_internal_standalone_self_upgrade::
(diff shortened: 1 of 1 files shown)


nickanderson:
We're you running upgrades with inform, verbose or debug?

I think the trigger upgrade condition should be anded to the existing condition. Can you please make that change and squash the commits?

nicholasa78:
That's a good point. I modified it so we're just adding the trigger_upgrade class to the conditions for the report.

nickanderson:
Thanks!

PR #1564, opened 2019-10-29
Travis: Sleep 10 more seconds after killing cf-serverd

olehermanse:
This test failed 2/5 times in the recent master builds,
because cf-serverd didn't exit in time. Increasing the
sleep from 30 to 40 seconds, to see if this helps.

We could probably work on a system which sleeps 10
seconds at a time, and checks if the process is
still alive.

### DIFF
--- a/travis-scripts/valgrind.sh
+++ b/travis-scripts/valgrind.sh
@@ -193,7 +193,7 @@ echo "Killing valgrind cf-execd"
 kill $exec_pid
 echo "Killing valgrind cf-serverd"
 kill $server_pid
-sleep 30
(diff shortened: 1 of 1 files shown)


olehermanse:
I'll test this at least 5 times in travis:

- [ ] Pass
- [ ] Pass
- [ ] Pass
- [ ] Pass
- [ ] Pass

olehermanse:
@vpodzime @nickanderson this passed 5 times in a row. (But it might be that load is worse later in the day). Let's merge and watch the nightlies through the week?

PR #1567, opened 2019-10-29
Fixed issues where stderr are piped to stdout instead of /dev/null

oleorhagen:
Follow up of the previous PR...

Same fixups, be sure to write stderr to where ( I think it is intended). Be sure to look it over though. Also, why are we appending to /dev/null?

### DIFF
--- a/travis-scripts/valgrind.sh
+++ b/travis-scripts/valgrind.sh
@@ -47,9 +47,9 @@ systemctl --version
 cd core/
 echo "Building CFEngine core"
 set +e
-git fetch --unshallow 2>&1 >> /dev/null
+git fetch --unshallow > /dev/null 2>&1
 git remote add upstream https://github.com/cfengine/core.git  \
-    && git fetch upstream 'refs/tags/*:refs/tags/*' 2>&1 >> /dev/null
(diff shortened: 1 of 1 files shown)


travis-scripts/valgrind.sh
```diff
@@ -47,9 +47,9 @@ systemctl --version
 cd core/
 echo "Building CFEngine core"
 set +e
-git fetch --unshallow 2>&1 >> /dev/null
+git fetch --unshallow >> /dev/null 2>&1
 git remote add upstream https://github.com/cfengine/core.git  \
-    && git fetch upstream 'refs/tags/*:refs/tags/*' 2>&1 >> /dev/null
+    && git fetch upstream 'refs/tags/*:refs/tags/*' >> /dev/null 2>&1
```
vpodzime: Might as well change these to be `> /dev/null`.

olehermanse:
> Also, why are we appending to /dev/null?

@oleorhagen My guess is that they were copy pasted from somewhere else, and the original command from way back when was appending to a log file. Somebody changed it to `/dev/null`, but didn't change the redirection operator.

oleorhagen:
Bit disappointing actually, I was hoping for some wizard revelation here from someone. Real neckbeard trick, and this is it :p

PR #1578, opened 2019-11-12
CFE-3193/3.12.x: Excluded Enterprise federation policy parsing on incompatible versions

nickanderson:
The federated reporting in CFEngine Enterprise leverages features that are
unavailable in versions older than 3.12.0. This change prevents the federation
policy from being parsed and generating errors on older versions (both from
including policy containing features not present, and from specifying a bundle
that is not present when the federation policy is not parsed.

3.10.x was not targeted for support because Federated Reporting was targeted for
supported use starting in 3.15.0 as 3.10.x ended it's standard support life.

### DIFF
--- a/cfe_internal/enterprise/CFE_hub_specific.cf
+++ b/cfe_internal/enterprise/CFE_hub_specific.cf
@@ -14,6 +14,7 @@
 #
 ##################################################################
 
+@if minimum_version(3.12.0)
 bundle common cfe_hub_specific_file_control
 {
   vars:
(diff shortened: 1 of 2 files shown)


cf-bottom:

nickanderson:
@vpodzime 3.7.x doens't know how ot do minimum version with patch numbers, so I had to back off to 3.12 from 3.12.0. I don't intend to forward port this difference to master/3.15 unless someone thinks I should.

PR #1591, opened 2019-11-20
ENT-5077: Switched to using semanage for cftransport user on selinux systems (3.12.x)

craigcomstock:


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -129,6 +129,44 @@ bundle agent config
...
 }
 
+bundle agent semanage_installed
+# @brief Install semanage utility if selinux enabled and
+# cfengine_mp_fr_dependencies_auto_install class is defined
+# if not defined then only warn
+{
+  vars:
+    "semanage_action"
+      string => ifelse( "default:_stdlib_path_exists_semanage", "fix", "default:cfengine_mp_fr_dependencies_auto_install", "fix", "warn" ),
```
nickanderson: Style nitpick: I would break the line after "fix", (in both cases).

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -129,6 +129,44 @@ bundle agent config
...
+      comment => "We only want to use semanage if it's available, or if we have
+                  indicated it's ok to install it automatically. This variable
+                  is subsequently used by a commands and packages promises to
+                  warn or fix based.";
+    debian_6|debian_7|debian_8|ubuntu_12|ubuntu_14|ubuntu_16|rhel_5::
+      "semanage_package" string => "policycoreutils";
+    debian_9|debian_10|ubuntu_18|redhat_8|centos_8::
+      "semanage_package" string => "policycoreutils-python-utils";
+    redhat_6|centos_6|redhat_7|centos_7::
+      "semanage_package" string => "policycoreutils-python";
```
nickanderson: It might be more readable in a single promise using ifelse()

```
"semanage_pacakge" string => ifelse( "debian_6|debian_7|debian_8|ubuntu_12|ubuntu_14|ubuntu_16|rhel_5", "policycoreutils",
"redhat_6|centos_6|redhat_7|centos_7", "policycoreutils-python",
"policycoreutils-python-utils"); # Seems like the newest releases all use this package name, so this /might/ keep working for debian 11 and centos 9 whenever they come out.
```

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -129,6 +129,44 @@ bundle agent config
...
+
+  packages:
+    debian|ubuntu::
+      "$(semanage_package)"
+        policy => "present",
+        package_module => default:apt_get,
```
nickanderson: Really? You should not have to specify the package module. At least you shouldn't have to if you are running with the full policy set. If we are letting people run this directly, then I think this must be specified as you have done.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -129,6 +129,44 @@ bundle agent config
...
+      "$(semanage_package)"
+        policy => "present",
+        package_module => default:yum,
+        action => default:policy ( $(semanage_action) );
+
+  reports:
+    default:DEBUG|DEBUG_semanage_installed::
+      "paths.semanage = $(paths.semanage)";
+    !default:_stdlib_path_exists_semanage.!default:cfengine_mp_fr_dependencies_auto_install::
+      "semanage command is not available at $(paths.semanage). Will only install needed package if cfengine_mp_fr_dependencies_auto_install class is defined in augments(def.json) or with --define cf-agent option.";
```
nickanderson: Your so nice with these reports

lib/paths.cf
```diff
@@ -360,7 +360,7 @@ bundle common paths
       "path[printf]"        string => "/usr/bin/printf";
       "path[restorecon]"    string => "/sbin/restorecon";
       "path[sed]"           string => "/bin/sed";
-      "path[semodule]"      string => "/usr/sbin/semodule";
```
nickanderson: Shouldn't we keep semodule in paths, just add semanage?

cfe_internal/enterprise/federation/federation.cf
craigcomstock: awesome! Will make that change in a follow-up!

lib/paths.cf
craigcomstock: will add back in with a follow up. :+1:

PR #1598, opened 2019-12-02
Fixup ini management

mikeweilgart:
See commit messages for details.


cf-bottom:

nickanderson:
I don't see any issue offthe top of my head, but we should add a test for this before it's merged.

If you want to take a shot at it checkout https://github.com/cfengine/masterfiles/blob/master/tests/acceptance/lib/files/manage_variable_values_ini.cf

mikeweilgart:
Thanks, I won't have time to write an acceptance test for this, but if it makes any difference I tested with our internal code (what I needed this change for) and it works as expected.  :)  I know that's not the same as automated testing.

nickanderson:
Closing in favor of https://github.com/cfengine/masterfiles/pull/1609

Thanks!

PR #1599, opened 2019-12-03
Only check the 'semanage' package if SELinux is enabled (in FR)

vpodzime:
Otherwise the promise is not kept and blocks the setup from being
completed even if SELinux is disabled and the 'semanage' package
is not actually needed at all.

Ticket: ENT-5124
Changelog: none

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -258,7 +258,8 @@ bundle agent transport_user
       edit_line => default:insert_lines("IdentityFile $(ssh_priv_key)");
 
   methods:
-    "semanage_installed" usebundle => semanage_installed;
+    selinux_enabled::
+      "semanage_installed" usebundle => semanage_installed;
 
   commands:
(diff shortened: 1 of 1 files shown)


vpodzime:
Triggered a build in jenkins with FR tests only:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/4243/)

vpodzime:
> I agree with this change though wonder why we didn't catch it in fr-tests.
This is what the FR tests do:
```
"/var/cfengine/bin/cf-agent --define cfengine_mp_fr_dependencies_auto_install -KI"
```

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -258,7 +258,8 @@ bundle agent transport_user
       edit_line => default:insert_lines("IdentityFile $(ssh_priv_key)");
 
   methods:
-    "semanage_installed" usebundle => semanage_installed;
+    selinux_enabled::
```
nickanderson: Makes sense.

PR #1601, opened 2019-12-05
ENT-3153: cf-hub query timeout can now be set from augments

olehermanse:
Variable `control_hub_query_timeout`.

Merge together:
https://github.com/cfengine/core/pull/3964
https://github.com/cfengine/masterfiles/pull/1601
https://github.com/cfengine/nova/pull/1560

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -889,6 +889,29 @@ For example:
 
 - Added in 3.13.0, 3.12.2
 
+### Change hub to client connection timeout
+
+By default, cf-hub times out a connection after 30 seconds.
+This can be configured in augments.
+
+For example:
+
+```json
+{
+  "vars": {
+    "control_hub_query_timeout": "10"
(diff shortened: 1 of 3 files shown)


MPF.md
```diff
@@ -889,6 +889,29 @@ For example:
...
+{
+  "vars": {
+    "control_hub_query_timeout": "10"
+  }
+}
+```
+
+**Note:**
+
+- A value of `"0"` will cause the default to be used.
```
nickanderson: `"0"` seems a bit odd? why literal around the quotes?

olehermanse: @nickanderson I don't know. In CFEngine, I like to show that strings are strings (`0` is a valid bundle / body name), so to me it made sense to write it like that.

nickanderson: k

PR #1608, opened 2019-12-08
Added inline docs in places they were missing

nickanderson:


Lex-2008:
Looks good!

But wouldn't you prefer see all commits squashed?

At least "Correced spelling error" and "Fix spelling" could be squashed :-)
Same goes for "fix param lines" and "added lines paramater description" which goes right after it.

Commit message "Describe control attribute for cf-serverd" goes out of line compared to others - I'd expect it to say "Added description for body server control"

nickanderson:
This was just something I was working on over the course of several sessions, I will squash all into a single commit "Added missing inline docs and fixed minor spelling errors" if no one objects.

PR #1609, opened 2019-12-08
CFE-3221/master: Made set_variable_values_ini prefer whitespace around =

nickanderson:
- Added test that leading whitespace on the RHS is not removed

Ticket: CFE-3221
Changelog: None

----

#

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -481,12 +481,12 @@ bundle edit_line set_variable_values_ini(tab, sectionName)
       # If the line is there, but commented out, first uncomment it
       "#+\s*$(index)\s*=.*"
       select_region => INI_section(escape("$(sectionName)")),
-      edit_field => col("=","1","$(index)","set"),
+      edit_field => col("\s*=\s*","1","$(index)","set"),
       ifvarclass => "edit_$(cindex[$(index)])";
 
       # match a line starting like the key something
       "$(index)\s*=.*"
-      edit_field => col("=","2","$($(tab)[$(sectionName)][$(index)])","set"),
+      edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
       select_region => INI_section(escape("$(sectionName)")),
       classes => results("bundle", "set_variable_values_ini_not_$(cindex[$(index)])"),
       ifvarclass => "edit_$(cindex[$(index)])";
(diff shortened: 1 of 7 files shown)


nickanderson:
Ref: https://github.com/cfengine/masterfiles/pull/1598

lib/files.cf
```diff
@@ -426,7 +426,7 @@ bundle edit_line manage_variable_values_ini(tab, sectionName)
 
       # match a line starting like the key something
       "$(index)\s*=.*"
-      edit_field => col("=","2","$($(tab)[$(sectionName)][$(index)])","set"),
+      edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
```
nickanderson: @mikeweilgart This does not seem to have an effect on the test.

lib/files.cf
```diff
@@ -483,7 +483,7 @@ bundle edit_line set_variable_values_ini(tab, sectionName)
...
```
nickanderson: I can see this resulted in the RHS having a space after the = https://travis-ci.org/cfengine/masterfiles/jobs/622422779#L1247

tests/acceptance/lib/files/manage_variable_values_ini.cf.start
```diff
@@ -1,5 +1,9 @@
 [section]
 keyone=valueone
 keytwo=value2
+[CFE-3221]
+variablename= desired value
+variablename2 = desired value2
```
nickanderson: These lines in the managed section have spaces after the = when the test starts.

lib/files.cf
nickanderson: @mikeweilgart I believe that these changes to `manage_variable_values_ini` have no effect since this bundle manages the complete contents of the section.

This [delete_lines promise](https://github.com/cfengine/masterfiles/pull/1609/files#diff-aaead9adbc8b4f10c3e8dc460612d0ccR434-R437) will empty the section first, then `insert_lines` will populate the entire section and there will never be any opportunity for the `field_edits` to make any adjustments.

You said you tested in your environment, did you `test manage_variable_values_ini`, or `just set_variable_values_ini`?

I am inclined to redact the changes to `manage_variable_values_ini` since they have no effect.

mikeweilgart: I only tested `set_variable_values_ini`.  However, the *only* difference between the two bundles is the `delete_lines` promise in one of them.  I think it would be strange to let them diverge in other respects.

nickanderson: thanks for clarifying that you tested set_variable_values_ini. I am redacting the changes to manage_variable_values_ini and I will squash this down into 2 commits. One extending the test to show the behavior before the change, and then your changes along with updates to the test.

PR #1618, opened 2019-12-16
ENT-5168 FR: Don't require SELinux Makefile for running semanage / restorecon

olehermanse:


olehermanse:
3.15 PR: https://github.com/cfengine/masterfiles/pull/1619

PR #1633, opened 2019-12-20
Spellfix

oleorhagen:
I think these are wrong.

Dearest Ole#1

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -519,8 +519,8 @@ bundle edit_line insert_ini_section(name, config)
 {
   vars:
       # TODO: refactor once 3.7.x is EOL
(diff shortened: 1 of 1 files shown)


oleorhagen:
@olehermanse

lib/files.cf
```diff
@@ -519,8 +519,8 @@ bundle edit_line insert_ini_section(name, config)
 {
   vars:
       # TODO: refactor once 3.7.x is EOL
```
nickanderson: you fix the spelling, and you skip the TODO?

oleorhagen: So the naming was the refactoring needed? xD
Sry, missed that one.
@nickanderson If that is what's needed I can fix it.

PR #1637, opened 2019-12-23
Allow whitespace at the beginning of ini key-values

oleorhagen:
Previously the {manage,set}_variable_values_ini did not match on any ini files
containing whitespace at the beginning of a line, hence

[section foo]
key=value

could be handled. However, if the file looks like

[section bar]
  key=value

then the previous regex would not match.

What I can get from the INI spec, is that whitespace is optional. This commit
adds this to the regex, with an optional whitespace match at the beginning of
the line.

Changelog: Title

Signed-off-by: Ole Petter <ole.orhagen@northern.tech>

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -428,7 +428,7 @@ bundle edit_line manage_variable_values_ini(tab, sectionName)
       ifvarclass => "edit_$(cindex[$(index)])";
 
       # match a line starting like the key something
-      "$(index)\s*=.*"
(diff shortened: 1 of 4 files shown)


oleorhagen:
Mind you, I don't know if this was intentional or not.

oleorhagen:
Superseeded by https://github.com/cfengine/masterfiles/pull/1651

PR #1640, opened 2020-01-06
Make the 'selinux_enabled' in FR policy class namespace-scoped

vpodzime:
It's used outside of the bundle in which it is defined.


vpodzime:
Superseded by #1641.

PR #1642, opened 2020-01-07
Change logic for dmi inventory to prefer sysfs, except for UUID

mikeweilgart:
Changelog: Fixed system UUID inventory for certain VMWare VMs where dmidecode gives UUID bytes in wrong order.

Changelog: Fixed dmi inventory to prefer sysfs to dmidecode for most variables for improved performance and to handle e.g. CoreOS hosts that don't have dmidecode.

Using sysfs has two advantages over using dmidecode:

1. It's faster; no need to start a process (actually two processes,
sh -c 'dmidecode -s ...' and the dmidecode process itself)
for each value we want to collect.

2. It makes it possible to collect DMI info on hosts without
dmidecode, such as CoreOS hosts.

When present (it's not present e.g. on RHEL 5 hosts),
sysfs is at least as reliable as using dmidecode directly.

However, the UUID is returned incorrectly on some VMWare VMs
(those with hardware version 13) - the byte order is swapped around.

See:

- https://kb.vmware.com/s/article/53609
- https://kb.vmware.com/s/article/1880
- https://github.com/vmware/pyvmomi/issues/688

Some such affected VMs show the same wrong UUID in sysfs and some show
the correct UUID in sysfs and the wrong UUID only in dmidecode.

The workaround for hosts showing the wrong uuid (byte order swapped)
is to use an Awk script that parses raw dmidecode output
(script given in the first link above).  In this commit,
the use of the awk script is not limited to affected VMWare VMs because
(a) there is no clearcut and easy way to identify affected hosts and
(b) the awk script works on all hosts anyway - it is strictly more reliable
than "dmidecode -s system-uuid".
Also, (c) the awk script even works on hosts that have an older version
of dmidecode that doesn't support the "system-uuid" keyword
(e.g. dmidecode-2.7-1.28.2.el5).

The fallback to get the uuid from sysfs is still needed for hosts
that have sysfs but not dmidecode (such as CoreOS).

In summary:

1. Hosts without dmidecode (which formerly would not have dmi inventory at all)
   will now have dmi inventory taken from sysfs.  (E.g. coreos hosts)

2. Hosts currently showing the wrong UUID or an error message in place
   of a UUID will now show the correct UUID via dmidecode and Awk.
   (E.g. VMWare 13 VMs, or RHEL 5 hosts with dmidecode-2.7-1.28.2.el5)

3. Other hosts will start pulling their dmi inventory from sysfs instead
   of from "dmidecode -s ...", except UUID which they will pull from
   "dmidecode -u -t1 | awk ...".  The values will all remain the same.
   Performancewise, pulling from sysfs will be strictly faster than
   pulling from "dmidecode -s ...", which should offset the negligible
   impact from using the awk script for collection of one value (uuid).

(I also cleaned up the comment list of "other values you may want to collect"
since system-uuid has been collected by default since
6935a10e2f5292929a45263e6ab6a1ec75d39796.)

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -724,7 +724,15 @@ bundle agent cfe_autorun_inventory_fstab
 bundle agent cfe_autorun_inventory_dmidecode
 # @brief Do hardware related inventory
 #
-# This agent bundle runs dmidecode when present to inventory:
+# This agent bundle reads dmi information from the sysfs and/or from dmidecode.
+# Sysfs is preferred for most variables, but if no sysfs (e.g. on RHEL 5),
+# or no sysfs equivalent to a dmidecode variable (e.g. system-version),
+# then dmidecode is run to collect the info.
+# For system-uuid, a parsed version of dmidecode is the preferred source.
+#
+# The variable names dmi[...] are all based on dmidecode string keywords.
+#
+# Information collected is:
 # - BIOS vendor
 # - BIOS version
 # - System serial number
@@ -739,7 +747,41 @@ bundle agent cfe_autorun_inventory_dmidecode
 # - System serial number
 # - System manufacturer
 {
+
   vars:
+    any::
+      "sysfs_name_for"
+        comment => "The names in /sys/devices/virtual/dmi/id/ don't match
+                    the strings to be passed to dmidecode, even though the
+                    values do.  We use the dmidecode string names for our
+                    variables since that was the original source (i.e. for
+                    backward compatibility with policies based on prior
+                    versions of this code).",
+        # system-version has no equivalent in sysfs that I can find.
+        # Items after the line break aren't currently collected, but mapping is provided
+        # in case someone adds them to a custom dmidefs (so that they could be gotten
+        # from sysfs in that case).
+        data => parsejson('
+          {
+            "bios-vendor": "bios_vendor",
+            "bios-version": "bios_version",
+            "system-serial-number": "product_serial",
+            "system-manufacturer": "sys_vendor",
+            "system-product-name": "product_name",
+            "system-uuid": "product_uuid",
+
+            "baseboard-manufacturer": "board_vendor",
+            "baseboard-product-name": "board_name",
+            "baseboard-serial-number": "board_serial",
+            "baseboard-version": "board_version",
+            "bios-release-date": "bios_date",
+            "chassis-manufacturer": "chassis_vendor",
+          }');
+
+  vars:
+    any::
+      # The dmidefs variable controls which values are collected
+      # (and what are their inventory tags)
       "dmidefs" data => parsejson('
 {
   "bios-vendor": "BIOS vendor",
@@ -772,7 +814,6 @@ bundle agent cfe_autorun_inventory_dmidecode
       # processor-family
       # processor-frequency
       # processor-manufacturer
-      # system-uuid
       #"processor-version": "CPU model" <- Collected by default, but not by iterating over the list
 
       "dmivars" slist => getindices(dmidefs);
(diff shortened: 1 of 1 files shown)


mikeweilgart:
Here is before and after on a RHEL 5 host with no sysfs and an old dmidecode:

Before:

 ```
      variablename        |                     variablevalue
---------------------------+--------------------------------------------------------
 dmi[bios-vendor]          | Phoenix Technologies LTD
 dmi[bios-version]         | 6.00
 dmi[system-manufacturer]  | VMware, Inc.
 dmi[system-product-name]  | VMware Virtual Platform
 dmi[system-serial-number] | VMware-42 11 e9 72 9d 58 b6 59-12 e2 7f 0e b6 52 ca 7d
 dmi[system-uuid]          | Invalid string keyword: system-uuid                   +
                           | Valid string keywords are:                            +
                           |   bios-vendor                                         +
                           |   bios-version                                        +
                           |   bios-release-date                                   +
                           |   system-manufacturer                                 +
                           |   system-product-name                                 +
                           |   system-version                                      +
                           |   system-serial-number                                +
                           |   baseboard-manufacturer                              +
                           |   baseboard-product-name                              +
                           |   baseboard-version                                   +
                           |   baseboard-serial-number                             +
                           |   baseboard-asset-tag                                 +
                           |   chassis-manufacturer                                +
                           |   chassis-version                                     +
                           |   chassis-serial-number                               +
                           |   chassis-asset-tag                                   +
                           |   processor-manufacturer                              +
                           |   processor-version
 dmi[system-version]       | None
 total_physical_memory_MB  | 2048
(8 rows)
```

After:

```
cfdb=> select variablename, variablevalue from variables where hostkey = 'SHA=9e6d151c7a2aa93966ee2b4777528b51c04b143ddef8338bebd01c712a333d90' and bundle = 'cfe_autorun_inventory_dmidecode' order by 1;
       variablename        |                     variablevalue
---------------------------+--------------------------------------------------------
 dmi[bios-vendor]          | Phoenix Technologies LTD
 dmi[bios-version]         | 6.00
 dmi[system-manufacturer]  | VMware, Inc.
 dmi[system-product-name]  | VMware Virtual Platform
 dmi[system-serial-number] | VMware-42 11 e9 72 9d 58 b6 59-12 e2 7f 0e b6 52 ca 7d
 dmi[system-uuid]          | 4211E972-9D58-B659-12E2-7F0EB652CA7D
 dmi[system-version]       | None
 total_physical_memory_MB  | 2048
(8 rows)
```

And before and after on a CentOS 7 host with VMWare hardware version 13:

Before:

```
 dmi[system-uuid]          | 1A9F3E42-E036-1668-E3A3-A4EC3F301F3C
```

After:

```
 dmi[system-uuid]          | 423E9F1A-36E0-6816-E3A3-A4EC3F301F3C
```

File: inventory/any.cf [hunk 1]
```diff
@@ -772,20 +814,57 @@ bundle agent cfe_autorun_inventory_dmidecode
...
+      # Awk script from https://kb.vmware.com/s/article/53609
+      # Edited only to add "-t1" (an improvement tested on RHEL 4/5/6/7 and FreeBSD)
+      # and to take out the "UUID: " prefix in the output.
+      # This works on a superset of systems where dmidecode -s system-uuid works,
+      # e.g. RHEL 5 with dmidecode-2.7-1.28.2.el5 where system-uuid is not one of the valid keywords;
+      # also, this returns the correct UUID on systems (such as VMWare VMs with hardware version 13)
+      # where dmidecode -s system-uuid shows the wrong UUID.  Some such VMWare VMs also show the
+      # wrong UUID in sysfs, which is why we prefer the "dmidecode | awk" version to sysfs for UUID.
+      # (We still need to check sysfs for UUID to handle hosts without dmidecode such as CoreOS.)
+      "dmi[system-uuid]"
```
vpodzime: I know the comment above this is already way too long, but one important piece is missing there -- and example of the output from `dmidecode -u -t1`. That would make the *AWK* script much easier to understand and review.

mikeweilgart: It won't make it as much easier as you think, and it's long enough that I don't think it belongs inline in that file, but here is an example.  It's easy to produce on any box with dmidecode anyway.

```
# dmidecode 3.0
Getting SMBIOS data from sysfs.
SMBIOS 2.5 present.

Handle 0x0001, DMI type 1, 27 bytes
	Header and Data:
		01 1B 01 00 01 02 03 04 88 20 6F B5 62 27 4B 7A
		83 1D 5C 58 59 B3 25 46 06 00 05
	Strings:
		69 6E 6E 6F 74 65 6B 20 47 6D 62 48 00
		"innotek GmbH"
		56 69 72 74 75 61 6C 42 6F 78 00
		"VirtualBox"
		31 2E 32 00
		"1.2"
		30 00
		"0"
		56 69 72 74 75 61 6C 20 4D 61 63 68 69 6E 65 00
		"Virtual Machine"

```

And the output of the script on the same box:

```
88206FB5-6227-4B7A-831D-5C5859B32546
```

The code is on the linked page, though.  I think that's the proper place where the documentation should be improved.

I've submitted feedback on the linked page saying:

> Using dmidecode -u -t1 makes for a lot less output for awk to parse.  Also, it makes the output size manageable enough that you could include an example output of dmidecode -u -t1, which would make the script much easier to understand.

vpodzime: Thanks for the example. It will at least be available here for the record. You're right it doesn't belong in that file.

nickanderson:
Actually, never mind. The developers corrected me. Multiple changelog entires are OK. but please add the ticket keyword.

mikeweilgart:
@nickanderson done; commit amended (only the message changed, to add the Ticket keyword) and force-pushed.  Thanks for the review!

nickanderson:
Thanks!

PR #1646, opened 2020-01-09
CFE-3244/master: Tolerate whitespace before key in manage_variable_values_ini

nickanderson:


oleorhagen:
Nice! So it switched them up, huh.

PR #1647, opened 2020-01-10
ENT-5235/master: Fixed permission flip flop when policy analyzer is enabled

nickanderson:


nickanderson:
Cherry Pick: https://github.com/cfengine/masterfiles/pull/1654

PR #1650, opened 2020-01-11
ENT-5233/master: Suppress useless inform output from /bin/true in ec2 inventory

nickanderson:
The commands promise /bin/true is used to delay execution of url_get until the
bundle is explicitly called (so that it doesn't get called and cached during
pre-eval). This change suppresses the useless inform output on versions 3.15.0
and later yet maintains backwards compatibility with 3.12 and 3.10 (where the
inform output can not be suppressed).

Ticket: ENT-5233
Changelog: Title


nickanderson:
I tested manually on 3.15 only. But since it passes travis for old versions i can tell that it wont blow them up and that the guard is working properly.

nickanderson:
@olehermanse adjusted names as requested

olehermanse:
Should be backported to 3.15.1 AFAICT.

nickanderson:
Cherry Pick: https://github.com/cfengine/masterfiles/pull/1655

PR #1651, opened 2020-01-13
CFE-3244: Tolerate whitespace before key in manage_variable_values_ini and set_variable_values_ini 

oleorhagen:
Updated with the fix added to `set_variable_values_ini` as well.

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -428,7 +428,7 @@ bundle edit_line manage_variable_values_ini(tab, sectionName)
       ifvarclass => "edit_$(cindex[$(index)])";
 
       # match a line starting like the key something
-      "$(index)\s*=.*"
+      "\s*$(index)\s*=.*"
       edit_field => col("=","2","$($(tab)[$(sectionName)][$(index)])","set"),
       select_region => INI_section(escape("$(sectionName)")),
       classes => results("bundle", "manage_variable_values_ini_not_$(cindex[$(index)])"),
@@ -485,7 +485,7 @@ bundle edit_line set_variable_values_ini(tab, sectionName)
       ifvarclass => "edit_$(cindex[$(index)])";
 
       # match a line starting like the key something
-      "$(index)\s*=.*"
+      "\s*$(index)\s*=.*"
       edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
       select_region => INI_section(escape("$(sectionName)")),
       classes => results("bundle", "set_variable_values_ini_not_$(cindex[$(index)])"),
--- a/tests/acceptance/lib/files/set_variable_values_ini.cf
+++ b/tests/acceptance/lib/files/set_variable_values_ini.cf
@@ -18,11 +18,11 @@ bundle agent init
   files:
 
       "$(G.testfile).actual"
-      copy_from => local_cp("$(this.promise_filename).start");
+        copy_from => local_cp("$(this.promise_filename).start");
 
 
       "$(G.testfile).expected"
-      copy_from => local_cp("$(this.promise_filename).finish");
+        copy_from => local_cp("$(this.promise_filename).finish");
 
 }
 
@@ -31,7 +31,7 @@ bundle agent init
 bundle agent test
 {
   meta:
-      "description" -> { "CFE-3221" }
+      "description" -> { "CFE-3221", "CFE-3244" }
         string => "Test expected behavior of set_variable_values_ini";
 
   vars:
@@ -47,13 +47,19 @@ bundle agent test
       "config[CFE-3221][variablename7]" string => "desired value7";
       "config[CFE-3221][variablename8]" string => "desired value8";
 
+      "config[CFE-3244][keyone]" string => "1";
+      "config[CFE-3244][keytwo]" string => "valuetwo";
+
   files:
 
       "$(G.testfile).actual"
-      edit_line => set_variable_values_ini("test.config", "section");
+        edit_line => set_variable_values_ini("test.config", "section");
 
       "$(G.testfile).actual" -> { "CFE-3221" }
         edit_line => set_variable_values_ini( "test.config", "CFE-3221" );
+
+      "$(G.testfile).actual" -> { "CFE-3244" }
+        edit_line => set_variable_values_ini( "test.config", "CFE-3244" );
 
 }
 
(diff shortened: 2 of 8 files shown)


lib/files.cf
```diff
@@ -428,7 +428,7 @@ bundle edit_line manage_variable_values_ini(tab, sectionName)
       ifvarclass => "edit_$(cindex[$(index)])";
 
       # match a line starting like the key something
-      "$(index)\s*=.*"
+      "\s*$(index)\s*=.*"
       edit_field => col("=","2","$($(tab)[$(sectionName)][$(index)])","set"),
```
oleorhagen: @nickanderson there is an inconsistency here (or at least I think so). The separator regex is not the same for `manage_variable_values_ini` and `set_variable_values_ini`. Where in the former case, there is no whitespace matching, while the latter does add this. I guess this should be standardized also?

This is what results in the weird output you can see in `set_variable_values_ini.cf.finish`.

tests/acceptance/lib/files/set_variable_values_ini.cf.finish
```diff
@@ -10,5 +10,8 @@ variablename6= desired value6
 variablename7 = desired value7
 variablename8 = desired value8
 variablename4=desired value4
+[CFE-3244]
+  keyone= 1
+keytwo = valuetwo
```
oleorhagen: Inconsistent and weird here.

lib/files.cf
```diff
@@ -485,7 +485,7 @@ bundle edit_line set_variable_values_ini(tab, sectionName)
       ifvarclass => "edit_$(cindex[$(index)])";
 
       # match a line starting like the key something
-      "$(index)\s*=.*"
+      "\s*$(index)\s*=.*"
       edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
```
oleorhagen: adds `\s*=\s*` which is not present in `set_variable_values_ini`

craigcomstock:
@oleorhagen usually we have a `Ticket: CFE-3244` in a commit as well. I tend to mostly do single commit PRs so it makes most sense in that case.

lib/files.cf
```diff
@@ -428,7 +428,7 @@ bundle edit_line manage_variable_values_ini(tab, sectionName)
       ifvarclass => "edit_$(cindex[$(index)])";
 
       # match a line starting like the key something
-      "$(index)\s*=.*"
```
olehermanse: Please add 

```
Ticket: CFE-3244
```

to the commit message, and it will be included in the changelog.

oleorhagen: @olehermanse So should I squash it as well? Into one commit?

olehermanse: @oleorhagen No, why do you say that?

oleorhagen: Because of what Craig said above...

olehermanse: I don't know exactly what he means there. But you should add a ticket number to the commit that adds a changelog entry :)

olehermanse: @oleorhagen and make the Changelog entry past tense(!) 😠 

```
Whitespace is now allowed at the beginning of ini key-values
```

oleorhagen: haha, allowed is past tense, no? :p 
Anywho. Do you have any answers to the questions I posted for Nick, @olehermanse ?

oleorhagen: Like, I don't think this is ready to merge as is.

tests/acceptance/lib/files/set_variable_values_ini.cf
```diff
@@ -34,6 +34,9 @@ bundle agent test
       "description" -> { "CFE-3221" }
         string => "Test expected behavior of set_variable_values_ini";
 
+      "description" -> { "CFE-3244" }
```
nickanderson: Instead of re-setting the description here, please move the promisee/stakeholder ( "CFE-3244" ) to the previous one.

```
"description" -> { "CFE-3221", "CFE-3244" }
```

nickanderson: Also, should have similar change for manage_variable_values_ini.cf

nickanderson:
After that, I think it's good. Perhaps squash it all into one commit?

lib/files.cf
nickanderson: Is this comment outdated?

The regular expressions look the same to me:

```
manage_variable_values_ini
     "\s*$(index)\s*=.*"
          edit_field => col("=","2","$($(tab)[$(sectionName)][$(index)])","set"),

set_variable_values_ini
     "\s*$(index)\s*=.*"
      edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
```

oleorhagen: > ```
> manage_variable_values_ini
>      "\s*$(index)\s*=.*"
>           edit_field => col("=","2","$($(tab)[$(sectionName)][$(index)])","set"),
> 
> set_variable_values_ini
>      "\s*$(index)\s*=.*"
>       edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
> ```

The regexes are the same here, but:
>           edit_field => col("=","2","$($(tab)[$(sectionName)][$(index)])","set"),
and
>       edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
differ for the `=`, and `\s*=\s*` separators. 

I'm thinking this is not on purpose. So I guess I should unify it?

nickanderson: I see. I am not against unifying them, but I don't have an opinion on which way they ought to be unified. I guess I don't care so long as the test captures any subtle change in behavior.

oleorhagen:
@olehermanse, fixed!

lib/files.cf
oleorhagen: Leaving it as a follow up.

olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4487/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4487/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4487/

olehermanse:
Retry of deployment tests:

[Build Status](https://ci.cfengine.com/job/testing-run-deployment-tests/4661/)

oleorhagen:
Hell yeah!
Merge @olehermanse ?

olehermanse:
@oleorhagen @nickanderson please discuss whether this should be backported.

nickanderson:
No, I don't think it should be backported. It's not a bug fix and I think it would be an unexpected behavior change during a patch release.

PR #1652, opened 2020-01-14
Do not bind to "::" on RHEL/CentOS 5 by default

vpodzime:
Trying to bind to "::" (IPV6_ANY) fails on RHEL/CentOS 5.


vpodzime:
@mikeweilgart what do you think about this? Your re-fix #1614 broke tests on RHEL 5. This fixes the problem.

vpodzime:
jenkins build (deployment tests only): [Build Status](https://ci.cfengine.com/job/pr-pipeline/4381/)

vpodzime:
> @mikeweilgart what do you think about this? Your re-fix #1614 broke tests on RHEL 5. This fixes the problem.

Merging the PR to get our tests to green. We can do more stuff in a follow-up PR if there's something else to do.

PR #1664, opened 2020-01-24
Simplify the periodic inventory refresh

vpodzime:
The changes in cfengine/nova#1584 introduce a better way to
refresh inventory than just doing it periodically from agent
run. However, we still need to refresh the information about the
variables shown on the Inventory page in Mission Portal.

A new PHP CLI task with a more appropriate name
'inventory_variables_refresh' was introduced in
cfengine/mission-portal@d2550582, but we want to keep using the
old task for backwards compatibility.

Ticket: ENT-4864
Changelog: Inventory refresh is no longer part of agent run on the hub

### DIFF
--- a/cfe_internal/enterprise/CFE_hub_specific.cf
+++ b/cfe_internal/enterprise/CFE_hub_specific.cf
@@ -552,20 +552,8 @@ bundle agent cfe_internal_enterprise_maintenance
         ifvarclass => isgreaterthan( $(enterprise_maintenance_bundle_count), 0 );
 }
 
-bundle agent cfe_internal_refresh_inventory_args
-{
-  vars:
-    cfengine_enterprise_federation:am_superhub::
-      "args" -> {"ENT-4527"}
-        string => "inventory_refresh_by_hostkey $(sys.key_digest)",
-        comment => "Only refresh inventory for the superhub, other hosts get refreshed on data import";
-
-    !cfengine_enterprise_federation:am_superhub::
-      "args" string => "inventory_refresh";
-}
-
 bundle agent cfe_internal_refresh_inventory_view
-# @brief Refresh materialized view every 5 minutes
+# @brief Refresh list of inventory variables every 5 minutes
 {
   meta:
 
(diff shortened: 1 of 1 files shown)


nickanderson:
@vpodzime What versions are you targeting here?

We need to remember to maintain compatibility with 3.10.x throughout 3.15.x, including the time when older versions are running newer MPF with old binaries. I don't see any version handling in this change.

vpodzime:
> compatibility with 3.10.x

3.10.x is no longer supported, right?

vpodzime:
> I don't see any version handling in this change.

I've added a variable for the CLI task name with two possible values based on classes. Do you think we can sacrifice performance on 3.12.x superhubs using *master* MPF in sake of keeping this much simpler?

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -552,44 +552,33 @@ bundle agent cfe_internal_enterprise_maintenance
...
     (policy_server|am_policy_hub).enterprise_edition::
 
       "tags" slist => { "enterprise_maintenance" };
 
-  methods:
-@if minimum_version(3.12.0)
-      # we need to know if we are running on a superhub
-      "superhub_config" usebundle => "cfengine_enterprise_federation:config";
-@endif
+  vars:
```
nickanderson: Would be nice to have cf_version_less|greater_than function(s) for this.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -305,7 +305,7 @@ bundle agent inventory_refresh_cmd
 {
   vars:
     "cmd" data => parsejson('{"inventory_refresh_cmd":
-                                "$(sys.workdir)/httpd/php/bin/php $(cfe_internal_hub_vars.docroot)/index.php cli_tasks inventory_refresh"}');
+                                "$(sys.workdir)/httpd/php/bin/php $(cfe_internal_hub_vars.docroot)/index.php cli_tasks inventory_variables_refresh"}');
```
nickanderson: Doesn't this need similar treatment with respect to backward compatibility?

PR #1674, opened 2020-02-04
CFE-3238/3.15.x: Standard services now considers systemd services in ActiveState=activ…

nickanderson:


nickanderson:
failure in master is because of missing nightly package.

PR #1710, opened 2020-03-18
ENT-5420/3.15.x Added use of services promise for FR postgresql reconfig in case of systemd

craigcomstock:
Using pg_ctl directly on systemd systems causes instability in
services due to systemd reacting to postgresql being restarted.

Ticket: ENT-5420
Changelog: title

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -452,12 +452,18 @@ bundle agent postgres_config
         classes => default:results( "bundle", "postgresql_conf" );
 
   commands:
-    am_superhub.postgresql_conf_repaired::
+    am_superhub.postgresql_conf_repaired.!systemd::
       # smart mode tries to wait for operations to finish and clients to
       # disconnect, fast mode terminates open connections gracefully
       "$(sys.bindir)/pg_ctl --pgdata $(sys.statedir)/pg/data --log /var/log/postgresql.log --wait --mode smart restart ||
        $(sys.bindir)/pg_ctl --pgdata $(sys.statedir)/pg/data --log /var/log/postgresql.log --wait --mode fast  restart"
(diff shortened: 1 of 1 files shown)


craigcomstock:
> This may or may not be OK. The change itself looks ok, but going from very explicit options (`-w -m smart`) to opaque abstraction via services might have insidious differences.
> 
> Should review the details for -m smart and -w between init scripts and systemd units to be sure things are all well aligned.

@nickanderson and I talked on chat and I think there is very little risk of FR enablement causing a `-m fast` restart of cf-postgres since the hub in question will almost certainly be freshly installed and have little or no bootstrapped hosts other than itself and very little traffic to postgres which would be lost by `-m fast`.

I will test on ubuntu16 and debian7 (no systemd I think).

craigcomstock:
debian7/init scripts will need to use old method. Working on the modification.

craigcomstock:
If https://github.com/cfengine/masterfiles/pull/1711 works, I think that would be a better solution.

craigcomstock:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4872/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4872/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4872/

craigcomstock:
jenkins debian9 failing is flaky in my opinion. Manually tested jenkins builds on centos7 (selinux), debian7 (no systemd, no selinux), ubuntu16 (systemd, no selinux) and debian9 (systemd, no selinux).
Existing pain point is that centos7 (and really any selinux system) requires two agent runs in order to enable superhub.
Will cherry-pick back to master.

PR #1719, opened 2020-03-23
 CFE-3291/3.12.x: Made python symlink fall back to platform-python

nickanderson:


nickanderson:
rebased and pushed, let's see what travis has to say this time.

PR #1724, opened 2020-03-30
ENT-5432: Fixed runalerts processes promise on non-systemd systems

craigcomstock:
active_hub class is needed to properly manage runalerts.php
processes promise on non-systemd systems so had to swap
order of methods promises.

Ticket: ENT-5432
Changelog: title
(cherry picked from commit 4a490adfc881ad0f88cc1fa0c6840c98ab0f861b)


craigcomstock:
@nickanderson will squash after travis is green and you review. :+1:

PR #1738, opened 2020-04-13
CFE-3237 Copy modules from masterfiles to workdir on hub during update

Lex-2008:
Since now they are managed by masterfiles.

#

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -236,15 +236,17 @@ bundle agent cfe_internal_update_policy_cpv
       action => u_immediate,
       classes => u_if_repaired("validated_updates_ready");
 
-    !am_policy_hub::
+    any::
 
       "$(sys.workdir)$(const.dirsep)modules"
-      comment => "Always update modules files on client side",
+      comment => "Update modules from masterfiles",
       handle => "cfe_internal_update_policy_files_update_modules",
-      copy_from => u_rcp("$(modules_dir_source)", @(update_def.policy_servers)),
+      copy_from => u_cp("$(master_location)$(const.dirsep)modules"),
       depth_search => u_recurse("inf"),
       perms => u_m("755"),
       action => u_immediate;
+
+    !am_policy_hub::
 
       "$(sys.workdir)$(const.dirsep)plugins"
       comment => "Always update plugins files on client side",
--- a/modules/packages/Makefile.am
+++ b/modules/packages/Makefile.am
@@ -1,3 +1,3 @@
-package_modulesdir = $(prefix)/modules/packages
+package_modulesdir = $(prefix)/masterfiles/modules/packages
 
 dist_package_modules_SCRIPTS = apt_get yum pkg nimclient pkgsrc freebsd_ports zypper slackpkg msiexec.bat WiRunSQL.vbs


cfe_internal/update/update_policy.cf
```diff
@@ -236,6 +236,16 @@ bundle agent cfe_internal_update_policy_cpv
       action => u_immediate,
       classes => u_if_repaired("validated_updates_ready");
 
+    am_policy_hub::
+
+      "$(sys.workdir)$(const.dirsep)modules"
```
nickanderson, line 241: I don't think that only the hub should do this. If we are moving to having MPF put modules in place, I think that should happen on each host.

cfe_internal/update/update_policy.cf
```diff
@@ -236,16 +236,18 @@ bundle agent cfe_internal_update_policy_cpv
...
 
-    !am_policy_hub::
+    any::
 
       "$(sys.workdir)$(const.dirsep)modules"
-      comment => "Always update modules files on client side",
+      comment => "Update modules from masterfiles",
       handle => "cfe_internal_update_policy_files_update_modules",
-      copy_from => u_rcp("$(modules_dir_source)", @(update_def.policy_servers)),
+      copy_from => u_cp("$(master_location)$(const.dirsep)modules"),
```
vpodzime, line 244: It should still copy the package modules from the policy server(s), right?

cfe_internal/update/update_policy.cf
```diff
@@ -236,16 +236,18 @@ bundle agent cfe_internal_update_policy_cpv
...
       depth_search => u_recurse("inf"),
       perms => u_m("755"),
```
nickanderson, line 246: Will this be the right perms on windows?

cfe_internal/update/update_policy.cf
nickanderson: Probably it should first copy modules from the policy server(s), then it should place the modules that are vendored by the mpf.

vpodzime: ? I thought the package modules will be part of MPF and thus under `/var/cfengine/masterfiles/` and thus updated like everything else.

Lex-2008: Right. Somewhere else policy copies from `/var/cfengine/masterfiles/` on hub to `/var/cfengine/inputs` on agent. Here we should be copying from `/var/cfengine/inputs/modules` (not masterfiles!) to `/var/cfengine/modules` - and do it on all machines.

@nickanderson : is it okay to do it in two steps or is it better to copy to `/var/cfengine/modules` from `/var/cfengine/masterfiles/modules` at the same time as copying to `/var/cfengine/inputs` from `/var/cfengine/masterfiles` on hub?

nickanderson: Two promises are OK I think, they are different targets. or did you mean something else?

Lex-2008: I mean that if they don't get executed in proper order, there might be an agent run when a policy is updated, but the modules are not there.

cfe_internal/update/update_policy.cf
Lex-2008: from https://docs.cfengine.com/docs/master/reference-promise-types-files.html#mode:

> Notes: This is ignored on Windows, as the permission model uses ACLs.

Also note that Windows doesn't have "executable" permission bit, but checks file extensions instead. So if you can read an `*.exe` file - you can as well execute it.

craigcomstock:
This commit breaks agents running updates. PR in the works to fix.

craigcomstock:
On second thought, my quick-fix PR wasn't fully baked, so will revert this.

Failed cases in CI from this PR are I believe:

- [ ] bootstrap-tests
  - [ ] new_tests/bootstrap_tests/.correct_permissions_after_bootstrap.test
  - [ ] new_tests/bootstrap_tests/.restart.test
- [ ] sequential-tests
  - [ ] new_tests/sequential_tests/01-basic.01-policy_distribution.test
  - [ ] new_tests/sequential_tests/02-upgrade-minor.01-upgrade.test

PR #1741, opened 2020-04-15
Changed path to scripts from docroot to public_docroot

aleksandrychev:
ChangeLog: None


aleksandrychev:
@nickanderson  please review and merge if OK

PR #1751, opened 2020-04-29
ENT-5562 Changed m_inventory dumping behavior to exclude when values is null (3.12.x)

craigcomstock:
Previously an empty values column in m_inventory such as when a
host has not reported would cause corrupt SQL to be dumped.
In this case we should simply not include that host in the
output in a similar way that __inventory is built from v_inventory
in latest schema.

Changelog: Title
Ticket: ENT-5562

### DIFF
--- a/templates/federated_reporting/dump.sh
+++ b/templates/federated_reporting/dump.sh
@@ -58,7 +58,8 @@ log "Dumping tables: $CFE_FR_TABLES"
 
   # in case of 3.12 must copy m_inventory as if it was __inventory
   if [[ "$CFE_VERSION" =~ "3.12." ]]; then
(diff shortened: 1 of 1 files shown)


craigcomstock:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5095/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5095/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5095/

craigcomstock:
> Looks good to me. Just please also add a comment about why we cannot use `pg_dump` for `m_inventory`.

Ok. Added a comment. Please approve and we can merge.

PR #1777, opened 2020-06-04
CFE-3362 Changed group for state dir files promise to match defaults

craigcomstock:
Especially this concerns lmdb files.
system_owned perms body was used previously but for solaris systems the use
of "sys" group seems incorrect so changing to "root" for solaris is the
net result of this change.

Ticket: CFE-3362
Changelog: Title


olehermanse:
This looks identical to the other PR and can be closed?

PR #1780, opened 2020-06-08
Fixed travis release jobs

craigcomstock:
Travis jobs which test changes to MPF against different versions of core are broken for 3.15.x currently because the .travis.yml script installs 3.15.2 (no cf-secret) but runs testall from core 3.15.x which wants cf-secret.

I will try to change the script to use core with the cf-agent version as a tag instead so things should be "in sync" better.


craigcomstock:
will cherry-pick the master PR I have up instead.

PR #1781, opened 2020-06-08
Changed travis release jobs to use release tag of core matching packages installed

craigcomstock:
Otherwise they can be out of sync, such as 3.15.x tests/acceptance/testall
wanting cf-secret but 3.15.2 not having it.

Ticket: none
Changelog: none


craigcomstock:
Despite a comment here from travis, this PR is green there and should be merged I think.

https://travis-ci.org/github/cfengine/masterfiles/builds/696498745

I can cherry-pick afterward.

PR #1782, opened 2020-06-09
test travis integration with empty commit on master

craigcomstock:


craigcomstock:
was invalid .travis.yml file (included && I think was the problem). Only found by looking at https://travis-ci.org/github/cfengine/masterfiles/requests page to see "Could not parse" messages there.

PR #1804, opened 2020-07-15
CFE-3288 Adjusted package_module and paths for termux platform

craigcomstock:
Packages promises only possibly supported back to 3.15.1 due to
need for interpreter attribute on package_module body and no
earlier version available for termux.

Paths added to support the interpreter path for packages.
Also several convenience paths added both for termux and generic unix:
- paths.tmp_path, paths.bin_path, paths.var_path, paths.etc_path

Ticket: CFE-3288
Changelog: Title

### DIFF
--- a/lib/packages.cf
+++ b/lib/packages.cf
@@ -32,6 +32,9 @@ bundle common package_module_knowledge
     windows::
       "platform_default" string => "msiexec";
 @endif
+
+    termux::
+      "platform_default" string => "apt_get";
 }
 
 body package_module apt_get
@@ -39,6 +42,10 @@ body package_module apt_get
     query_installed_ifelapsed => "$(def.package_module_query_installed_ifelapsed)";
     query_updates_ifelapsed => "$(def.package_module_query_updates_ifelapsed)";
     #default_options =>  {};
+@if minimum_version(3.12.2)
+    termux::
+      interpreter => "$(paths.bin_path)/python";
+@endif
 }
 
 body package_module zypper
--- a/lib/paths.cf
+++ b/lib/paths.cf
@@ -1,8 +1,8 @@
 # Paths bundle (used by other bodies)
 
 bundle common paths
-# @brief Defines an array `path` with common paths to standard binaries,
-# and classes for defined and existing paths.
+# @brief Defines an array `path` with common paths to standard binaries and
+# directories as well as classes for defined and existing paths.
 #
 # If the current platform knows that binary XYZ should be present,
 # `_stdlib_has_path_XYZ` is defined. Furthermore, if XYZ is actually present
@@ -25,6 +25,22 @@ bundle common paths
 #        comment => "Alternate array reference style";
 # }
 # ```
+#
+# Additionally several path entries are present to aid in policy sharing between
+# unix systems and Android Termux environments.
+#
+# **Example:**
+#
+# ```cf3
+# bundle agent track_sshd_config
+# {
+#   files:
+#     "$(paths.etc_path)/sshd/sshd_config"
+#       changes => detect_all_change;
+# }
+# ```
+#
+# In case of termux, `paths.etc_path` will be `/data/data/com.termux/files/usr/etc`.
 {
   vars:
 
@@ -569,6 +585,32 @@ bundle common paths
     aix::
 
       "path[shadow]"       string => "/etc/security/passwd";
+
+    termux::
+      "path[tar]"           string => "/usr/bin/tar";
+      "path[true]"          string => "/usr/bin/true";
+      "path[false]"         string => "/usr/bin/false";
+      "path[cat]"           string => "/usr/bin/cat";
+      "path[sysctl]"        string => "/usr/bin/sysctl";
+      "path[env]"           string => "/usr/bin/env";
+
+      # now, mangle the values by prepending the TERMUX_PREFIX
+      "files_path" string => "/data/data/com.termux/files";
+      "etc_path" string => "$(files_path)/usr/etc";
+      "tmp_path" string => "$(files_path)/usr/tmp";
+      "bin_path" string => "$(files_path)/usr/bin";
+      "var_path" string => "$(files_path)/usr/var";
+      "tmp_paths" slist => getindices("path");
+      "tmp_path[$(tmp_paths)]" string => "$(files_path)$(path[$(tmp_paths)])";
+      "path[$(tmp_paths)]" string => "$(tmp_path[$(tmp_paths)])";
+
+    !(termux|windows)::
+      # reasonable defaults for unix systems to allow for writing
+      # more portable paths between termux and other systems
+      "etc_path" string => "/etc";
+      "tmp_path" string => "/tmp";
+      "bin_path" string => "/bin";
+      "var_path" string => "/var";
 
     any::
 
--- a/promises.cf.in
+++ b/promises.cf.in
@@ -107,6 +107,8 @@ body common control
           package_inventory => { $(package_module_knowledge.platform_default) };
           package_module => $(package_module_knowledge.platform_default);
 @endif
+    termux::
+          package_module => $(package_module_knowledge.platform_default);
 
     any::
         ignore_missing_bundles => "$(def.control_common_ignore_missing_bundles)";


craigcomstock:
ah, right, package_module body interpreter attribute is not supported in 3.10.x, will adjust accordingly with macros right?

lib/paths.cf
```diff
@@ -570,6 +570,29 @@ bundle common paths
 
       "path[shadow]"       string => "/etc/security/passwd";
 
+@if minimum_version(3.16)
```
craigcomstock, line 573: @nickanderson this section and addition of root_path and prefix_path as usable elsewhere in policy is starting to feel like "too much" focused on termux issues. Should these differences have to be handled by folks each time they write policy for termux?

lib/packages.cf
```diff
@@ -32,13 +32,21 @@ bundle common package_module_knowledge
     windows::
       "platform_default" string => "msiexec";
 @endif
+@if minimum_version(3.16)
+    termux::
```
nickanderson, line 36: I don't think it is necessary to guard this based on minimum version.

termux class may not be defined until 3.16.1, but that just means it will be out of context, so won't apply.

lib/paths.cf
```diff
@@ -570,6 +570,29 @@ bundle common paths
...
+
+      # now, mangle the values by prepending the TERMUX_PREFIX
+      "root_path" string => "/data/data/com.termux/files";
+      # a prefix to use elsewhere
+      "prefix_path" string => "$(root_path)$(const.dirsep)usr";
+      "tmp_paths" slist => getindices("path");
+      "tmp_path[$(tmp_paths)]" string => "$(root_path)$(path[$(tmp_paths)])";
+      "path[$(tmp_paths)]" string => "$(tmp_path[$(tmp_paths)])";
+
+    !termux::
```
nickanderson, line 589: perhaps `!(termux|windows)`

lib/packages.cf
nickanderson: Actually, i do think you need this, but not for 3.16, more for 3.12.2? whenever interpreter was introduced.

lib/paths.cf
nickanderson: its nice if it just works for people, not having to be super explicit if doing so doesn't add a bunch of overhead for other platforms.

lib/packages.cf
```diff
@@ -32,13 +32,21 @@ bundle common package_module_knowledge
...
+      "platform_default" string => "apt_get";
+@endif
 }
 
 body package_module apt_get
 {
     query_installed_ifelapsed => "$(def.package_module_query_installed_ifelapsed)";
     query_updates_ifelapsed => "$(def.package_module_query_updates_ifelapsed)";
     #default_options =>  {};
+@if minimum_version(3.16.1)
```
craigcomstock, line 46: @nickanderson so maybe ONLY have a guard here for 3.12.2 where interpreter was introduced and remove all others since termux hard-class will almost certainly never be defined?

lib/packages.cf
```diff
@@ -32,13 +32,22 @@ bundle common package_module_knowledge
...
+      "platform_default" string => "apt_get";
+@endif
 }
 
 body package_module apt_get
 {
     query_installed_ifelapsed => "$(def.package_module_query_installed_ifelapsed)";
     query_updates_ifelapsed => "$(def.package_module_query_updates_ifelapsed)";
     #default_options =>  {};
+@if minimum_version(3.12.2)
```
nickanderson, line 47: This is required because prior to 3.12.2, interpreter was not understood by the language.

lib/packages.cf
```diff
@@ -32,13 +32,22 @@ bundle common package_module_knowledge
     windows::
       "platform_default" string => "msiexec";
 @endif
+
+@if minimum_version(3.12.2)
```
nickanderson, line 36: Probably don't need this guard.

nickanderson: This is in a common bundle, no langauge issues related to this one. platform_default is just a variable, fine to set on any version.

lib/packages.cf
nickanderson: Note, this will break on 3.13.x

lib/paths.cf
```diff
@@ -570,6 +570,27 @@ bundle common paths
...
+      "prefix_path" string => "$(root_path)$(const.dirsep)usr";
+      "tmp_paths" slist => getindices("path");
+      "tmp_path[$(tmp_paths)]" string => "$(root_path)$(path[$(tmp_paths)])";
+      "path[$(tmp_paths)]" string => "$(tmp_path[$(tmp_paths)])";
+
+    !(termux|windows)::
+      # reasonable defaults for unix systems to allow for writing
+      # more portable paths between termux and other systems
+      "root_path" string => "/";
+      "prefix_path" string => "/usr";
```
nickanderson, line 592: Where is prefix_path used?

lib/paths.cf
```diff
@@ -570,6 +570,27 @@ bundle common paths
...
+      # a prefix to use elsewhere
+      "prefix_path" string => "$(root_path)$(const.dirsep)usr";
+      "tmp_paths" slist => getindices("path");
+      "tmp_path[$(tmp_paths)]" string => "$(root_path)$(path[$(tmp_paths)])";
+      "path[$(tmp_paths)]" string => "$(tmp_path[$(tmp_paths)])";
+
+    !(termux|windows)::
+      # reasonable defaults for unix systems to allow for writing
+      # more portable paths between termux and other systems
+      "root_path" string => "/";
```
nickanderson, line 591: root_path probably isn't correct for windows

lib/paths.cf
craigcomstock: intended to be shared, so needs docs I would say.

craigcomstock: I will find a better name and add docs+example like

```
$(thing)/etc/sshd/sshd_config
```

So that `$(thing)` will be "right" on both unix systems and termux.

lib/packages.cf
craigcomstock: Can I do complex expressions then? Will check. :+1:

craigcomstock: How about 3.15.1? That's the earliest termux version and handles the situation nicely.

https://github.com/termux/termux-packages/commits/master/packages/cfengine

lib/paths.cf
```diff
@@ -25,6 +25,22 @@ bundle common paths
 #        comment => "Alternate array reference style";
 # }
 # ```
+#
+# Additionally a `root_path` entry is present in the `path` array to aid in policy
```
nickanderson, line 29: in the `path` array, or in the `paths` bundle?

lib/paths.cf
```diff
@@ -570,6 +586,25 @@ bundle common paths
...
+      "files_path" string => "/data/data/com.termux/files";
+      "root_path" string => "$(root_path)$(const.dirsep)usr";
+      "tmp_paths" slist => getindices("path");
+      "tmp_path[$(tmp_paths)]" string => "$(files_path)$(path[$(tmp_paths)])";
+      "path[$(tmp_paths)]" string => "$(tmp_path[$(tmp_paths)])";
+
+    !(termux|windows)::
+      # reasonable defaults for unix systems to allow for writing
+      # more portable paths between termux and other systems
+      "root_path" string => "/";
```
nickanderson, line 606: Does it make sense to have this be `/` vs having it be an empty string?

e.g.
`"$(paths.root_path)/etc/fstab"` == `//etc/fstab` vs `/etc/fstab`

lib/paths.cf
craigcomstock: oh, right, bundle.

lib/paths.cf
craigcomstock: correct. I would have caught that. Thanks! :)

lib/paths.cf
```diff
@@ -570,6 +586,31 @@ bundle common paths
...
+    termux::
+      # some adjustments for the basics
+      "path[tar]"           string => "/usr/bin/tar";
+      "path[true]"          string => "/usr/bin/true";
+      "path[false]"         string => "/usr/bin/false";
+      "path[sysctl]"        string => "/usr/bin/sysctl";
+
+      # now, mangle the values by prepending the TERMUX_PREFIX
+      "files_path" string => "/data/data/com.termux/files";
+      "etc_path" string => "$(files_path)/usr/etc";
```
craigcomstock, line 598: @nickanderson this is what I came up with after some experimentation. Too much? Too arbitrary that I chose bin, etc, var, tmp? I figured at least etc is super useful for "configuration" :) So I could limit it to that and bin maybe (used `$(paths.bin_path)` for python in packages.cf above.

lib/paths.cf
```diff
@@ -570,6 +586,31 @@ bundle common paths
...
+      "bin_path" string => "$(files_path)/usr/bin";
+      "var_path" string => "$(files_path)/usr/var";
+      "tmp_paths" slist => getindices("path");
+      "tmp_path[$(tmp_paths)]" string => "$(files_path)$(path[$(tmp_paths)])";
+      "path[$(tmp_paths)]" string => "$(tmp_path[$(tmp_paths)])";
+
+    !(termux|windows)::
+      # reasonable defaults for unix systems to allow for writing
+      # more portable paths between termux and other systems
+      "etc_path" string => "/etc";
```
craigcomstock, line 609: @nickanderson granted, this section could start varying a WHOLE lot so probably best to limit it to only VERY useful things to avoid maintenance worry.

lib/packages.cf
```diff
@@ -32,13 +32,20 @@ bundle common package_module_knowledge
...
+    termux::
+      "platform_default" string => "apt_get";
 }
 
 body package_module apt_get
 {
     query_installed_ifelapsed => "$(def.package_module_query_installed_ifelapsed)";
     query_updates_ifelapsed => "$(def.package_module_query_updates_ifelapsed)";
     #default_options =>  {};
+@if minimum_version(3.15.1)
```
nickanderson, line 45: Perhaps hash comment about why this is targeting 3.15.1 instead of 3.12.something (because it woudlnt work for 3.13.

craigcomstock:
WIP, need to double-check/test paths.(all the commands) to make sure things are "all good".

lib/paths.cf
```diff
@@ -570,6 +586,32 @@ bundle common paths
 
       "path[shadow]"       string => "/etc/security/passwd";
 
+    termux::
+      # some adjustments for the basics
+      "path[tar]"           string => "/usr/bin/tar";
+      "path[true]"          string => "/usr/bin/true";
+      "path[false]"         string => "/usr/bin/false";
+      "path[cat]"           string => "/usr/bin/cat";
+      "path[sysctl]"        string => "/usr/bin/sysctl";
```
nickanderson, line 595: Curiosu if these really need to be over-ridden from debian or linux.

lib/paths.cf
nickanderson: I think it's OK.

lib/paths.cf
craigcomstock: linux has:

```
      "path[tar]"           string => "/bin/tar";
      "path[true]"          string => "/bin/true";
      "path[false]"         string => "/bin/false";
```

So that's a yes for those and the others aren't handled otherwise and would default to just /bin so must be over-ridden.

Termux environment doesn't define debian even though it uses apt. :)

PR #1803, opened 2020-07-14
Moved systemd service management to own bundle

nickanderson:
This change moves the systemd service management promises to their own bundle.
There should be no change in behavior as using standard_services when systemd is
present will still result in the use of the same promises. This change makes it
easier to maintain the policy as it is logically separate and allows explicit
management of systemd services.

PR #1805, opened 2020-07-15
CFE-3381/master: Moved systemd service management to own bundle

nickanderson:
This change moves the systemd service management promises to their own bundle.
There should be no change in behavior as using standard_services when systemd is
present will still result in the use of the same promises. This change makes it
easier to maintain the policy as it is logically separate and allows explicit
management of systemd services.

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -76,7 +76,20 @@ body service_method standard_services
         service_bundle => default:standard_services( $(this.promiser), $(this.service_policy) );
 }
 
-##
+body service_method systemd_services
+# @brief systemd service method
+#
+# **Example:**
+#
+# ```cf3
+# services:
+#     "ssh"
+#       service_policy => "enabled",
+#       service_method => systemd_services;
+# ```
+{
+        service_bundle => default:systemd_services( $(this.promiser), $(this.service_policy) );
+}
 
 bundle agent standard_services(service,state)
 # @brief Standard services bundle, used by CFEngine by default
@@ -94,7 +107,7 @@ bundle agent standard_services(service,state)
 # This bundle receives the service name and the desired service state,
 # then does the needful to reach the desired state.
 #
-# If you're running systemd, systemctl will be used.
+# If you're running systemd, systemctl will be used via `systemd_services`.
 #
 # Else, if chkconfig is present, it will be used.
 #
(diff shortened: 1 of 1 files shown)


nickanderson:
@craigcomstock, vratislav already approved this in a different PR https://github.com/cfengine/masterfiles/pull/1803

nickanderson:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5372/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5372/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5372/

lib/services.cf
```diff
@@ -118,10 +131,16 @@ bundle agent standard_services(service,state)
 # methods:
 #     "" usebundle => standard_services("sshd", "start"); # direct
 # ```
+#
+# Alternatively, since services promises are an abstraction around bundles, the service state can be promised via a methods type promise.
+#
+# ```cf3
+#
+# methods:
+#     "SSHD should be running" usebundle => standard_services("sshd", "start");
```
craigcomstock, line 140: Why would you want to do this? Other than that it is technically possible?

lib/services.cf
```diff
@@ -340,8 +281,134 @@ bundle agent standard_services(service,state)
     verbose_mode.fallback::
       "$(this.bundle): falling back to classic_services to $(state) $(service)";
 
-    systemd.service_notfound.(start|restart|reload).(inform_mode|verbose_mode)::
-        "$(this.bundle): Could not find service: $(service)";
+}
+
+bundle agent systemd_services(service,state)
```
craigcomstock, line 286: Why not move these to their own files? Might make it even easier to extend for different services frameworks, granted there might only be a few others to add? termux and openrc and what does osx use?

lib/services.cf
nickanderson: it /might/ be preferable because of difference in normal ordering. But generally, you woudlnt do this because it's mroe clear to use the services syntax.

lib/services.cf
nickanderson: Do you mean like `lib/services-systemd.cf`?

We don't really subdivide other parts of the stdlib, not sure if it makes sense to do for services. How does having in separate file make it easier to extend?

Also, the network protocol is a bit chatty, so try not to be super liberal about adding files to MPF.

lib/services.cf
craigcomstock: so maybe add the why to this comment so people don't just use the pattern.

lib/services.cf
craigcomstock: good reasons. I just figure smaller files are usually easier to read in one go and easier to grok. Also easier to maintain since the diffs for systemd won't affect the main framework file. But :+1: on this due to your network protocol reason.

PR #1807, opened 2020-07-24
ENT-6035 Added duplicate host cleanup during FR import

craigcomstock:
merge together:
https://github.com/cfengine/nova/pull/1693
https://github.com/cfengine/masterfiles/pull/1807
https://github.com/cfengine/system-testing/pull/353

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -316,6 +316,8 @@ bundle agent federation_manage_files
     "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
     "feeder" data => parsejson('{"feeder": "$(sys.key_digest)"}');
     "cf_version" data => parsejson('{"cf_version":"$(sys.cf_version)"}');
+    "handle_duplicates_value" string => ifelse("cfengine_mp_fr_handle_duplicate_hostkeys", "yes", "no");
+    "handle_duplicates" data => parsejson('{"handle_duplicates":"$(handle_duplicates_value)"}');
 
   files:
     enterprise_edition.(policy_server|am_policy_hub)::
@@ -364,7 +366,11 @@ bundle agent federation_manage_files
         create => "true",
         template_method => "mustache",
         edit_template => "$(this.promise_dirname)/../../../templates/federated_reporting/config.sh.mustache",
-        template_data => mergedata(@(login), @(feeder_username), @(feeder), @(cf_version),
+        template_data => mergedata(@(login),
+                                   @(feeder_username),
+                                   @(feeder),
+                                   @(cf_version),
+                                   @(handle_duplicates),
                                    parsejson('{"inventory_refresh_cmd": ""}')),
         perms => default:mog( "640", "root", "$(transport_user)" );
 
--- a/templates/federated_reporting/import.sh
+++ b/templates/federated_reporting/import.sh
@@ -17,6 +17,7 @@ true "${CFE_FR_COMPRESSOR_EXT?undefined}"
 true "${CFE_FR_EXTRACTOR?undefined}"
 true "${CFE_FR_TABLES?undefined}"
 true "${CFE_FR_INVENTORY_REFRESH_CMD?undefined}"
+true "${CFE_FR_HANDLE_DUPLICATES?undefined}"
 
 if ! type "$CFE_FR_EXTRACTOR" >/dev/null; then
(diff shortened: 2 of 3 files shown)


craigcomstock:
@nickanderson do you feel we should enable this de-dupe call by default? Or have it be an optional thing to enable by editing templates/federated_reporting/config.sh.mustache or even augments maybe?

templates/federated_reporting/import.sh
```diff
@@ -100,6 +102,22 @@ else
...
+  hostkey_list=$(printf "'%s'," ${hostkeys[*]} | sed -e 's/,$//')
+  "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
+                      -c "SELECT handle_duplicate_hostkeys_in_import(ARRAY[$hostkey_list]);"  \
+    > duplicates.log || failed=1
+  if [ "$failed" = "0" ]; then
+    log "Handle Duplicate Hostkeys: DONE"
+  else
+    log "Handle Duplicate Hostkeys: FAILED"
+    log "last 10 lines of duplicates.log"
+    tail -n 10 duplicates.log
```
vpodzime, line 116: How big is this file? Shall we delete it? Or should it be compressed?

templates/federated_reporting/import.sh
```diff
@@ -100,6 +102,22 @@ else
...
+    exit 1
```
vpodzime, line 117: I don't think this should prevent the further steps in the import process below.

craigcomstock: I am inclined to think it should fail. If we don't fail it then the database could be in an odd state, especially if the users are expecting "no duplicates" and then suddenly there are a lot of duplicates.

If we `exit 1` then that will show up and could be handled like a "warning" to a sysadmin right?

templates/federated_reporting/import.sh
craigcomstock: It shouldn't be very big. There is no logging, say, for each host involved so will only scale linearly compared to the number of feeder hubs. We should make it the most recent one available the same as we do for the sql logs.

craigcomstock: Interestingly, I don't see where we remove the `schema_setup.log` either. I think any other logs besides the big compressed log from the actual data import should be left around for the last run and removed just before the next run.

I'll check a running system to confirm how this is working.

templates/federated_reporting/import.sh
vpodzime: Yes, but it will also mean they will have no data from the problematic feeder(s) at all. But then again, this only happens if they explicitly enable de-duplication, right? In that case it's OK to `exit` here, I think. :+1:

templates/federated_reporting/import.sh
craigcomstock: The file is very small and will be a constant size regardless of the number of records processed.

This along with the schema_setup.log and schema_attach.log will remain at `/opt/cfengine/federation/cfapache` and be re-written on next run. We might want to mention this location as a debugging aid as well as the tarred up import logs.

craigcomstock: Though I did need to add a `2>&1` so that stderr also goes to that logfile instead of the agent log.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -316,6 +316,8 @@ bundle agent federation_manage_files
     "feeder_username" data => parsejson('{"feeder_username":"$(cfengine_enterprise_federation:config.transport_user)"}');
     "feeder" data => parsejson('{"feeder": "$(sys.key_digest)"}');
     "cf_version" data => parsejson('{"cf_version":"$(sys.cf_version)"}');
+    "handle_duplicates_value" string => ifelse("cfengine_mp_fr_handle_duplicate_hostkeys","yes","no");
```
vpodzime, line 319: Missing spaces around `,`

templates/federated_reporting/import.sh
```diff
@@ -100,6 +102,22 @@ else
   done
 fi
 
+if [ "$CFE_FR_HANDLE_DUPLICATES" = "yes" ]; then
```
vpodzime, line 105: This new config variable should be added near the top of the script where it checks if all such variables are defined.

vpodzime (commented):
Looks good to me otherwise.

vpodzime (commented):
Please squash the review fixes.

PR #1822, opened 2020-08-24
Removed delay in refreshing software installed inventory

nickanderson:
This change sets the delay for updating the inventory of currently installed
packages to 0, effectively disabling any delay. This inventory affects
packagesmatching() as well as inventory collected by an Enterprise hub. Without
this change, changes to software made outside CFEngine may not be realized for
up to 60 minutes.

Ticket: ENT-6154
Changelog: Title

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1210,8 +1210,8 @@ For example:
 ### Configure periodic package inventory refresh interval
 
 CFEngine refreshes software inventory when it makes changes via packages
-promises. Additionally, by default, CFEngine periodically refreshes it's
-internal cache of packages installed (every 60 minutes) and package updates that
+promises. Additionally, by default, CFEngine refreshes it's
+internal cache of packages installed (during each agent run) and package updates that
 are available (once a day) according to the default package manager in order to
 pick up changes made outside packages promises.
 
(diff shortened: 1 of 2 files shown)


olehermanse (approved):
LGTM, but I would probably not backport, just in case it causes some unforeseen issues.

vpodzime (approved):
ACK. I also agree with **not** backporting this.

nickanderson:
/fine/ wont backport.

MPF.md
```diff
@@ -1210,8 +1210,8 @@ For example:
 ### Configure periodic package inventory refresh interval
 
 CFEngine refreshes software inventory when it makes changes via packages
-promises. Additionally, by default, CFEngine periodically refreshes it's
-internal cache of packages installed (every 60 minutes) and package updates that
+promises. Additionally, by default, CFEngine refreshes it's
+internal cache of packages installed (during each agent) run and package updates that
```
olehermanse, line 1214: @nickanderson run should be inside parens.

olehermanse (changes requested):

MPF.md
nickanderson: ACK, thanks!

PR #1823, opened 2020-08-27
ENT-6035 Added optional handle duplicates step in federated reporting import (3.15.x)

craigcomstock:
This feature removes duplicate hostkey entries during federated reporting
import based on lastreporttimestamp in __hosts table.

Enable this feature by defining a class in augments (def.json):

```
{
  "classes": {
    "cfengine_mp_fr_handle_duplicate_hostkeys": ["any::"]
  }
}
```

Ticket: ENT-6035
Changelog: title
(cherry picked from commit 998c953a584299a9c30634338f38688546e5d0ac)

Conflicts:
	cfe_internal/enterprise/federation/federation.cf

merge together:
https://github.com/cfengine/nova/pull/1701
https://github.com/cfengine/masterfiles/pull/1823

### DIFF
--- a/templates/federated_reporting/import.sh
+++ b/templates/federated_reporting/import.sh
@@ -17,6 +17,7 @@ true "${CFE_FR_COMPRESSOR_EXT?undefined}"
 true "${CFE_FR_EXTRACTOR?undefined}"
 true "${CFE_FR_TABLES?undefined}"
 true "${CFE_FR_INVENTORY_REFRESH_CMD?undefined}"
+true "${CFE_FR_HANDLE_DUPLICATES?undefined}"
 
 if ! type "$CFE_FR_EXTRACTOR" >/dev/null; then
   log "Extractor $CFE_FR_EXTRACTOR not available!"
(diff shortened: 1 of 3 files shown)


cf-bottom:

Lex-2008 (approved):

templates/federated_reporting/import.sh
```diff
@@ -90,6 +93,22 @@ else
...
+  log "Handle Duplicate Hostkeys"
+  hostkey_list=$(printf "'%s'," ${hostkeys[*]} | sed -e 's/,$//')
+  "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
+                      -c "SELECT handle_duplicate_hostkeys_in_import(ARRAY[$hostkey_list]);"  \
+    > duplicates.log 2>&1 || failed=1
+  if [ "$failed" = "0" ]; then
+    log "Handle Duplicate Hostkeys: DONE"
+  else
+    log "Handle Duplicate Hostkeys: FAILED"
+    log "last 10 lines of duplicates.log"
```
nickanderson, line 106: Where is this `dupliactes.log`?

I think we need some documentation about this, in MPF.md and perhaps in a MD living inside the FR directory, that we can just pull into the docs site. It seems like this log file could grow unbounded, so we need to highlight that, or add it to our log files that are managed.

vpodzime: Let's not do it in this cherry-pick, though.

vpodzime (approved):
ACK

templates/federated_reporting/import.sh
craigcomstock: Added a checkbox todo on ENT-6035.

craigcomstock: > Where is this `dupliactes.log`?
> 
> I think we need some documentation about this, in MPF.md and perhaps in a MD living inside the FR directory, that we can just pull into the docs site. It seems like this log file could grow unbounded, so we need to highlight that, or add it to our log files that are managed.

This `duplicates.log` file won't grow very large. Unless there are errors it will be a constant size regardless of the number of hosts or duplicates found. It is also written to with `<` so will always be written over on each import.

PR #1837, opened 2020-09-18
Update MPF.md

troymuller:
This parameter is incorrect based on the spelling in the controls/cf_serverd.cf file

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1021,7 +1021,7 @@ This can be configured via [augments][Augments]:
 ```
 {
   "vars":{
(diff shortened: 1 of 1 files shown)


troymuller:
Maximum number of concurrent connections.
      # Suggested value >= total number of hosts
      maxconnections        => "$(def.control_server_maxconnections)";

nickanderson (approved):
Indeed, good catch!

Thanks for the contribution Troy, it's much appreciated.

nickanderson:
Picked back to 3.15.x and 3.12.x branches.

PR #1838, opened 2020-09-23
ENT-6209 Added override augment for controls/cf_execd.cf

craigcomstock:
Ticket: ENT-6209
Changelog: title

### DIFF
--- a/controls/cf_execd.cf
+++ b/controls/cf_execd.cf
@@ -4,6 +4,30 @@
 # These body settings determine the behaviour of cf-execd,
 # including scheduling times and output capture to
 # $(sys.workdir)/outputs and relay via email.
+#
+# If different schedules are needed based on class expressions then
+# an override for this file can be specified as an augment variable:
+# `cf_execd_override_policy`. The value is a path relative to `$(sys.inputs)`
+# to load instead of `controls/cf_execd.cf`.
+#
+# ```json
+# {
+#   "vars": {
+#     "cf_execd_override_policy": "custom/cf_execd.cf"
+#   }
+# }
+# ```
+#
+# Typically a copy of the current `controls/cf_execd.cf` file should be made
+# and then modified to suit such as:
+#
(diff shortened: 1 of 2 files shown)


craigcomstock:
Thinking about this some more and proceeding on doing something similar with controls/cf_serverd.cf I am wondering if just copying the whole file over which is especially crucial for cf_serverd.cf and then customizing it really solves many problems.

When doing an upgrade you will still have to do about the same types of work whether you modify the vendored controls/*.cf or a copy/paste/modify custom/*.cf that overrides them.

The crux is that these files contain things you cannot duplicate or override in any other way.

controls/cf_execd.cf
```diff
@@ -4,6 +4,30 @@
 # These body settings determine the behaviour of cf-execd,
 # including scheduling times and output capture to
 # $(sys.workdir)/outputs and relay via email.
+#
+# If different schedules are needed based on class expressions then
+# an override for this file can be specified as an augment variable:
+# `cf_execd_override_policy`. The value is a path relative to `$(sys.inputs)`
```
craigcomstock, line 10: note: have to name it `body executor control`

nickanderson (commented):
I am not convinced that this buys us very much.
This basically alows you to specify an *alternate* file for the vendored `controls/cf_execd.cf. There is only loose coupling between body executor control and controls/cf_execd.cf.

I think it does make it marginally easier to maintain a custom `body executor control`, but I don't think it's significant.

@olehermanse, @vpodzime  thoughts?

controls/cf_execd.cf
craigcomstock: yes, but I will be closing this, recommend if customizations are needed, just change the vendored file. :+1:

craigcomstock:
Closing this as I agree with Nick (in fact I brought it up independently) that it doesn't get us much. :+1:

PR #1845, opened 2020-10-08
CFE-3451 Added apk package module support for alpinelinux

craigcomstock:
Note that repo-install and file-install API methods do not support multiple
packages. This should not affect most usage of this module.

Ticket: CFE-3451
Changelog: title

### DIFF
--- a/modules/packages/apk
+++ b/modules/packages/apk
@@ -1,0 +1,110 @@
+#!/bin/sh -e
+get_package_data() {
+  file="${INPUT_File?File must be given to get-package-data}"
+  set +e
+  apk verify $file 2>/dev/null >/dev/null
+  rc=$?
+  set -e
+  if [ 0 = $rc ]; then
+    echo "PackageType=file"
+    name=$(basename $file)
+    echo $name | sed -e 's/-/ /' -e 's/.apk/ apk/' | awk '
+{
+  printf("Name=%s\n",$1)
+  printf("Version=%s\n",$2)
+}'
+  else
+    echo PackageType=repo
+    echo Name=$file
+  fi
+}
+
+list_installed() {
+  # Example `apk list --installed` output:
+  # busybox-1.32.0-r3 x86_64 {busybox} (GPL-2.0-only) [installed]
+  #
+  # After rewrite:
+  # Name=busybox
+  # Version=1.32.0-r3
+  # Architecture=x86_64
+  apk list --installed | sed 's/-/ /' | awk '
+{
+    printf("Name=%s\n",$1)
+    printf("Version=%s\n",$2)
+    printf("Architecture=%s\n",$3)
+}'
+}
+
+repo_install() {
+  name="${INPUT_Name?Name must be given to repo-install}"
+  version="${INPUT_Version}"
+  if [ ! -z "${INPUT_Version}" ]; then
+    apk add "$name=$version" 2>/dev/null >/dev/null
+  else
+    apk add --upgrade "$name" 2>/dev/null >/dev/null
+  fi
+}
+
+file_install() {
+  file="${INPUT_File?File must be given to file-install}"
+  apk add $file 2>/dev/null >/dev/null
+}
+
+list_updates() {
+# for some odd reason --upgradable does not work where -u does
+  apk list -u | sed 's/-/ /' | awk '
+{
+    printf("Name=%s\n",$1)
+    printf("Version=%s\n",$2)
+    printf("Architecture=%s\n",$3)
+}'
+}
+
+remove() {
+  name="${INPUT_Name?Name must be given to remove}"
+  apk del "$name" 2>/dev/null >/dev/null
+}
+
+main() {
+  command=$1
+
+  # Output maybe contain backslashes, and we don't want those to end up escaping
+  # something so we use use -r with read.
+  while read -r line; do
+    # Note that line is a variable assignment, e.g.
+    # INPUT_File=syncthing
+    export INPUT_$line
+  done
+
+
+  case $command in
+    supports-api-version)
+      echo 1
+      ;;
+    get-package-data)
+      get_package_data
+      ;;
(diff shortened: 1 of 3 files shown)


Lex-2008 (changes requested):

modules/packages/apk
```diff
@@ -0,0 +1,91 @@
...
+
+main() {
+  command=$1
+
+  # Output maybe contain backslashes, and we don't want those to end up escaping
+  # something so we use use -r with read.
+  while read -r line; do
+    # Note that line is a variable assignment, e.g.
+    # INPUT_File=syncthing
+    export INPUT_$line
```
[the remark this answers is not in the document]
craigcomstock: so true. well done. We should mark this as an issue with `snap` module as well then, as this is a copy/paste/modify :(

modules/packages/apk
```diff
@@ -0,0 +1,91 @@
+#!/bin/sh -e
+get_package_data() {
+  name="${INPUT_File?File must be given to get-package-data}"
+  echo PackageType=repo
+  echo Name=$name
+}
+
+list_installed() {
+  # Example pkg output:
+  # Name         Version  Rev   Developer  Notes
```
nickanderson, line 10: thanks for including this.

modules/packages/apk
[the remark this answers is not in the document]
craigcomstock: Interesting question though, is when does this actually happen?

Each packages promise can only specify one package name and each is processed individually as far as I can tell.

I'll take a look at the core C code a bit to see if I can tell when multiples would be sent.

modules/packages/apk
```diff
@@ -0,0 +1,91 @@
...
+  # Architecture=none
+  apk list --installed | sed 's/-/ /' | awk '
+{
+    printf("Name=%s\n",$1)
+    printf("Version=%s\n",$2)
+    printf("Architecture=%s\n",$3)
+}'
+}
+
+repo_install() {
```
nickanderson, line 26: I see repo_installl, what about file_install if I have an apk locally on the file system?

modules/packages/apk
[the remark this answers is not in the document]
Lex-2008: good point! Maybe docs need to be updated

nickanderson (changes requested):
Looks pretty good to me, just a few things.
1) Can we have a ticket to link this to
2) Looks like file based source isn't implemented?

modules/packages/apk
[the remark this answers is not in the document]
nickanderson: Good find @Lex-2008, @kacf might know about why it needs to handle mutlple.

modules/packages/apk
Lex-2008: have a look at `get_package_data()` function - it unconditionally considers all packages to be coming from repo - so installation from files is not supported.

modules/packages/apk
```diff
@@ -0,0 +1,91 @@
...
+      ;;
+    get-package-data)
+      get_package_data
+      ;;
+    list-installed)
+      list_installed
+      ;;
+    repo-install)
+      repo_install
+      ;;
```
craigcomstock, line 76: Note also that `file-install` is not supported. Ditto for `snap` module.

https://docs.cfengine.com/docs/3.12/reference-package_modules.html#file-install

modules/packages/apk
craigcomstock: I commented elsewhere about this. get_package_data() would need to be refined. ack.

modules/packages/apk
craigcomstock: s/pkg/apk/ and include actual output from `apk` :)

modules/packages/apk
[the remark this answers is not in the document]
kacf: I *think* that it only plays a role if you are installing one package for multiple architectures (quite common in the i386->amd64 transition, but getting less and less common). The versions of these packages cannot ever conflict, so it will try to install both types at once. See [this section](https://github.com/cfengine/core/blob/master/cf-agent/package_module.c#L1232-L1242).

I think the theory was also that we could optimize by installing slists of packages, but I don't think this was ever implemented.

modules/packages/apk
craigcomstock: implemented below. :+1:

modules/packages/apk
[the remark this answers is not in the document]
craigcomstock: So @Lex-2008 and @nickanderson do you want to move forward with as-is or implement supporting multiple packages?

craigcomstock:
Will squash when review is complete. Thanks y'all!

modules/packages/apk
```diff
@@ -0,0 +1,110 @@
...
+  if [ ! -z "${INPUT_Version}" ]; then
+    apk add "$name=$version" 2>/dev/null >/dev/null
+  else
+    apk add --upgrade "$name" 2>/dev/null >/dev/null
+  fi
+}
+
+file_install() {
+  file="${INPUT_File?File must be given to file-install}"
+  apk add $file 2>/dev/null >/dev/null
```
nickanderson, line 50: cool beans

nickanderson (changes requested):
yay for file based install. Now, I think it just needs a ticket to correlate with in jira.

modules/packages/apk
craigcomstock: CFE-3451, in title and will be in squashed single commit in a moment.

nickanderson (approved):
LGTM

modules/packages/apk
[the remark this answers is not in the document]
Lex-2008: Right, since it wasn't implemented, but was working, before - it should be fine to keep it like this and maybe file a ticket for later.

Lex-2008 (approved):

PR #1851, opened 2020-10-19
CFE-3469/3.12.x: Replaced curl with url_get for AWS ec2 inventory

nickanderson:
Ticket: CFE-3469
Changelog: Title


craigcomstock (approved):

nickanderson:
Oh, silly me. 3.7.x was still under support (supported until August 31st 2018) when 3.12.0 was released (June 28th 2018). So, no, we can't use url_get in 3.12.x. Closing without merging.

PR #1853, opened 2020-10-27
Fix converge bug with anchoring and add converge_prepend

tzz:
In my testing with 3.16, the deletion doesn't work today, so `converge` is not functioning properly.

Add `converge_prepend` with same behavior and bugfix, but inserting at start of content.

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -1077,10 +1077,18 @@ bundle edit_line converge(marker, lines)
 }
 
 bundle edit_line converge_prepend(marker, lines)
-# @brief Converge `lines` marked with `marker` to start of content
+# @brief Converge `lines` marked with `marker` to start of content.
 #
 # Any content marked with `marker` is removed, then `lines` are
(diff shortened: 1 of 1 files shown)


lib/files.cf
```diff
@@ -1071,11 +1071,30 @@ bundle edit_line converge(marker, lines)
...
 
+bundle edit_line converge_prepend(marker, lines)
+# @brief Converge `lines` marked with `marker` to start of content
+#
+# Any content marked with `marker` is removed, then `lines` are
+# inserted at *start* of content.  Every `line` should contain `marker`.
+#
+# @param marker The marker (not a regular expression; will be escaped)
+# @param lines The lines to insert; all must contain `marker`
+{
```
nickanderson, line 1087: Can you put a block with example usage. What do you use this with?

nickanderson (changes requested):

lib/files.cf
tzz: pam.d edits, for instance, where you need something at the very top

PR #1854, opened 2020-10-27
ENT-6225 Enable ec2 metadata cache on windows

craigcomstock:


lib/paths.cf
```diff
@@ -650,6 +650,10 @@ bundle common paths
       "tmp_path[$(tmp_paths)]" string => "$(files_path)$(path[$(tmp_paths)])";
       "path[$(tmp_paths)]" string => "$(tmp_path[$(tmp_paths)])";
 
+    windows:
+      "path[true]"          string => "exit 0"; # designed to be used with contain => in_shell
```
nickanderson, line 654: Hummmm, I dunno about htis one. what if it was used with powershell instead? also, i don't think that it will get the _stdlib_path_exists_true class since there is no file at path "exit 0". and the dummy true commands only run if that is defined.

inventory/any.cf
```diff
@@ -608,6 +608,7 @@ bundle agent cfe_autorun_inventory_aws_ec2_metadata_cache
         comment => "This promise is used for delaying execution of url_get, since locking does not work directly on vars type promises",
         classes => kept_successful_command,
         handle => "daily_dummy_job_suppress_inform_incapable",
+        contain => in_shell,
```
nickanderson, line 611: instead of running paths.true in a shell, perhaps we set a local variable for true and use that instead.

vars:
  "_true_command" string => ifelse( "windows", '$(sys.winsysdir)\cmd.exe /q /c "(exit 0)"', $(paths.true) );

nickanderson (changes requested):
I don't think that adding a bad path is the right thing to do, good idea to enable the cache, but I have commented with a suggestion for a slightly different tactic.

inventory/any.cf
craigcomstock: Right. The real problem here is preventing url_get from firing every agent run.

Would it make more sense to change this back to a wget commands promise which we could limit with `if_elapsed_day`? That would bring us back to having to have a different command for Windows as we did before.

nickanderson: If we brought it back to using external tool I think it would 1, be more expensive than multple url_get calls and 2, make us hunt for an equivlant command to run to get the data.

What do you think about that suggestion of faking true out just inside this bundle to run cmd.exe /q /c exit 0?

craigcomstock: This seems like a good idea. I'll try it! :)

craigcomstock:
stale

PR #1862, opened 2020-10-29
CFE-3483/3.12.x: Added bundle edit_line converge_prepend

nickanderson:


lib/files.cf
```diff
@@ -1085,6 +1085,59 @@ bundle edit_line converge(marker, lines)
...
+#       "/etc/pam.d/su"
+#         edit_line => converge_prepend( "session", "session       required   pam_env.so readenv=1 envfile=/etc/default/locale
+# session    optional   pam_mail.so nopen
+# session    required   pam_limits.so" );
+# }
+# ```
+#
+# **History:**
+#
+# * Introduced in 3.17.0
```
craigcomstock, line 1129: This "Introduced in 3.17.0" seems a bit odd in 3.12.x but I understand it. :)

PR #1870, opened 2020-11-24
Added redirect to remove index.php from Mission Portal's URL

aleksandrychev:
ChangeLog: Title

Ticket: ENT-6464

### DIFF
--- a/cfe_internal/enterprise/mission_portal.cf
+++ b/cfe_internal/enterprise/mission_portal.cf
@@ -55,6 +55,15 @@ bundle agent cfe_internal_enterprise_mission_portal_apache
   reports:
     DEBUG|DEBUG_cfe_internal_enterprise_mission_portal_apache::
       "DEBUG $(this.bundle): Should stage '$(staged_config)' from '$(template)'.";
+
+  classes:
+
+    (policy_server|am_policy_hub).enterprise_edition::
+      "mission_portal_index_php_redirect_enabled"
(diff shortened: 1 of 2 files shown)


aleksandrychev:
@nickanderson, please review

cfe_internal/enterprise/mission_portal.cf
```diff
@@ -55,6 +55,15 @@ bundle agent cfe_internal_enterprise_mission_portal_apache
   reports:
     DEBUG|DEBUG_cfe_internal_enterprise_mission_portal_apache::
       "DEBUG $(this.bundle): Should stage '$(staged_config)' from '$(template)'.";
+
+  classes:
+
+    (policy_server|am_policy_hub).enterprise_edition.active_hub::
```
nickanderson, line 61: mmmm active_hub may only be set on the hub if its running in HA AND if its not the slave. 

So, if you want this to apply on a hub that is not necessarily in ha I would drop the active_hub class, or say .!passive_ha_hub to say any hub that isn't passive.

nickanderson (changes requested):
Looks OK, but i think the conditon for editing the config needs to be adjusted.

cfe_internal/enterprise/mission_portal.cf
aleksandrychev: ah, I thought active_hub is one more validation that hub is OK =) I will remove that

aleksandrychev: @nickanderson done

nickanderson (approved):

PR #1871, opened 2020-11-24
CFE-3504/master: Enabled packages promises using package_module without bundle def

nickanderson:
This change moves the default settings from bundle def to bundle
package_module_knowledge, enabling standalone policy to include the stdlib and
make packages promises implemented by package modules. Before this change,
errors would be emitted because query_installed_ifelapsed and
query_updates_ifelapsed were not set.

This change does not inhibit the ability to configure these settings via
augments (def.json) by setting package_module_query_installed_ifelapsed and
package_module_query_updates_ifelapsed.

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -183,15 +183,6 @@ bundle common def
         int => "200",
         if => not( isvariable( "control_server_maxconnections" ) );
 
-      # Package inventory refresh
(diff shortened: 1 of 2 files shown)


vpodzime (approved):
ACK. Nice!

PR #1873, opened 2020-12-02
CFE-413: set_variable_values_ini for values with equal sign

ncharles:
Allow equal sign in set_variable_values_ini

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -423,6 +423,8 @@ bundle edit_line set_variable_values_ini(tab, sectionName)
 # If the line is commented out with `#`, it gets uncommented first.
 # Adds a new line if none exists.
 #
+# It does support = in value
+#
 # @param tab An associative array containing `tab[sectionName][LHS]="RHS"`.
 # The value is not changed when the `RHS` is "dontchange"
 # @param sectionName The section in the file within which values should be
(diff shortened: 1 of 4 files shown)


ncharles:
so this fixes CFE-413, but breaks CFE-3221

lib/files.cf
```diff
@@ -436,33 +438,43 @@ bundle edit_line set_variable_values_ini(tab, sectionName)
...
-      edit_field => col("\s*=\s*","1","$(index)","set"),
-      if => "edit_$(cindex[$(index)])";
-
-      # match a line starting like the key something
-      "\s*$(index)\s*=.*"
-      edit_field => col("\s*=\s*","2","$($(tab)[$(sectionName)][$(index)])","set"),
-      select_region => INI_section(escape("$(sectionName)")),
-      classes => results("bundle", "set_variable_values_ini_not_$(cindex[$(index)])"),
-      if => "edit_$(cindex[$(index)])";
+      "pass2" expression => "pass1";
```
nickanderson, line 448: Ewwwww. Why are you forcing a specific pass restriction?

Presumably, something isn't yet defined or hasn't been completed. Or, not able to select region until the second pass?

nickanderson (commented):
Unsure about the breakage it introduces, will need to dig a bit deeper, but I do think we should try and avoid the explicit restriction to second pass, surely there is something we can fix to make that un-necessary.

lib/files.cf
ncharles: Exactly. If it's on the first pass, if the section doesn't exist yet, it cannot select the region, and fails. It won't reevaluate at the 2nd pass

nickanderson: Right, but since the promise to insert the section is before the promise to set key values for the region, it should be able to select it during the same pass.

PR #1874, opened 2020-12-02
ENT-4094/master: Fixed self upgrade for windows

nickanderson:


nickanderson:
These should all be squashed, but early review and testing appricated.

vpodzime (approved):
I'd put the template into a separate file (does it even have to be a template?), but this looks good to me otherwise. I wouldn't squash the commits, more details in the history are welcome.

nickanderson:
It doesn't HAVE to be a template, I can move it to an external template.

nickanderson:
Alright, I moved the script to an external file. It's still rendered (unnecessarily)  instead of being copied, but that is because I didn't want to add a copy_from body into standalone_self_upgrade.cf

vpodzime (approved):
ACK

Lex-2008 (approved):
I think commits "Added external version comparison tool for windows", "Moved version comparison template to external file", and "Hand the template no data" can be squashed together.

Also, I'm not very good a PS scripting - does the version comparison script currently print these `"$version1 is less than $version2"` lines? I'd expect the script to print the "PROGRAMMING ERROR" message, but not others. But if you prefer otherwise - no need to change :)

olehermanse (approved):

nickanderson:
@Lex-2008 Yes, it does print those however it's not seen in any output as far as I can tell. I left it printing output because it was much easier to understand when doing manual testing.

I will squash the last 3 commits together.

PR #1876, opened 2020-12-06
CFE-3513/master: query service status on FreeBSD using onestatus

systemheld:
If a service is not enabled in rc.conf on FreeBSD, trying to get the status always results in something like:

```
# /etc/rc.d/ntpd status
Cannot 'status' ntpd. Set ntpd_enable to YES in /etc/rc.conf or use 'onestatus' instead of 'status'.
# echo $?
0
```

As the exit code is zero, cf-agent would try to stop this service. Using onestatus always returns the wanted exit code, regardless if the service is enabled or not. 

```
# /etc/rc.d/ntpd onestatus
ntpd is not running.
# echo $?
1
```

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -171,7 +171,10 @@ bundle agent standard_services(service,state)
 
       "have_init" expression => fileexists($(init));
 
-    chkconfig.have_init::
+    chkconfig.have_init.freebsd::
+      "running" expression => returnszero("$(init) onestatus > /dev/null", "useshell");
+      
+    chkconfig.have_init.!freebsd::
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
Hi @systemheld 

Thanks for your contribution!

It looks simple enough and OK to me, but perhaps @BasementTrix or @skreuzer  could give it a once over.

skreuzer:
Seems reasonable to me

nickanderson (approved):

nickanderson:
This one was merged in https://github.com/cfengine/masterfiles/pull/1878

PR #1879, opened 2020-12-10
CFE-3510: Added init.cf for defining custom promise types

olehermanse:


olehermanse:
Goes with: https://github.com/cfengine/core/pull/4427

PR #1880, opened 2020-12-10
CFE-3501/master: services/init.cf will be loaded automatically by promises.cf if it's present

nickanderson:
services/init.cf provides a location for users to define custom promise types.
It will be loaded very early (with controls) if found.

Ticket: CFE-3510
Changelog: Title


nickanderson:
After thinking about it more i think yours in #1879 is better. Its more simple, and it advertizes the capability and where to do it.

PR #1885, opened 2020-12-11
ENT-6603/master: Added ability to specify a list of bundles to run before autorun (for classification)

nickanderson:
With this change, you can specify a list of bundles to run before autorun
polices. This can be useful if you want to have autorun policies execute based
on soft classes that are not persistent. E.g. prod hosts.

Ticket: ENT-6603
Changelog: Title

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1079,7 +1079,27 @@ Primarily for developer conveniance, this setting allows you to easily disable t
 }
 ```
 
-### Append to the main bundlesequence
+### Bundlesequence
+
+#### Classification bundles before autorun
+
+You can specify a list of bundles which should be run before autorun policies (if enabled).
+
+```json
+{
+  "vars":{
+    "control_common_bundlesequence_classification": [ "classification_one", "classification_two" ]
+  },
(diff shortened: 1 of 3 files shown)


bim9262:
@nickanderson, this looks great!

nickanderson:
@olehermanse bikeshed name?

MPF.md
```diff
@@ -1079,7 +1079,27 @@ Primarily for developer conveniance, this setting allows you to easily disable t
...
+### Bundlesequence
+
+#### Classification bundles before autorun
+
+You can specify a list of bundles which should be run before autorun policies (if enabled).
+
+```json
+{
+  "vars":{
+    "control_common_bundlesequence_classification": [ "classification_one", "classification_two" ]
```
olehermanse, line 1091: I think the name is fine, could also use a name which describes exactly when the bundles are run, rather than what to put in them.

```
control_common_bundlesequence_pre_autorun
```

I think I prefer your name.

olehermanse (approved):

PR #1899, opened 2021-01-15
CFE-3544/master: Added new interface for controlling users allowed to initiate cf-agent via cf-runagent

nickanderson:
This change makes def.control_server_allowusers the singular variable used to
control the list of users allowed to execute policy via cf-runagent. It
supersedes def.control_server_allowusers_policy_server and
def.control_server_allowusers_non_policy_server which are still used to set the
default value for def.control_server_allowusers. If
vars.control_server_allowusers is defined via Augments (def.json) that setting
will be used for both policy servers and non policy servers.

If further segmentation is desired, multiple augments can be used.

Ticket: CFE-3544
Changelog: Title

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -92,7 +92,20 @@ bundle common def
         slist => {},
         if => not( isvariable( "control_server_allowusers_policy_server" ) );
 
-      # Executor Controls
+    policy_server::
+      "control_server_allowusers" -> { "CFE-3544" }
+        handle => "def_control_server_allowusers_policy_server",
+        slist => { @(control_server_allowusers_policy_server) },
+        if => not(isvariable("control_server_allowusers"));
+
+    !policy_server::
+      "control_server_allowusers" -> { "CFE-3544" }
+        handle => "def_control_server_allowusers_non_policy_server",
+        slist => { @(control_server_allowusers_non_policy_server) },
+        if => not(isvariable("control_server_allowusers"));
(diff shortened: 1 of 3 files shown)


controls/def.cf
```diff
@@ -92,7 +92,19 @@ bundle common def
         slist => {},
         if => not( isvariable( "control_server_allowusers_policy_server" ) );
 
-      # Executor Controls
+    policy_server::
+      "control_server_allowusers" -> { "CFE-3544"}
```
olehermanse, line 96: Bad formatting, missing space before `}`.

controls/def.cf
```diff
@@ -92,7 +92,19 @@ bundle common def
...
+        handle => "def_control_server_allowusers_policy_server",
+        slist => { @(control_server_allowusers_policy_server) },
+        if => not(isvariable("control_server_allowusers"));
+
+    !policy_server::
+      "control_server_allowusers" -> { "CFE-3544"}
```
olehermanse, line 102: Bad formatting, missing space before `}`.

olehermanse (approved):
LGTM, just some small mistakes on the formatting.

controls/def.cf
nickanderson: ack

controls/def.cf
nickanderson: ack

PR #1900, opened 2021-01-15
CFE-3546/master: Fixed ability to define users authorized for using cf-runagent on policy servers

nickanderson:
Ticket: CFE-3546
Changelog: Title


craigcomstock (approved):
Right. Set a default empty list if the variable doesn't exist. :+1:

PR #1903, opened 2021-01-19
ENT-6673/master: Added vars.mpf_admit_cf_runagent_shell to control admission for cf-runagent requests

nickanderson:
This change adds a new augmentable variable to control hosts that should be
allowed to initiate agent runs via cf-runagent.

Ticket: ENT-6673
Changelog: Title


olehermanse (approved):
Thanks, LGTM ( I assume you tested it ).

nickanderson:
> Thanks, LGTM ( I assume you tested it ).

Indeed, details of testing in the ticket.

PR #1907, opened 2021-02-03
ENT-4669 Use VBScript to enumerate installed packages

Lex-2008:


nickanderson (commented):
Well, it seems like an improvement because it is faster (I think, i didn't measure). But it's definitely not providing an exhaustive inventory of installed stuff on windows. Perhaps we need multiple modules to get proper exaustive inventory for windows software. I commented on the ticket with more details.

vpodzime (approved):
Looks good to me.

modules/packages/msiexec-list.vbs
```diff
@@ -0,0 +1,47 @@
...
+For Each strKey in strKeys
+  ' Enumerate the subkeys
+  objReg.EnumKey HKLM, strKey, arrSubkeys
+  ' Iterate through each subkey
+  For Each strSubkey In arrSubkeys
+    intRet1 = objReg.GetStringValue(HKLM, strKey & strSubkey, strDisplayNameEntryA , strDisplayNameValue)
+    ' Check return value
+    If intRet1 <> 0 Then
+      ' Read the Displayname
+      objReg.GetStringValue HKLM, strKey & strSubkey, strDisplayNameEntryB, strDisplayNameValue
```
craigcomstock, line 25: would you want to handle/notify about errors on the GetStringValue calls from this line 25 down? What will happen if they fail? Just missing/empty strings for those values?

craigcomstock (approved):
looks good, just wondering about error handling/behavior on failed function calls

olehermanse (approved):
LGTM, please test on Windows Server 2012, 2016 and 2019, and see that inventory shows up in MP.

craigcomstock:
> LGTM, please test on Windows Server 2012, 2016 and 2019, and see that inventory shows up in MP.

I checked with this PR MPF content in AWS and all three show inventory of 5 or 6 packages each including CFEngine Nova. I believe this PR is good and this ticket is done.

I will work on extending the software found in ENT-2553.

modules/packages/msiexec-list.vbs
Lex-2008: Good point! However I believe "version" and "install date" are optional - we shouldn't fail if they're not set in registry.

[The GetStringValue docs][d] don't specify what happens to the output argument in case of error. So let's set all variables to empty strings at the beginning of each loop iteration.

[d]: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/regprov/getstringvalue-method-in-class-stdregprov

Regarding this particular line - I believe it was assumed that in case of error the output variable "strDisplayNameValue" is set to empty string, since this check is performed afterwards. But since it's not documented, let's add an explicit error checking here, and for optional values follow the pattern of "using whatever GetStringValue set this variable to, even if it returned with error"

craigcomstock (approved):

PR #1913, opened 2021-02-12
CFE-3583 Fix pip package method if only pip3 command is available (alpine)

craigcomstock:
On alpine lack of pip broke the pip package method.

Ticket: CFE-3583
Changelog: title

### DIFF
--- a/lib/paths.cf
+++ b/lib/paths.cf
@@ -104,7 +104,9 @@ bundle common paths
 
     !(freebsd|darwin|smartos)::
       "path[npm]"      string => "/usr/bin/npm";
-      "path[pip]"      string => "/usr/bin/pip";
+      "path[pip]"      string => ifelse(
+        fileexists("/usr/bin/pip3"), "/usr/bin/pip3",
+        "/usr/bin/pip");
       "path[virtualenv]" string => "/usr/bin/virtualenv";
 
     !(freebsd|darwin)::


cf-bottom (changes requested):
Please use a company e-mail instead of c***@unreasonablefarm.org

lib/paths.cf
```diff
@@ -104,7 +104,9 @@ bundle common paths
 
     !(freebsd|darwin|smartos)::
       "path[npm]"      string => "/usr/bin/npm";
-      "path[pip]"      string => "/usr/bin/pip";
+      "path[pip]"      string => ifelse(
+        fileexists("/usr/bin/pip3"), "/usr/bin/pip3",
+        "/usr/bin/pip");
```
nickanderson, line 109: Is there ever any reason that both pip and pip3 would be present concurrently (it's the case on my system).

Is there ever any reason that someone would actually use both pip and pip3 concurrently? If so, seems more appropriate to introduce pip3.

nickanderson (commented):
I am not sure if there can really be only one pip.

lib/paths.cf
craigcomstock: Indeed. So then if they are the same, pip is a symlink to pip3 or if they are different, pip is pip2 and pip3 is pip3, then the legacy pip package promise might need to have an attribute about which pip to use?

As it is, the legacy pip package handler relies on this specific `$(paths.pip)` so that would need to be refactored somewhat to support choosing?

https://github.com/cfengine/masterfiles/blob/master/lib/packages.cf#L363

The actual package support is here and uses `$(paths.pip)` https://github.com/cfengine/masterfiles/blob/master/lib/packages.cf#L407

So I would suppose that we could add a pip3 path, but set `paths.pip` to pip3 if pip3 is available?

nickanderson (commented):
I guess I never submitted my last comment?

craigcomstock:
Closing this PR as I agree with @nickanderson that we should instead implement a separate package method for pip3: https://tracker.mender.io/browse/CFE-3697

PR #1915, opened 2021-02-15
CFE-3585 Fixed alpine apk packages module to parse names properly

craigcomstock:
package names can include dashes(-) so sed expression was incorrect.

Ticket: CFE-3585
Changelog: title


craigcomstock:
oops, I guess I didn't have the latest from MPF in my own masterfiles repo. Will re-work :)

PR #1916, opened 2021-02-15
CFE-3585 Fixed alpine apk packages module to parse names properly

craigcomstock:
package names can include dashes(-) so sed expression was incorrect.

Ticket: CFE-3585
Changelog: title

### DIFF
--- a/modules/packages/apk
+++ b/modules/packages/apk
@@ -27,7 +27,7 @@ list_installed() {
   # Name=busybox
   # Version=1.32.0-r3
   # Architecture=x86_64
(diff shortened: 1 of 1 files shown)


nickanderson (approved):

craigcomstock:
alpine edge is at 3.17.0, v3.13 at 3.16.0, v3.12 at 3.15.1, @olehermanse seems like we should just plan on 3.18 LTS having better alpine linux support and get it into v3.13 or v3.14 if that comes around the same time? I don't want to make a branch of 3.16 or 3.17 with this change. :)

PR #1918, opened 2021-02-22
ENT-6777/master: Added policy for cf-execd sockets

nickanderson:


MPF.md
```diff
@@ -738,6 +738,22 @@ Example configuration via augments:
 
 The above configuration would result in exec_command being triggered at the top and half hour and sleeping for up to `splaytime` before agent execution.
 
+### Configure cf-execd runagent socket users
+
+On Enterprise hubs, access to cf-execd sockets can be configure a list of users who should be allowed by defining `vars.control_executor_runagent_socket_allow_users`. By default on Enterprise Hubs, `cfapache` is allowed to access runagent sockets.
```
vpodzime, line 743: It can be configure**d** on all hosts, not just on Enterprise hubs. It's just the `cfapache` as default that is different on Enterprise hubs.

cfe_internal/enterprise/CFE_knowledge.cf
```diff
@@ -221,6 +221,29 @@ bundle agent cfe_internal_permissions
 # @brief Specific expectations for permissions and ownership of CFEngine with respect to the Enterprise Edition
 {
 
+  vars:
+
+    # nonstandard Directoires in statedir
```
vpodzime, line 226: *Directories*

controls/def.cf
```diff
@@ -162,6 +162,15 @@ bundle common def
...
 
 
+      # Users allowed on sockets
+      "control_executor_runagent_socket_allow_users"
+        slist => { "cfapache" },
+        unless => isvariable( $(this.promiser) );
+
+      "control_executor_runagent_socket_allow_users"
+        slist => { @(def.control_executor_runagent_socket_allow_users) },
+        if => isvariable( $(this.promiser) );
```
vpodzime, line 172: Shouldn't this be `if => isvariable("def.control_executor_runagent_socket_allow_users")`?

vpodzime (commented):
Looks good to me otherwise. Let's see if it works.

vpodzime:
>  Let's see if it works.

```
Feb 23 10:12:11 localhost cf-execd[5312]: CFEngine(executor)  Failed to get UID for user '@(cfe_internal_permissions._cf_execd_socket_allow_users)': Success
Feb 23 10:12:11 localhost cf-execd[5312]: CFEngine(executor)  Failed to allow new runagent_socket_allow_users users access the runagent socket (on policy reload)
```
So that's a NO to the question whether *cf-execd* will be able to resolve the variable. Also note that I replaced `$()` with `@()`, it should be a list expansion, right?

controls/cf_execd.cf
```diff
@@ -16,6 +16,8 @@ body executor control
 
       agent_expireafter  => "$(def.control_executor_agent_expireafter)";
 
+      "runagent_socket_allow_users" => { "$(cfe_internal_permissions._cf_execd_socket_allow_users)" };
```
vpodzime, line 19: The body attribute name cannot be quoted.

vpodzime:
The permissions and ACLs work as expected (if I replace the unexpanded variable in the body with `{ "cfapache" }`) :+1:

MPF.md
nickanderson: Yeah, I worded it this way because in the MPF where the policy applies the ACLs to the directory and socket, that is only done on enterprise hubs. Yes, if you configure it via augments, cf-execd will set perms, but cf-agent won't be enforcing it unless it's an enterprise hub.

cfe_internal/enterprise/CFE_knowledge.cf
nickanderson: ACK.

controls/def.cf
nickanderson: Well, that's what this.promiser expands to (sans leading `def.`), but it should work with this.promiser. I will see if there is any difference when I test the build.

controls/cf_execd.cf
nickanderson: oooops!

nickanderson: Also, i shouldn't iterate on that list, I should @() it.

vpodzime (commented):
Looks good to me and seems to work as expected. :+1: Please squash the commits.

nickanderson:
@vpodzime as-you-wish

MPF.md
vpodzime, line 743: *configure**d as a list***

cfe_internal/enterprise/CFE_knowledge.cf
```diff
@@ -221,6 +221,29 @@ bundle agent cfe_internal_permissions
...
+#+BEGIN_SRC question
+    # Should we respect this Augment?
+    # - Will cf-execd be able to resolve the variable
+    #   $(cfe_internal_permissions._cf_execd_socket_allow_users) for it to be
+    #   used by the daemon?
+
+    # Derive the users that should be able to access cf-execd sockets
+    "_cf_execd_socket_allow_users"
+      slist => { @(def.control_executor_runagent_socket_allow_users) };
+#+END_SRC
```
vpodzime, line 239: I guess this question can go away now?

controls/def.cf
vpodzime: It's still unclear to me how this works. The first promise defines a variable unless it is defined, the latter re-defines a variable if it is already defined?

vpodzime (commented):
Thanks. Looks good to me otherwise.

cfe_internal/enterprise/CFE_knowledge.cf
nickanderson: Gone.

MPF.md
nickanderson: ACK.

nickanderson:
@vpodzime, I yanked the unnecessary comments, removed the unnecessary setting of the variable if it's already set, and also added a macro to protect against setting the control body option unless the version is 3.18.0 or higher, else we would have backward compatibility issues.

controls/def.cf
nickanderson: the second one was not necessary.

vpodzime:
With the following contents of `def.json`, I don't get the expected ACLs:
```
[root@hub vagrant]# cat /var/cfengine/masterfiles/def.json
{
  "vars": {
    "control_executor_runagent_socket_allow_users:" [ "cfapache", "cfpostgres" ]
  }
}
[root@hub vagrant]# getfacl /var/cfengine/state/cf-execd.sockets{,/runagent.socket}
getfacl: Removing leading '/' from absolute path names
# file: var/cfengine/state/cf-execd.sockets
# owner: root
# group: root
user::rwx
user:cfapache:r-x
group::---
mask::r-x
other::---

# file: var/cfengine/state/cf-execd.sockets/runagent.socket
# owner: root
# group: root
user::rw-
user:cfapache:rw-
group::---
mask::rw-
other::---
```
build: http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6143/

I also had to remove the `@if minimum_version(3.18.0)` macro because apparently it evaluates to `false` on 3.18.0a.?

vpodzime:
I've merged the PR handling things on the *cf-execd* side. I'll rebuild this and check again if it works as expected or not. There also seems to be a problem with *vagrant-checks* caused by this PR which will need to be handled (in the *vagrant-checks* job, probably).

controls/def.cf
```diff
@@ -161,6 +161,11 @@ bundle common def
         comment => "This variable defines the list of classes that should
                     trigger exec_command if any of them are defined.";
 
+      # Users allowed on sockets
+      # Set the users allowed by default, if not already set (e.g. via augments)
+      "control_executor_runagent_socket_allow_users"
+        slist => { "cfapache" },
```
olehermanse, line 167: cfapache user only exists for enterprise hubs. There is no class guard, so this will fail everywhere else.

olehermanse (changes requested):

vpodzime:
Rebuild with current *master* core: http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6150/

nickanderson:
> With the following contents of `def.json`, I don't get the expected ACLs:
> 
> ```
> [root@hub vagrant]# cat /var/cfengine/masterfiles/def.json
> {
>   "vars": {
>     "control_executor_runagent_socket_allow_users:" [ "cfapache", "cfpostgres" ]
>   }
> }
> ```

Yes, the JSON referenced is *invalid*.

Note, your colon (`:`) was inside the quotes. Your JSON should look like this:

{
  "vars": {
    "control_executor_runagent_socket_allow_users": [ "cfapache", "cfpostgres" ]
  }
}

Do you know what's a bit strange and unexpected though? The fact that CFEngine somehow tried to handle your typo and actually ended up defining `control_executor_runagent_socket_allow_users:`. This output illustrates this oddity.

```
[root@hub masterfiles]# cf-promises --show-vars=control_executor_runagent_socket
Variable name                            Variable value                                               Meta tags                               
default:def.control_executor_runagent_socket_allow_users  {"cfapache"}                                                source=promise                          
default:def.control_executor_runagent_socket_allow_users:  {"cfapache","vagrant"}                                      source=augments_file                    
```

> I also had to remove the @if minimum_version(3.18.0) macro because apparently it evaluates to false on 3.18.0a.?

I do not find this when I re-test. Both 3.18 and 3.18.0 work equally well for me.

controls/def.cf
nickanderson: I can restrict this variable to be set only when the default user actually exists, then I can define a class, e.g. `have_control_executor_runagent_socket_allow_users` if `def.control_executor_runagent_socket_allow_users` is defined and use `have_control_executor_runagent_socket_allow_users::` to restrict setting the list of users in the control body to when we have a list of users defined.

nickanderson:
@vpodzime  I fixed the docs so that you don't define invalid json by copy pasting example.

@olehermanse I implemented a context restriction. Now, the default socket user is only set to cfapache if that user actually exists, which is presumably only true on Enterprise Hub installs. In order to prevent errors from cf-execd about not being able to find user in the event that no default list of users is defined, there is a context restricting that control attribute based on the presence of a class defined when we do have a variable defined.

controls/def.cf
olehermanse: I think I would prefer if it was based on enterprise and hub classes, rather than if user exists. But I guess it doesn't matter too much.

olehermanse: Actually I don't know what I prefer, no strong opinion.

nickanderson: alright, i switched to using classes, squashed

vpodzime (approved):
Works great and the *valgrind-checks* job passes. :+1:

cfe_internal/enterprise/CFE_knowledge.cf
```diff
@@ -221,6 +221,22 @@ bundle agent cfe_internal_permissions
...
+
+    # Derive the users that should be able to access cf-execd sockets
+    "_cf_execd_socket_allow_users"
+      slist => { @(def.control_executor_runagent_socket_allow_users) };
+
+    # Derive the aces for users allowed to access cf-execd sockets
+    "_cf_execd_socket_dir_aces" -> { "ENT-6777" }
+      slist => maplist( "user:$(this):rx:allow", @(_cf_execd_socket_allow_users) );
+    "_cf_execd_socket_runagent_aces" -> { "ENT-6777" }
+      slist => maplist( "user:$(this):rw:allow", @(_cf_execd_socket_allow_users) );
```
vpodzime, line 238: I've only just now found out that we need to set the `r-x` ACLs on the state directory too. *cfapache* and *cfpostgres* are special users for which this is not necessary because the state directory is group-owned by the *cfpostgres* group which includes the *cfapache* user. So these two users can access the state directory, but other users cannot.

vpodzime (changes requested):
Please add the policy for setting the ACLs on the state directory itself.

cfe_internal/enterprise/CFE_knowledge.cf
nickanderson: OK, I have added ACLs for sys.statedir https://github.com/cfengine/masterfiles/pull/1918/commits/2052313c121a65da677da49a31c9d0c8acf806cd

cfe_internal/enterprise/CFE_knowledge.cf
```diff
@@ -231,11 +231,17 @@ bundle agent cfe_internal_permissions
     "_cf_execd_socket_allow_users"
       slist => { @(def.control_executor_runagent_socket_allow_users) };
 
+    "_cf_statedir_allow_users"
+      slist => { "cfpostgres", "cfapache", @(def.control_executor_runagent_socket_allow_users) },
```
vpodzime, line 235: Do we need to de-duplicate this list or is it okay if it has duplicates? I guess you tested the defaults which actually have a duplicate `cfapache` item in the list?

vpodzime (commented):
Looks good to me otherwise. Please squash the commits.

cfe_internal/enterprise/CFE_knowledge.cf
nickanderson: In my testing, it was fine to be duplicated.

nickanderson:
@vpodzime, squashed.

vpodzime (approved):
Works great and looks good to me. Thanks for working on this!

olehermanse (approved):

PR #1920, opened 2021-02-26
ENT-6220 Added verbose logfile for msiexec package module file installs

craigcomstock:
File is always \cfengine_package_install.log

See ENT-6824 TODO to save based on basename of msi file.

Ticket: ENT-6220
Changelog: Title

### DIFF
--- a/modules/packages/msiexec.bat
+++ b/modules/packages/msiexec.bat
@@ -96,8 +96,12 @@ rem Install this file if it exists
     goto :EOF
   )
 
-  %MSIEXEC% /quiet /passive /qn /norestart /i %1
-  rem TODO options, error checking
+  REM TODO: ENT-6824 save this logfile based on msi filename
+  set logfile="\cfengine_package_install.log"
+  %MSIEXEC% /quiet /passive /qn /norestart /l*vx %logfile% /i %1
+  if not errorlevel 0 (
+    echo ErrorMessage=msiexec.exe ErrorLevel was %ErrorLevel% for file %1 logfile at %logfile%
+  )
 goto :EOF
 
 


craigcomstock:
Tested with script at https://tracker.mender.io/browse/ENT-6846 (DT ticket, script using cf-remote) and was able to install package and see log generated on win12, win16, win19.

modules/packages/msiexec.bat
```diff
@@ -96,8 +96,12 @@ rem Install this file if it exists
     goto :EOF
   )
 
-  %MSIEXEC% /quiet /passive /qn /norestart /i %1
-  rem TODO options, error checking
+  REM TODO: ENT-6824 save this logfile based on msi filename
+  set logfile="\cfengine_package_install.log"
+  %MSIEXEC% /quiet /passive /qn /norestart /l*vx %logfile% /i %1
```
craigcomstock, line 101: I added both v and x options to the `/l` argument as they seemed "different" in the help/usage given by msiexec.exe dialog popup. In practice I didn't see much difference but figured "might as well".

modules/packages/msiexec.bat
```diff
@@ -96,8 +96,12 @@ rem Install this file if it exists
     goto :EOF
   )
 
-  %MSIEXEC% /quiet /passive /qn /norestart /i %1
-  rem TODO options, error checking
+  REM TODO: ENT-6824 save this logfile based on msi filename
+  set logfile="\cfengine_package_install.log"
```
nickanderson, line 100: Where on the file system is this log file?
I think we /must/ have something to prevent it from growing without bound. We have log file control in the MPF here: https://github.com/cfengine/masterfiles/blob/master/cfe_internal/core/log_rotation.cf

nickanderson (changes requested):
I think we need to make sure this file won't just grow without bound before we merge it.

modules/packages/msiexec.bat
craigcomstock: it is only one file. in the root directory. It cannot grow without bound I think.

nickanderson (approved):
OK, if it's not going to grow without bound.

modules/packages/msiexec.bat
craigcomstock: I might just include ENT-6824 (unique log file names) and some retention policy then in this PR?

craigcomstock:
cherry pick to 3.15.x and 3.12.x

PR #1921, opened 2021-02-26
ENT-6824 Changed msiexec package module install logs to be unique for each msi file

craigcomstock:
Logs are now placed in a unique "basename"_install.log file in\cfengine_package_logs.
That dir is also added in controls/def.cf to cfe_log_dirs to be
pruned by cfe_internal/core/log_rotation.cf policy every $(def.mpf_log_dir_retention) days.

Ticket: ENT-6824
Changelog: title

### DIFF
--- a/modules/packages/msiexec.bat
+++ b/modules/packages/msiexec.bat
@@ -96,11 +96,16 @@ rem Install this file if it exists
     goto :EOF
   )
 
-  REM TODO: ENT-6824 save this logfile based on msi filename
-  set logfile="\cfengine_package_install.log"
-  %MSIEXEC% /quiet /passive /qn /norestart /l*vx %logfile% /i %1
+  set log_dir="\cfengine_package_logs\"
+  if not exist %log_dir% (
+    mkdir %log_dir%
+  )
+  for /F "delims=" %%i in (%1) do @set basename="%%~ni"
+  REM %log_dir:"=% replaces quotes with nothing, otherwise you get two double-quotes which causes failures
+  set log_file="%log_dir:"=%%basename:"=%_install.log"
+  %MSIEXEC% /quiet /passive /qn /norestart /l*vx %log_file% /i %1
   if not errorlevel 0 (
-    echo ErrorMessage=msiexec.exe ErrorLevel was %ErrorLevel% for file %1 logfile at %logfile%
+    echo ErrorMessage=msiexec.exe ErrorLevel was %ErrorLevel% for file %1 log at %log_file%
   )
 goto :EOF
 
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -354,6 +354,10 @@ bundle common def
       "log_dir[outputs]" string => "$(sys.workdir)/outputs";
       "log_dir[reports]" string => "$(sys.workdir)/reports";
 
+    # TODO ENT-6845 - move package module logs to $(sys.workdir)/log/something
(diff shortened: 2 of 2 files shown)


modules/packages/msiexec.bat
```diff
@@ -96,8 +96,10 @@ rem Install this file if it exists
     goto :EOF
   )
 
-  %MSIEXEC% /quiet /passive /qn /norestart /i %1
-  rem TODO options, error checking
+  %MSIEXEC% /quiet /passive /qn /norestart /l*v %1_cfengine_install.log /i %1
+  if not errorlevel 0 (
+    echo ErrorMessage=msiexec ErrorLevel was %ErrorLevel% for file %1 logfile at %1_cfengine_install.log
+  )
```
craigcomstock, line 102: might be a good idea to put
```
) else (
  del %1_cfengine_install.log
)
```

To remove the log file in case of success.

nickanderson: What is %1? the package name? So a log file for each named package that is failing? Log file is over-written or appended to if it fails over multiple agent runs?

craigcomstock: overwritten

craigcomstock: So I have changed this slightly by putting the files in a dir so that log_rotation.cf will clean things up here.

Each unique msi file will have it's own log file and on repeated attempts it will be overwritten.

In most cases I would imagine this will be just fine as likely the errors will be the same over and over for a specific msi.

modules/packages/msiexec.bat
```diff
@@ -96,11 +96,15 @@ rem Install this file if it exists
...
 
-  REM TODO: ENT-6824 save this logfile based on msi filename
-  set logfile="\cfengine_package_install.log"
-  %MSIEXEC% /quiet /passive /qn /norestart /l*vx %logfile% /i %1
+  set log_dir="\cfengine_package_logs\"
+  if not exist %log_dir% (
+    mkdir %log_dir%
+  )
+  for /F "delims=" %%i in (%1) do @set basename="%%~ni"
+  set log_file="%log_dir:"=%%basename:"=%_install.log"
```
craigcomstock, line 104: replace quotes with nothing because otherwise I end up with two double quotes together which breaks the path (i think)

craigcomstock: might be worth an inline comment `REM`

nickanderson: +1

modules/packages/msiexec.bat
nickanderson: Ack

nickanderson (approved):
Looks good to me (not reviewing the implementation specifics of the bat stuff), but from a high level, good.

modules/packages/msiexec.bat
craigcomstock: > might be a good idea to put
> 
> ```
> ) else (
>   del %1_cfengine_install.log
> )
> ```
> 
> To remove the log file in case of success.

@nickanderson actually I'm wondering if just keeping the success log isn't a good idea as well, in case `msiexec.exe` returns success but the install actually doesn't quite work for some reason, then you have access to the detailed log!

nickanderson (approved):

craigcomstock:
picked back to 3.12.x and 3.15.x

PR #1922, opened 2021-03-02
CFE-3598/master: Added example illustrating definition of services_autorun class

nickanderson:
Ticket: CFE-3598
Changelog: None


vpodzime (approved):
ACK

PR #1927, opened 2021-03-08
ENT-6788/master: Made Enterprise CMDB data update after policy update

nickanderson:
This change makes sure that hosts update their local cache of CMDB data from an
Enterprise hub.

Ticket: ENT-6788
Changelog: Title

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -61,6 +61,10 @@ bundle agent cfe_internal_update_policy
       "Stock policy update"
         usebundle => cfe_internal_update_policy_cpv;
 
+    enterprise_edition::
+      "CMDB data update" -> { "ENT-6788" }
+        usebundle => cfe_internal_update_cmdb;
+
   reports:
 
     inform_mode|verbose_mode|DEBUG|DEBUG_cfe_internal_update_policy::
@@ -308,9 +312,33 @@ bundle agent cfe_internal_update_policy_cpv
         usebundle => cfe_internal_setup_python_symlink("$(python_symlink)");
 }
 
+bundle agent cfe_internal_update_cmdb
+# @brief Ensure local cache of CMDB data is up to date
+{
+  files:
+      "$(sys.workdir)/data/." create => "true";
+
+      "$(sys.workdir)/data/." -> { "ENT-6788" }
+        depth_search => u_recurse( inf ),
+        file_select => u_all,
+        copy_from => u_cmdb_data;
+}
 #########################################################
 # Self-contained bodies from the lib to avoid dependencies
 #########################################################
+
+body copy_from u_cmdb_data
+# @brief Sync CMDB data from policy server
+# Note: Not all hosts necessarily have CMDB data
+{
+        copy_backup => "false";
+        trustkey => "false";
+        compare => "digest";
+        source  => "hub_cmdb";
+        servers => { "$(sys.policy_hub)" };
+        purge => "true";
+        missing_ok => "true";
+}
 
 body perms u_m(p)
 # @brief Ensure file mode is `p`
@@ -342,6 +370,13 @@ body perms u_shared_lib_perms
 }
 
 #########################################################
+
+body file_select u_all
+# @brief Select all file system entries
+{
+        leaf_name => { ".*" };
+        file_result => "leaf_name";
+}
 
 body file_select u_cf3_files
 # @brief Select files starting with `cf-` (cfengine binaries)
--- a/controls/cf_serverd.cf
+++ b/controls/cf_serverd.cf
@@ -136,10 +136,16 @@ bundle server mpf_default_access_rules()
         if => isdir( "$(sys.workdir)/master_software_updates" ),
         admit => { @(def.acl) };
 
-      enterprise_edition.policy_server::
-        "collect_calls"
-          resource_type => "query",
-          admit_ips => { @(def.mpf_access_rules_collect_calls_admit_ips) };
+    enterprise_edition.policy_server::
+      "collect_calls"
+        resource_type => "query",
+        admit_ips => { @(def.mpf_access_rules_collect_calls_admit_ips) };
+
+      "$(sys.workdir)/cmdb/$(connection.key)/" -> { "ENT-6788" }
+        handle => "server_access_grant_access_cmdb",
+        comment => "Grant access to host specific CMDB data",
+        shortcut => "hub_mdr",
+        admit_keys => { $(connection.key) };
 
     !windows::
       "$(def.cf_runagent_shell)" -> { "ENT-6673" }


controls/cf_serverd.cf
```diff
@@ -141,6 +141,12 @@ bundle server mpf_default_access_rules()
           resource_type => "query",
           admit_ips => { @(def.mpf_access_rules_collect_calls_admit_ips) };
 
+      "$(sys.workdir)/cmdb/$(connection.key)/" -> { "ENT-6788" }
```
olehermanse, line 144: Indentation seems off, this is inside the enterprise and policy_server class guard.

controls/cf_serverd.cf
```diff
@@ -141,6 +141,12 @@ bundle server mpf_default_access_rules()
...
+        handle => "server_access_grant_access_cmdb",
+        comment => "Grant access to host specific CMDB data",
+        if => isdir( "/var/cfengine/cmdb/$(connection.key)" ),
```
olehermanse, line 147:
```suggestion
        if => isdir( "$(sys.workdir)/cmdb/$(connection.key)" ),
```

cfe_internal/update/update_policy.cf
```diff
@@ -308,10 +312,33 @@ bundle agent cfe_internal_update_policy_cpv
...
 
+body copy_from u_cmdb_data
+# @brief Sync CMDB data from policy server
+#
+# Note: Not all hosts necessarily have CMDB data
+{
+        copy_backup => "false";
+        trustkey => "false";
+        compare => "digest";
+        source  => "/var/cfengine/cmdb/$(sys.key_digest)/host_specific.json";
```
olehermanse, line 336: Do we have a variable for "sys.workdir on the policy server" ?

olehermanse (changes requested):
Great, just some smaller comments.

controls/cf_serverd.cf
nickanderson: Isn't that the only type of host that should be promising access to the data?

cfe_internal/update/update_policy.cf
nickanderson: No, not that I know of.

controls/cf_serverd.cf
nickanderson: Oh indentation

vpodzime (commented):
Squash the commits, please. Looks good to me otherwise. Nice and clear. :+1:

controls/cf_serverd.cf
nickanderson: The indentation was wrong for the promise above.

nickanderson: Fixed.

cfe_internal/update/update_policy.cf
nickanderson: I don't want to try and get it from `remotescalar()` either (Enterprise only, plus kind of painful to use).

I could however look to leverage a `shortcut` letting the server decide where this file actually is.

nickanderson:
Commits squashed. @olehermanse Do you want me to pursue shortcut for cmdb?

cfe_internal/update/update_policy.cf
```diff
@@ -308,10 +312,33 @@ bundle agent cfe_internal_update_policy_cpv
         usebundle => cfe_internal_setup_python_symlink("$(python_symlink)");
 }
 
+bundle agent cfe_internal_update_cmdb
+# @brief Ensure local cache of CMDB data is up to date
+{
+  files:
+      "$(sys.workdir)/data/." create => "true";
+
+      "$(sys.workdir)/data/host_specific.json" -> { "ENT-6788" }
```
nickanderson, line 321: @olehermanse, @vpodzime I still think this file name could be better.

`host_specific.json` doesn't really describe the data provenance.

In another place we discussed integrating other data sources, and that might end up named  something like servicenow.json. servicenow.json

olehermanse: > host_specific.json doesn't really describe the data provenance.

And?

This is the most straight forward and user friendly name I can think of. It's data. It's host specific. I think being easy to remember and easy to understand, especially for new users, is important.

All individual variables should have descriptive tags of where they came from, not necessarily encoded in the filename.

In the future we might have other files based on classes, like `ubuntu.json`, `windows.json`, `policy_server.json`, based on integrations, like `servicenow.json` or potentially other groupings / sources (class expressions? mission portal groups? regexes matching class names? Variable conditions? Maybe we want to add information about who set the data? And how they did it? (API, MP, integration, CLI?)

Trying to encode all of that in the file name is not a good idea IMO.

For me, as a user, seeing a directory with:

```
/var/cfengine/data
 - host_specific.json
 - ubuntu.json
 - webservers.json
```

Is awesome. I don't need to ask anyone about abbreviations, or anything else. It's self explanatory.

I then open the files, and see something like:

```json
{
  "vars": {
    "roles": {
      "value": ["cfengine_client", "webserver", "production"],
      "mission_portal_title": "Host Roles",
      "description": "The different roles a machine fulfills, including it's cfengine role, application purposes and environment(s)",
      "from": {
        "interface": "Mission Portal",
        "using": "GUI",
        "at": "2021-03-09T20:16:00Z",
        "by": "Nick Anderson"
      }
    }
  }
}
```

(That's just an example, and not the correct structure or keys, but I think it illustrates my point).

nickanderson:
@olehermanse, updated the PR, implemented shortcut and dir mirroring instead of just copying the one file.

controls/cf_serverd.cf
```diff
@@ -144,7 +144,7 @@ bundle server mpf_default_access_rules()
       "$(sys.workdir)/cmdb/$(connection.key)/" -> { "ENT-6788" }
         handle => "server_access_grant_access_cmdb",
         comment => "Grant access to host specific CMDB data",
-        if => isdir( "$(sys.workdir)/cmdb/$(connection.key)" ),
+        shortcut => "hub_mdr",
```
vpodzime, line 147: What kind of black magic is this?! Interesting... Anyway, I'd stick to *cmdb*, *mdr* is just confusing here.

cfe_internal/update/update_policy.cf
```diff
@@ -308,10 +312,34 @@ bundle agent cfe_internal_update_policy_cpv
...
 #########################################################
 
+body copy_from u_cmdb_data
+# @brief Sync CMDB data from policy server
+# Note: Not all hosts necessarily have CMDB data
+{
+        copy_backup => "false";
+        trustkey => "false";
+        compare => "digest";
+        source  => "hub_mdr";
```
olehermanse, line 337:
```suggestion
        source  => "hub_cmdb";
```

controls/cf_serverd.cf
```diff
@@ -136,10 +136,16 @@ bundle server mpf_default_access_rules()
...
-          admit_ips => { @(def.mpf_access_rules_collect_calls_admit_ips) };
+    enterprise_edition.policy_server::
+      "collect_calls"
+        resource_type => "query",
+        admit_ips => { @(def.mpf_access_rules_collect_calls_admit_ips) };
+
+      "$(sys.workdir)/cmdb/$(connection.key)/" -> { "ENT-6788" }
+        handle => "server_access_grant_access_cmdb",
+        comment => "Grant access to host specific CMDB data",
+        shortcut => "hub_mdr",
```
olehermanse, line 147:
```suggestion
        shortcut => "hub_cmdb",
```

controls/cf_serverd.cf
olehermanse: Woah, the shortcut expands individual `$(connection.key)` variables? Cool! 🚀

olehermanse (approved):

controls/cf_serverd.cf
nickanderson: Yeah, shortcuts are cool stuff. It's how we can let the remote host decide what file we get.

PR #1945, opened 2021-03-24
Allows overriding modules and plugins directories path on policy server

gpiero:
Introduces mpf_update_modules_master_location and
mpf_update_plugins_master_location variables (configurable via augments
file) for configuring where on the policy server the modules and plugins
directories reside. Do not rely on the client's operating system, as the
policy server can be installed on a different one.

This patch removes the
cfe_internal_update_policy_vars_modules_dir_windows and
cfe_internal_update_policy_vars_plugins_dir_windows handles, that
didn't seem to be referred anywhere.

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -171,27 +171,15 @@ bundle agent cfe_internal_update_policy_cpv
         comment => "The path to request updates from the policy server.",
         handle => "cfe_internal_update_policy_vars_master_location";
 
-    windows::
+      "modules_dir_source" -> { "CFE-3318" }
+        string => "$(update_def.mpf_update_modules_master_location)",
+        comment => "Directory containing CFEngine modules on the policy server.",
+        handle => "cfe_internal_update_policy_vars_modules_dir";
 
-      "modules_dir_source"        string => "/var/cfengine/masterfiles/modules",
-      comment => "Directory containing CFEngine modules",
-      handle => "cfe_internal_update_policy_vars_modules_dir_windows";
-
-      "plugins_dir_source"        string => "/var/cfengine/plugins",
-      comment => "Directory containing CFEngine plugins",
-      handle => "cfe_internal_update_policy_vars_plugins_dir_windows";
-
-    !windows::
-
-      "modules_dir_source"        string => translatepath("$(master_location)/modules"),
-      comment => "Directory containing CFEngine modules",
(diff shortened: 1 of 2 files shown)


gpiero:
This is related to https://tracker.mender.io/browse/CFE-3318
Sorry for the title of the original bug report, it was  mangled by some paste malfunction but I cannot edit it.

nickanderson (changes requested):
Commented on related ticket in Jira, but for convenience here as well:

On Removing `modules_dir_source` and `plugins_dir_source` from `cfe_internal/upste/update_policy.cf`:

The current hard coded paths for windows are not ideal, but with this change `update_def.mpf_update_modules_master_location` will be wrong by default for windows agents as *@prefix@* will be set to `C:\Program Files\Cfengine` which won't be the correct path on the hub.

On adding `mpf_update_modules_master_location` and `mpf_update_plugins_master_location` to `update_def.cf`:

I think this looks OK, but I think you need to have a block for windows where the default is the hard coded expected paths like it is today. Perhaps a better option would be to introduce *shortcuts* for these so that the client could request `modules` or `plugins` and the server can decide what the path for that should be. In fact, I see that there is already a shortcut for `modules` in the MPF, but not one for `plugins`.

    "$(def.dir_modules)/"
    handle => "server_access_grant_access_modules",
    shortcut => "modules",
    comment => "Grant access to modules directory",
    if => isdir( "$(def.dir_modules)/" ),
    admit => { @(def.acl) };
    
    "$(def.dir_plugins)/"
    handle => "server_access_grant_access_plugins",
    comment => "Grant access to plugins directory",
    if => isdir( "$(def.dir_plugins)/" ),
    admit => { @(def.acl) };

Looks like the shortcut for *modules* has been there since 3.10.0, so we could just use the shortcut by default for that, but plugins would need some juggling. Then again, *plugins* is unused as far as I know, so perhaps we can just kill that directory. It looks like the plugins dir was introduced in 2013 by @tzz .

@olehermanse , thoughts about killing this dir and policy to copy it in master for 3.18?

nickanderson:
> This is related to https://tracker.mender.io/browse/CFE-3318
> Sorry for the title of the original bug report, it was mangled by some paste malfunction but I cannot edit it.

No worries, I cleaned it up. Thank you very much for creating the associated ticket.

nickanderson:
Hey @gpiero, I was working on this but got caught up, posted update in https://tracker.mender.io/browse/CFE-3318

PR #1951, opened 2021-03-29
CFE-3615: add /usr/bin/yum to paths.cf for aix

flynn1973:
yum has become an integral part of aix and it is  commonly used, as ibm has put  ehanced effort into the aix toolbox for linux applications (gnu freeware rpm packages for aix). so yum should be considered in cfengine as a valid package manager. this PR is part of ongoing efforts to make cfengine more aix aware.


cf-bottom:

nickanderson (approved):

PR #1987, opened 2021-05-12
Fixed new OS inventory string based on new vars

olehermanse:
Changed sys.os_name_major -> sys.os_version_major

One is correct, the other one isn't.


nickanderson:
:+1:

PR #1995, opened 2021-06-03
ENT-7233: Added cleanup of database and status semaphore when federation target_state is off

nickanderson:


craigcomstock (approved):

craigcomstock:
Will give these changes a test before we merge.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -314,7 +319,11 @@ bundle agent clean_when_off
       "has_cftransport_fcontext" expression => returnszero("$(paths.semanage) fcontext -l | grep $(home)", "useshell");
 
   commands:
-    # _stdlib_path_exists_<command> and paths.<command> are defined is masterfiles/lib/paths.cf
+      # Oh, the humanity! Where for art thou databases: promises for psql!
```
craigcomstock, line 322: The problem here is that if you put a feeder to target_state off you will get errors as no changes are made to the database, so need to guard with is_superhub here.

craigcomstock (changes requested):
Need to guard with `am_superhub` when cleaning up database. Will result in errors on a feeder.

cfe_internal/enterprise/federation/federation.cf
nickanderson: as-you-wish

nickanderson:
> Need to guard with `am_superhub` when cleaning up database. Will result in errors on a feeder.

craigcomstock (approved):
🎉

cfe_internal/enterprise/federation/federation.cf
craigcomstock: My comment wasn't right. Any hub 3.15.x and forward has the two cfsettings database tables. The issue here really is only do this if a hub and the version is 3.15 or greater. Currently https://tracker.mender.io/browse/ENT-8294 is a problem with triple upgrade CI test because an upgrade from 3.12.7 to 3.15.5 will break because 3.12.7 hubs don't have these tables in cfsettings.

PR #2009, opened 2021-07-07
Fix monitoring_include sample values

craigcomstock:
There are no "cpu_.*" measurements: https://docs.cfengine.com/docs/3.18/reference-special-variables-mon.html


nickanderson:
Superseded by https://github.com/cfengine/masterfiles/pull/2011

PR #2011, opened 2021-07-09
ENT-7417: Fixed example showing configuration for collecting CPU measurements (master)

nickanderson:
Also added History for when configuration was introduced.

Ticket: ENT-7417
Changelog: None


olehermanse:
@nickanderson was this cherry-picked? Should it be?

nickanderson:
> @nickanderson was this cherry-picked? Should it be?

Indeed it was.
image

PR #2017, opened 2021-07-14
ENT-5125 Synchronized systemd units with core/misc/systemd (3.18.x)

craigcomstock:
merge together
https://github.com/cfengine/core/pull/4737
https://github.com/cfengine/masterfiles/pull/2017


craigcomstock:
@cf-bottom jenkins

PR #2037, opened 2021-07-26
Added hub specific policy to maintain static directory

aleksandrychev:


nickanderson (approved):

cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -358,8 +358,10 @@ bundle agent cfe_internal_truncate_events
 bundle agent cfe_internal_purge_scheduled_reports_older_than_days(days)
 # @brief Clean up scheduled reports older than `days` days old
 {
+  vars:
+    "tmp_directories_list" slist => {"$(cfe_internal_hub_vars.public_docroot)/tmp/.", "$(cfe_internal_hub_vars.docroot)/static/."};
   files:
-    "$(cfe_internal_hub_vars.public_docroot)/tmp/."
+    "$(tmp_directories_list)"
```
nickanderson, line 364: I think that we need to add another constraint to this promise:

This _static_ directory will not exist when users are in the upgrade process while they are running new policy with old binaries. To avoid errors related to being unable to change into the directory and search for files we should restrict this promise to when the directory exists by adding:

```
if => isdir( $(tmp_directories_list) )
```

nickanderson (changes requested):
I spotted one thing that I think needs changed.

cfe_internal/enterprise/CFE_hub_specific.cf
aleksandrychev: thanks. added this condition.

nickanderson (approved):
Looks OK to me.

PR #2040, opened 2021-07-28
ENT-7474 Use Federated Reporting postgresql advisory lock

craigcomstock:


nickanderson (approved):
Wow, the advisory lock handles a lot.

this policy change looks ok to me

vpodzime:
> this policy change looks ok to me

Yes, especially thinking the policy would have had to be made even more complicated to fix issues with `cf-runalerts` instead.

craigcomstock:
ready to go, pending @aleksandrychev  approval of https://github.com/cfengine/nova/pull/1785

PR #2042, opened 2021-07-29
CFE-3721 Skip ownership of package modules on termux

craigcomstock:
Ticket: CFE-3721
Changelog: title

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -453,7 +453,7 @@ body perms u_mo(p,o)
 # @param p Desired file owner (username or uid)
 {
       mode   => "$(p)";
-      !windows::
+      !(windows|termux)::
         owners => {"$(o)"};
 }
 


cfe_internal/update/update_policy.cf
```diff
@@ -453,7 +453,7 @@ body perms u_mo(p,o)
 # @param p Desired file owner (username or uid)
 {
       mode   => "$(p)";
-      !windows::
+      !windows.!termux::
         owners => {"$(o)"};
```
nickanderson, line 457: Not windows AND not Termux don't we want to skip it if windows or termux is defined?

How about `!(windows|termux)::` instead? I think it reads more clearly

craigcomstock: sure.

nickanderson (approved):

craigcomstock:
needs 3.18.x cherry-pick

PR #2043, opened 2021-07-30
CFE-3724: apt_get package module fails to fully install packages from local files

nickanderson:
Prior to this change packages promises using the apt_get package module would
use dpkg to handle the case when the path to a package was the promiser. dpkg
does not resolve dependencies, so the result could be a half installed package,
or a package installed with unmet dependencies leaving it in a nonfunctional
state.

This change switches to using apt-get to install the package from a local file,
with apt-get, dependencies are automatically resolved and would be installed
from repositories if available.

Ticket: CFE-3724
Changelog: Title

### DIFF
--- a/modules/packages/vendored/apt_get.mustache
+++ b/modules/packages/vendored/apt_get.mustache
@@ -349,7 +349,7 @@ def remove():
     return ret
 
 def file_install():
-    cmd_line = [dpkg_cmd] + dpkg_options + ["-i"]
+    cmd_line = [apt_get_cmd] + apt_get_options + ["install"]
 
     found = False
     for line in sys.stdin:


cf-bottom:

vpodzime (approved):
ACK. Did it use to be the case that *apt-get* could not install files? I'm still confused why *dpkg* was used here in the first place.

nickanderson:
> ACK. Did it use to be the case that _apt-get_ could not install files? I'm still confused why _dpkg_ was used here in the first place.

I think that there are 2 reasons why `dpkg` is used here.
1. Presumably it's more lightweight than apt-get, apt, or aptitude, it doesn't hit the network
2. I think apt-get didn't used to be able to install from local files, I seem to recall when that was an aptitude only feature.

And, did you know that our yum package_module uses `rpm`?
https://github.com/cfengine/masterfiles/blob/master/modules/packages/vendored/yum.mustache#L12

PR #2068, opened 2021-08-18
ENT-2589: Switched from using package_method generic to default package_module for windows software inventory

nickanderson:
Now that the windows package module is used by default we can stop using the
legacy packages promise to generate software inventory. Furthermore, this change
should prevent log spam generated by MsiInstaller in the default policy which
was generated for each piece of software found as a result of the use of the
Win32_Product class.

Ticket: ENT-2589
Changelog: Title


cf-bottom:

craigcomstock (approved):

PR #2110, opened 2021-10-13
CFE-3790: Added lsmod to well known paths

nickanderson:
Ticket: CFE-3790
Changelog: Title

### DIFF
--- a/lib/paths.cf
+++ b/lib/paths.cf
@@ -139,6 +139,7 @@ bundle common paths
 
     linux::
       "path[lsattr]"        string => "/usr/bin/lsattr";
+      "path[lsmod]"         string => "/sbin/lsmod";
       "path[tar]"           string => "/bin/tar";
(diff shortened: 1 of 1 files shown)


lib/paths.cf
```diff
@@ -139,6 +139,7 @@ bundle common paths
 
     linux::
       "path[lsattr]"        string => "/usr/bin/lsattr";
+      "path[lsmod]"         string => "/sbin/lsmod";
```
craigcomstock, line 142: sure. on ubu 20 I get:

```
$ ls -l /sbin/lsmod
lrwxrwxrwx 1 root root 9 Aug 19 11:51 /sbin/lsmod -> /bin/kmod
$ ls -l /usr/sbin/lsmod
lrwxrwxrwx 1 root root 9 Aug 19 11:51 /usr/sbin/lsmod -> /bin/kmod
```

PR #2111, opened 2021-10-15
ENT-7714: Physical Memory (MB) inventory now handles dmidecode MB or GB units

nickanderson:
found.

Ticket: ENT-7714
Changelog: Title

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -983,7 +983,7 @@ bundle agent cfe_autorun_inventory_dmidecode
 
     have_dmidecode::
 
-      "$(decoder) -t 17 | $(paths.awk) '/Size.*MB/ {s+=$2} END {print s}' > '$(sys.statedir)/inventory-$(this.bundle)-total-physical-memory-MB.txt'" -> { "CFE-2896" }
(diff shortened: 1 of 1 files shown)


inventory/any.cf
```diff
@@ -983,7 +983,7 @@ bundle agent cfe_autorun_inventory_dmidecode
 
     have_dmidecode::
 
-      "$(decoder) -t 17 | $(paths.awk) '/Size.*MB/ {s+=$2} END {print s}' > '$(sys.statedir)/inventory-$(this.bundle)-total-physical-memory-MB.txt'" -> { "CFE-2896" }
+      "$(decoder) -t 17 | $(paths.awk) '/^\t*Size:.*MB/ {a+=$2} /^\t*Size:.*GB/ {b+=$2*1024} END {print a+b}' > '$(sys.statedir)/inventory-$(this.bundle)-total-physical-memory-MB.txt'" -> { "CFE-2896", "ENT-7714" }
```
craigcomstock, line 986: Is the `*` needed after `\t`? I suppose it doesn't hurt much but seems extra.

PR #2112, opened 2021-10-15
ENT-7714: Physical Memory (MB) inventory now handles dmidecode MB or GB units (3.18)

nickanderson:
found.

Ticket: ENT-7714
Changelog: Title
(cherry picked from commit 71dbe26fb66487bc8bf25e607c6787acf52963de)


cf-bottom:

PR #2113, opened 2021-10-15
ENT-7714: Physical Memory (MB) inventory now handles dmidecode MB or GB units (3.15)

nickanderson:
found.


cf-bottom:

PR #2133, opened 2021-10-28
CFE-3441: Test using attribute `with` with custom promise types

larsewi:
Dependent on: https://github.com/cfengine/core/pull/4842

Ticket: CFE-3441
Changelog: None

### DIFF
--- a/tests/acceptance/30_custom_promise_types/21_with_attribute.cf
+++ b/tests/acceptance/30_custom_promise_types/21_with_attribute.cf
@@ -1,0 +1,91 @@
+######################################################
+#
+#  Test that attribute with works with custom promise types
+#
+#####################################################
+body common control
+{
+    inputs => { "../default.cf.sub" };
+    bundlesequence  => { default("$(this.promise_filename)") };
+    version => "1.0";
+}
+
+#######################################################
+
+bundle common version_check
+{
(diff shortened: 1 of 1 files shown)


tests/acceptance/30_custom_promise_types/21_with_attribute.cf
```diff
@@ -0,0 +1,91 @@
...
+      string => "Test that depends_on works with custom promise types";
+
+  vars:
+    "letters"
+      slist => { "A", "B", "C", "D", "E", "F" };
+
+  append:
+    "$(G.testfile)"
+      string => "$(with)",
+      with => string_downcase("$(letters)");
```
nickanderson, line 66: I guess this works, you know that you are effectively making N promises here, where N is the number of Elements in `letters`?
This is just a test, so i thinkit's fine. Generally though, you wouldn't want to edit the same file multiple times if you can avoid it, instead you would want to perhpahs join the letters and promise it once (in normal real life).

nickanderson: and its very reasonable to test this behavior because that iteration aspect is important.

nickanderson:
Picked: 
image

PR #2139, opened 2021-11-11
Changelog for 3.15.x

Lex-2008:


nickanderson (changes requested):
I made one direct change, should be squashed

olehermanse (approved):

PR #2178, opened 2021-12-08
Added prelink to paths.cf

larsewi:
Ticket: None
Changelog: None


nickanderson (approved):

nickanderson (changes requested):
Change looks good to me, but in your commit message, I think you should have a changelog for this. `Added prelink to paths.cf`.

PR #2179, opened 2021-12-08
ENT-8079: Plumbing in support of distributing compliance reports from cfbs modules

nickanderson:
Merge together:
- https://github.com/cfengine/core/pull/4882
- https://github.com/cfengine/documentation/pull/2597

These changes to the MPF add .no-distrib as a directory that should not be copied as part of policy update so that cfbs modules could populate this directory with assets that should not be implicitly distributed but are not necessarily explicitly prohibited from distribution.

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -14,6 +14,7 @@ various aspects of the system.
 * `services/main.cf` - Contains an empty bundle agent main where custom policy
   can be integrated.
 * `services/autorun/` - Automatically included policy files.
+* `.no-distrib/` - A directory that is excluded from policy updates from remote agents.
 
 The MPF is continually updated. You can track its development
 on [github](https://github.com/cfengine/masterfiles/).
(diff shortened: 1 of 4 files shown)


basvandervlies:
@nickanderson thanks

MPF.md
```diff
@@ -14,6 +14,7 @@ various aspects of the system.
 * `services/main.cf` - Contains an empty bundle agent main where custom policy
   can be integrated.
 * `services/autorun/` - Automatically included policy files.
+* `.no-distrib/` - A directory that is excluded from policy updates.
```
craigcomstock, line 17: to be more clear here possibly, "excluded from policy updates from agents." maybe? Just to clarify "policy updates" means an agent running cf-agent -KIf update.cf.

cfe_internal/update/update_policy.cf
```diff
@@ -247,7 +247,7 @@ bundle agent cfe_internal_update_policy_cpv
       comment => "Copy policy updates from master source on policy server if a new validation was acquired",
       handle => "cfe_internal_update_policy_files_inputs_dir",
       copy_from => u_rcp("$(master_location)", @(update_def.policy_servers)),
-      depth_search => u_recurse("inf"),
+      depth_search => u_infinate_client_policy,
```
craigcomstock, line 250: s/infinate/infinite/ I looked up infinate and it is certainly a word, but looks like a typo.

MPF.md
nickanderson: ack

craigcomstock (approved):

PR #2180, opened 2021-12-09
Added prelink to paths.cf

larsewi:
Ticket: None
Changelog: Title
Signed-off-by: larsewi <lars.erik.wik@northern.tech>


craigcomstock (commented):
Seems to make sense to only do on linux? ELF only?

larsewi:
> Seems to make sense to only do on linux? ELF only?

I believe Linux is the correct scope for this

> On Linux, prelinking is accomplished via the prelink program, a free program written by Jakub Jelínek of Red Hat for ELF binaries. 
 - https://en.wikipedia.org/wiki/Prelink

craigcomstock (approved):

PR #2186, opened 2021-12-15
Auto-detect current release and add it to policy

Lex-2008:
For self-update needs.

Current release number is taken from env variables, or detected from git
tags. If everything else fails, it defaults to '1' as before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ stamp-h1
 /ylwrap
 /test-driver
 CFVERSION
+CFRELEASE
 
 # Misc.
 /cf_promises_*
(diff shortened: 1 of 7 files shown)


craigcomstock:
@Lex-2008 did you forget to include the new file `misc/determine-release.sh`?

Lex-2008:
yes, I did. Thanks! Force-pushed with new file

craigcomstock (commented):
I don't know autotools too well, the other bits look OK :)

vpodzime (commented):
The CFRELEASE file should be added to `.gitignore`. Looks good to me otherwise.

vpodzime (approved):
ACK

nickanderson (approved):
Policy changes seem ok

PR #2187, opened 2021-12-15
ENT-8274: Added support for Amazon Linux in standalone self upgrade

nickanderson:
Prior to this change Amazon Linux hosts (based on redhat|fedora) would not
attempt to upgrade themselves as they were not explicitly supported. With this
change hosts that have the amazon_linux class defined will use the same
methodology as redhat and centos hosts with respect to self upgrade.


cf-bottom:

craigcomstock (approved):

PR #2192, opened 2021-12-17
ENT-8066: Fixed windows unattended self upgrade on Windows 2008

nickanderson:
On some versions of windows escaping the space in the path to the script that
performs the upgrade from the script that schedules the upgrade. This change
uses "Here-strings" so that we can have matroska quotes.

This change also runs the script to schedule the upgrade with powershell
-NoProfile -ExecutionPolicy bypass -F path to schedule script because it was
observed to be necessary on some (but not all?) hosts. Probably related to some
other system setting that I just dunno (or care) about. Possibly, it's only
required for use when running on the command line interactively and is not
necessary to do from policy, but it works from policy to do it this way, so it
makes manual testing and investigation much easier.

Ticket: ENT-8066
Changelog: Title

### DIFF
--- a/cfe_internal/update/windows_unattended_upgrade.cf
+++ b/cfe_internal/update/windows_unattended_upgrade.cf
@@ -16,8 +16,10 @@ bundle agent windows_unattended_upgrade
         string => "$(const.dirsep)schedule_unattended_upgrade.ps1",
         comment => "sys.bindir has issues when used as execution path"; # TODO examine and fix/document specific issue(s)
 
+      "_upgrade_script_name" string => "unattended-upgrade.bat";
+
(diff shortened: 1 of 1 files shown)


nickanderson:
@larsewi, @craigcomstock and I hacked at this for a while and Maybe we got it working.

Can you please test this on the various versions of windows?

larsewi (approved):
Tested:
- [x] Windows 2008
- [x] Windows 2012
- [x] Windows 2016
- [x] Windows 2019

nickanderson:
https://github.com/cfengine/masterfiles/pull/2198
https://github.com/cfengine/masterfiles/pull/2199

PR #2198, opened 2021-12-20
ENT-8066: Fixed windows unattended self upgrade on Windows 2008 (3.18)

nickanderson:
On some versions of windows escaping the space in the path to the script that
performs the upgrade from the script that schedules the upgrade. This change
uses "Here-strings" so that we can have matroska quotes.

This change also runs the script to schedule the upgrade with powershell
-NoProfile -ExecutionPolicy bypass -F path to schedule script because it was
observed to be necessary on some (but not all?) hosts. Probably related to some
other system setting that I just dunno (or care) about. Possibly, it's only
required for use when running on the command line interactively and is not
necessary to do from policy, but it works from policy to do it this way, so it
makes manual testing and investigation much easier.

Ticket: ENT-8066
Changelog: Title
(cherry picked from commit cb3822534b59189ad8812fb492f7f60f1f3c3dfc)

### DIFF
--- a/cfe_internal/update/windows_unattended_upgrade.cf
+++ b/cfe_internal/update/windows_unattended_upgrade.cf
@@ -16,8 +16,10 @@ bundle agent windows_unattended_upgrade
         string => "$(const.dirsep)schedule_unattended_upgrade.ps1",
         comment => "sys.bindir has issues when used as execution path"; # TODO examine and fix/document specific issue(s)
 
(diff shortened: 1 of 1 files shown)


nickanderson:
@larsewi

Can you test this one on 3.18 please?

larsewi (approved):
- [x] Tested 3.18.x

PR #2200, opened 2021-12-22
Replace cfbs/cleanup.sh with prepare.sh

vpodzime:
Requiring autotools for just converting a checked-out version of
MPF into a directory that can actually be used as masterfiles is
unfortunate. However, autotools deal with many things, especially
on non-standard platforms, and provide many commonly used
commands and features.

To support the simple use case of just using the checked-out
version of MPF as masterfiles this commit introduces a script
that does the necessary bits:

  1. determine version and release
  2. expand values in templates
  3. remove unnecessary files

And since such use case is not specific to cfbs, let's just have
the script in the top-level directory.

Ticket: CFE-3872
Changelog: None


olehermanse (approved):
Looks good, excited to try this :)

PR #2214, opened 2022-01-13
Added TODO notes for redactions after 3.15 is EOL

nickanderson:


cfe_internal/enterprise/CFE_hub_specific.cf
```diff
@@ -642,6 +642,7 @@ bundle agent cfe_internal_refresh_hosts_view
 
 bundle agent cfe_internal_clear_last_seen_hosts_logs
 # @brief Clear last seen hosts log every 5 minutes
+# TODO Redact this bundle when 3.15 is EOL.
```
craigcomstock, line 645: a reason why and a reference to a ticket might be nice.

nickanderson: Yeah, it would have been smart of me to include that in this note since now I have to figure it out again.

nickanderson: Why should we redact `bundle agent cfe_internal_clear_last-seen_hosts_logs` when 3.15 is EOL?

Because:

The commands in the policy are scoped to run only on 
active enterprise hubs with `(cfengine_3_12|cfengine_3_13|cfengine_3_14)`.  3.15.0 was released while 3.12.x was still supported and this bundle should remain until 3.15.0 becomes EOL so that someone running 3.12.x could upgrade to 3.15.x policy while still running 3.12.x binaries and this bundle will continue to function as expected.

craigcomstock (approved):

PR #2225, opened 2022-01-28
Fixed Travis failure 'E: Unable to locate package python3.6'

larsewi:


vpodzime:
> ** Travis CI - Pull Request ** Failing after 4m — Build Failed

My bad, actually.

vpodzime:
> My bad, actually.

See https://github.com/cfengine/masterfiles/pull/2228

vpodzime:
```
==35050== Command: /var/cfengine/bin/cf-agent -B RX packets 0 bytes 0 (0.0 B)

==35050== 

   error: Could not resolve hostname 'RX', unable to bootstrap
```
looks like some place determining the hostname now gets unexpected data. We can just use `hostname`, I guess?

vpodzime (approved):
Nice! Good job!

travis-scripts/valgrind.sh
```diff
@@ -144,11 +144,10 @@ function check_masterfiles_and_inputs {
...
 VG_OPTS="--leak-check=full --track-origins=yes --error-exitcode=1"
-BOOTSTRAP_IP="$(ifconfig | grep -A1 Ethernet | sed '2!d;s/.*addr:\([0-9.]*\).*/\1/')"
 
 valgrind $VG_OPTS /var/cfengine/bin/cf-key 2>&1 | tee cf-key.txt
 check_output cf-key.txt
-valgrind $VG_OPTS /var/cfengine/bin/cf-agent -B $BOOTSTRAP_IP 2>&1 | tee bootstrap.txt
+valgrind $VG_OPTS /var/cfengine/bin/cf-agent -B $HOSTNAME 2>&1 | tee bootstrap.txt
 check_output bootstrap.txt
 
 echo "127.0.0.1" > /var/cfengine/policy_server.dat
```
olehermanse, line 153:
```suggestion
valgrind $VG_OPTS /var/cfengine/bin/cf-agent -B 127.0.0.1 2>&1 | tee bootstrap.txt
check_output bootstrap.txt
```

olehermanse: @larsewi we can bootstrap to localhost now, could you make this change in a followup PR?

olehermanse (approved):

travis-scripts/valgrind.sh
vpodzime: Hmm, wouldn't that bypass much of the networking code?

larsewi: What is the advantage of using localhost?

vpodzime: You don't have to bother about the hostname of the host. But if `$(HOSTNAME)` just works, we can just use that, IMHO.

olehermanse: 127.0.0.1 "always works" and makes the default ACL not allow connections from other machines.

We are already doing this by overwriting `policy_server.dat`, just bootstrapping to 127.0.0.1 is one step less.

vpodzime: Sure, my only concern is if bootstrapping to `127.0.0.1` goes through the same code paths as normal bootstrap to a (local) hostname. The valgrind test should have the best possible coverage, after all.

larsewi:
Cherry picks:
 * https://github.com/cfengine/masterfiles/pull/2230
 * https://github.com/cfengine/masterfiles/pull/2231

PR #2228, opened 2022-01-28
Make sure the 'test' bundle is defined in the 22_single_evaluation test

vpodzime:
It is supposed to be defined in the test on all versions, even on
those where the test is skipped.


cf-bottom:

larsewi (commented):
Maybe add a constraint to only read `G.testfile` if file exists?

vpodzime:
I don't understand why the test is not skipped on versions before 3.20.0

larsewi (approved):
Looks good now 👍

vpodzime:
:relieved: Finally... I just squashed the commits.

PR #2242, opened 2022-02-10
CFE-3920: Cleaned up policy related to versions prior to 3.12

nickanderson:
Now that 3.12.x is not under standard support, we can remove anything that was
not used in 3.12. Things used in 3.12 need to remain to cover the case where
someone is upgrading to 3.15.x because 3.12.x was supported when 3.15.0 was
released.

This change redacts policy related to cf-consumer and cf-redis-server.

Ticket: CFE-3920
Changelog: Title

### DIFF
--- a/lib/README.md
+++ b/lib/README.md
@@ -3,9 +3,6 @@ referred to as the Common Open Promise Body Library.
 
 Layout of this `lib` directory:
 
-* `3.6/*.cf`: Modularized version of the CFEngine Standard Library for
-  3.6.x.
-
 * `lib/*.cf`: Re-unified library compatible with 3.7+
   - *autorun.cf*: This file contains bundles that support automatic activation
     of bundles based on tags as well as automatically adding policy files found
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -204,15 +204,6 @@ bundle agent cfe_internal_update_policy_cpv
         comment => "Symlink to Python we found (if any)",
         handle => "cfe_internal_update_policy_python_symlink";
 
-    cfredis_in_enterprise::
-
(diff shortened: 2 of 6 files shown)


olehermanse (approved):

nickanderson:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/7571/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/7571/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-7571/

lib/README.md
```diff
@@ -3,9 +3,6 @@ referred to as the Common Open Promise Body Library.
 
 Layout of this `lib` directory:
 
-* `3.6/*.cf`: Modularized version of the CFEngine Standard Library for
```
craigcomstock, line 6: how is this related to 3.12? Did we make a switch from 3.6/*.cf to lib/*.cf at that point?

craigcomstock: or was 3.6 supported when 3.12 was released?

craigcomstock (commented):
looks good! one question about 3.6/*.cf

lib/README.md
nickanderson: It's not related to 3.12 at all, it's just another cleanup that I found while I was looking around. This is back from before we had macros to guard against future syntax and because of that we split the stdlib into version specific trees so that there were no new functions loaded by old agents that didn't understand the macros to avoid parsing functions and language features they didn't know about. 

It seemed too small to have in a separate PR, that's why its in this PR in a separate commit and that commit doesn't have a Ticket keyword :)

PR #2246, opened 2022-02-15
ENT-8393 Fixed inclusion of distributed cleanup python files during install

craigcomstock:
configure.ac was missing entry

controls/update.cf already includes *.py so OK there.

Ticket: ENT-8393
Changelog: title

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -225,7 +225,7 @@ do
 done
 for i in templates cfe_internal modules/packages/vendored
 do
-    for j in `find "$srcdir/$i" -name '*.mustache' -o -name '*.sh' -o -name '*.awk' -o -name '*.sed' -o -name '*.ps1'`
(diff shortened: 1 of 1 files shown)


nickanderson:
Do we need this? I found the files in the tarball so something is getting them into place.

craigcomstock:
> Do we need this? I found the files in the tarball so something is getting them into place.

Yes. I checked the latest nightly masterfiles tarball and these scripts are missing. I'll test make tar-ball locally and check.

craigcomstock:
`make tar-package` in master does indeed miss these python files in templates dir. So this change is needed.

nickanderson (approved):
Alrighty then.

Wonder how I got the files in my repo as untracked files then.

craigcomstock:
only 3.18 cherry pick was needed: https://github.com/cfengine/masterfiles/pull/2247

PR #2252, opened 2022-02-23
CFE-1234: Added update_line based

nickanderson:
Because all the reasons that are so good

balhablaha

Ticket: CFE-1234
Changelog: Title


cf-bottom:

nickanderson:
That was just a demonstration for someone.

PR #2255, opened 2022-02-28
CFE-3993: Removed policy deprecated by sys.os_release

nickanderson:
CFEngine 3.11.0 brought classification based on /etc/os-release to core.
This change removes policy pre-dating core support for classification based on /etc/os-release.

### DIFF
--- a/inventory/debian.cf
+++ b/inventory/debian.cf
@@ -27,7 +27,7 @@ bundle common inventory_debian
     any::
       "debian_derived_evaluated"
       scope => "bundle",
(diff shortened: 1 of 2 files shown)


nickanderson:
These need to be squashed together, but left as individual commits for easier review.

craigcomstock (approved):

vpodzime (approved):
Good riddance. And I think the commits can stay as they are, actually. The squashed commit would probably end up with less info in the commit message.

PR #2256, opened 2022-02-28
ENT-8424: Stopped lowercasing software inventory on Windows

nickanderson:
I can not find any legitimate reason that we are lowercasing software names on
Windows.

Ticket: ENT-8424
Changelog: Title


Lex-2008 (approved):

nickanderson:
@Lex-2008 , @olehermanse  do you think we should pick this back to 3.18, 3.15?

vpodzime:
> @Lex-2008 , @olehermanse do you think we should pick this back to 3.18, 3.15?

Don't see a reason why not. At worst it can break some users' queries, but that should be easy to fix and the new behavior is better.

nickanderson:
> > @Lex-2008 , @olehermanse do you think we should pick this back to 3.18, 3.15?
> 
> Don't see a reason why not. At worst it can break some users' queries, but that should be easy to fix and the new behavior is better.

Thanks!

nickanderson:
Cherry picked to 3.18 and 3.15
https://github.com/cfengine/masterfiles/pull/2259
https://github.com/cfengine/masterfiles/pull/2258

PR #2262, opened 2022-03-07
CFE-951: Adjusted rxdirs to have correct explicit default (3.18)

nickanderson:
Beginning with CFEngine 3.20 rxdirs default has changed from true to false.

This change explicitly sets the value to align with binary defaults. When
running binary versions prior to 3.20 rxdirs is enabled. For versions 3.20 and
greater rxdirs is disabled by default.

Starting with CFEngine 3.18.2, if a perms body sets mode without
specifying rxdirs a warning will be emitted. This change prevents those warnings
for perms bodies that are part of the MPF.

Ticket: CFE-951
Changelog: Title
(cherry picked from commit 655b0816301b91371b98cc95d1c7ffad4d5488c8)

### DIFF
--- a/cfe_internal/enterprise/CFE_knowledge.cf
+++ b/cfe_internal/enterprise/CFE_knowledge.cf
@@ -555,6 +555,14 @@ body depth_search cfe_internal_docroot_application_perms
 
 body perms state_dir_system_owned
 {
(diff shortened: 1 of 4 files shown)


vpodzime (approved):
ACK

PR #2266, opened 2022-03-10
DONOTMERGE: Use 'rxdirs => "true"' explicitly

vpodzime:
Effectively reverting the change of the default for rxdirs.


cf-bottom:

nickanderson:
@vpodzime we can close this one now right?

PR #2269, opened 2022-03-14
ENT-7505 Add debug option for federated reporting import process (3.18)

craigcomstock:
Make it possible to configure the system to include debug logs
which are reviewable to aid in issue analysis.

Ticket: ENT-7505
Changelog: title

### DIFF
--- a/templates/federated_reporting/import.sh
+++ b/templates/federated_reporting/import.sh
@@ -18,6 +18,7 @@ true "${CFE_FR_EXTRACTOR?undefined}"
 true "${CFE_FR_TABLES?undefined}"
 true "${CFE_FR_INVENTORY_REFRESH_CMD?undefined}"
 true "${CFE_FR_HANDLE_DUPLICATES?undefined}"
+true ${CFE_FR_PSQL_OPTIONS?undefined}"
 
 if ! type "$CFE_FR_EXTRACTOR" >/dev/null; then
   log "Extractor $CFE_FR_EXTRACTOR not available!"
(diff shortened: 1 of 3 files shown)


templates/federated_reporting/import.sh
```diff
@@ -120,13 +121,15 @@ if [ "$CFE_FR_HANDLE_DUPLICATES" = "yes" ]; then
 fi
 
 failed=0
+rm -f "$CFE_FR_SUPERHUB_IMPORT_DIR/schema_attach.log"
 log "Attaching schemas"
 for file in $dump_files; do
   if [ ! -f "${file}.failed" ]; then
     hostkey=$(basename "$file" | cut -d. -f1)
     "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
+                        --set "client_min_messages=DEBUG" \
```
craigcomstock, line 130: @nickanderson @vpodzime do you think we should have an option say in https://github.com/cfengine/masterfiles/blob/master/templates/federated_reporting/config.sh.mustache to enable debug logging here and maybe other places we run psql stuff? Save the last log always like here where I `rm -f` the previous? Seems like a good idea as a means to get more information out in case of problems. Something we could advise folks to do in case of bug reports and such.

nickanderson: Yeah,  as a general rule it's really nice to be able to turn easily something on to get more debug output

vpodzime: Absolutely.

craigcomstock:
stale

PR #2274, opened 2022-03-18
ENT-8464: Use FQDN for cert path in distributed cleanup

craigcomstock:
Testing was done on machines where platform.node() was
the same as socket.getfqdn() but this is often not
the case.

Ticket: ENT-8464
Changelog: title

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1490,6 +1490,10 @@ For example, to define this class via Augments:
 }
 ```
 
+**History:**
+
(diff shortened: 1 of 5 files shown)


cf-bottom:

nickanderson:
> Thank you for submitting a PR! Maybe @nickanderson can review this?

@vpodzime would be a better person to review this.

vpodzime (approved):
ACK

PR #2288, opened 2022-04-14
ENT-8547: Removed comments about underscore prefixed vars not being reported by default

nickanderson:
This has not been true since 3.6.0 when variables_exclude was removed. Prior to
it's removal it did indeed exclude underscore prefixed variables as part of
bce7032f603deb2b8df8f87a26838816d6d2266d.

Ticket: ENT-8547
Changelog: None


vpodzime (approved):
ACK

PR #2309, opened 2022-04-20
ENT-8501: Stopped loading Apache mod_log_forensic by default on Enterprise Hubs (3.18)

nickanderson:
Merge Together:
- https://github.com/cfengine/buildscripts/pull/985

default.

Ticket: ENT-8501
Changelog: Title
(cherry picked from commit 15a7df927d57c76d4190cb9b74fb3de430415ed9)


cf-bottom:

PR #2354, opened 2022-04-25
ENT-8524: Stopped loading Apache mod_filter and mod_ext_filter by default on Enterprise Hubs (3.15)

nickanderson:
Merge Together:
- https://github.com/cfengine/buildscripts/pull/1030

We do not use the features provided by these modules, so we should not load them by default.


cf-bottom:

PR #2378, opened 2022-05-11
ENT-8610: Stopped loading Apache mod_expires by default on Enterprise Hubs (3.18)

nickanderson:
We do not use the features provided by this module, so we should not load it by default.

Merge Together:
- https://github.com/cfengine/buildscripts/pull/1052


nickanderson:
@cf-bottom jenkins, please

PR #2383, opened 2022-05-16
ENT-8617: Added PostgreSQL tunables for Federated Reporting

nickanderson:
This change instruments tunables for `postgresql.conf` when Federated Reporting is
enabled.

The existing defaults for `shared_buffers`, `max_locks_per_transaction`, and
`log_lock_waits` have been exposed as `cfengine_enterprise_federation.postgres_config.shared_buffers`,
`cfengine_enterprise_federation.postgres_config.max_locks_per_transaction`, and
`cfengine_enterprise_federation.postgres_config.log_lock_waits` respectively.

Additionally, since the superhub common needs `checkpoint_timeout` and
`max_wal_size` adjusted, they have been added to the default
configuration with their default values and exposed as
`cfengine_enterprise_federation.postgres_config.checkpoint_timeout` and
`cfengine_enterprise_federation.postgres_config.max_wal_size` respectively.

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1510,6 +1510,106 @@ When custom certificates are in use distributed cleanup needs to know where to f
 
 * Added in CFEngine 3.20.0, 3.18.2
 
+#### PostgreSQL Configuration
+
+It's not uncommon to need to configure some PostgreSQL settings differently for Federated Reporting. The settings that are exposed as tunables which can be set via augments are listed here. These do not comprise all settings that may need adjusted, only those that are most commonly adjusted.
+
+**Note:** When [setting parameters for the PostgreSQL configuration](https://www.postgresql.org/docs/current/config-setting.html)
+file various units can be used. Valid memory units are B (bytes), kB
+(kilobytes), MB (megabytes), GB (gigabytes), and TB (terabytes). The multiplier
+for memory units is 1024, not 1000. Valid time units are us (microseconds), ms
+(milliseconds), s (seconds), min (minutes), h (hours), and d (days).
+
+##### shared_buffers
+
+Shared buffers are the amount of memory the database server uses for shared memory buffers. Settings significantly higher than the minimum are usually needed for good performance.
+
+The value should be set to 15% to 25% of the machine's total RAM. For example: if your machine's RAM size is 32 GB, then the recommended value for shared_buffers is 8 GB.
+
+To adjust this set `cfengine_enterprise_federation:postgres_config.shared_buffers` via Augments.
+
+For example:
+
+```json
+{
+  "variables": {
+    "cfengine_enterprise_federation:postgres_config.shared_buffers": "2560MB"
+  }
+}
+```
+
+**History:**
+
+* Added in 3.20.0, 3.18.2
+
+##### max_locks_per_transaction
+
+The ```max_locks_per_transaction``` value indicates the number of database objects that can be locked simultaneously. When Federated Reporting is enabled, the MPF default is `4000`.
+
+```json
+{
+  "variables": {
+    "cfengine_enterprise_federation:postgres_config.max_locks_per_transaction": "4100"
+  }
+}
+```
+
+**History:**
+
+* Added in 3.20.0, 3.18.2
+
+##### log_lock_waits
+
+Controls whether a log message is produced when a session waits longer than `deadlock_timeout` to acquire a lock. This is useful in determining if lock waits are causing poor performance. When Federated Reporting is enabled, the MPF default is `on`.
+
+```json
+{
+  "variables": {
+    "cfengine_enterprise_federation:postgres_config.log_lock_waits": "off"
+  }
+}
+```
+
+**History:**
+
+* Added in 3.20.0, 3.18.2
+
+##### max_wal_size
+
+Sets the WAL size that triggers a checkpoint.
+
+Maximum size to let the WAL grow during automatic checkpoints. This is a soft limit; WAL size can exceed `max_wal_size` under special circumstances, such as heavy load, a failing `archive_command`, or a high `wal_keep_size` setting. If this value is specified without units, it is taken as megabytes. The default is 1 GB (`1024MB`). Increasing this parameter can increase the amount of time needed for crash recovery.
+
+```json
+{
+  "variables": {
+    "cfengine_enterprise_federation:postgres_config.max_wal_size": "20G"
+  }
+}
+```
+
+**History:**
+
+* Added in 3.20.0, 3.18.2
+
+##### checkpoint_timeout
+
+Sets the maximum time between automatic WAL checkpoints.
+
+Maximum time between automatic WAL checkpoints. If this value is specified without units, it is taken as seconds. The valid range is between 30 seconds and one day. The default is five minutes (`5min`). Increasing this parameter can increase the amount of time needed for crash recovery.
+
+```json
+{
+  "variables": {
+    "cfengine_enterprise_federation:postgres_config.checkpoint_timeout": "30min"
+  }
+}
+```
+
+**History:**
+
+* Added in 3.20.0, 3.18.2
+
 ## Recommendations
 
 The MPF includes policy that inspects the system and makes recommendations about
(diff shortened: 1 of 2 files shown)


nickanderson:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/7933/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/7933/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-7933/

MPF.md
```diff
@@ -1510,6 +1510,100 @@ When custom certificates are in use distributed cleanup needs to know where to f
 
 * Added in CFEngine 3.20.0, 3.18.2
 
+#### PostgreSQL Configuration
```
craigcomstock, line 1513: change section heading to "Federated Reporting PostgreSQL Server Configuration" or similar... aka mention Federated Reporting.

MPF.md
```diff
@@ -1510,6 +1510,100 @@ When custom certificates are in use distributed cleanup needs to know where to f
...
+
+It's not uncommon to need to configure some PostgreSQL settings differently for Federated Reporting. The settings that are exposed as tunables which can be set via augments are listed here. These do not comprise all settings that may need adjusted, only those that are most commonly adjusted.
+
+##### shared_buffers
```
craigcomstock, line 1517: For all of these we might include a link to the respective postgresql doc pages so folks know things like what units can be used and such. e.g. 2560MB or 2GB etc.

MPF.md
```diff
@@ -1510,6 +1510,100 @@ When custom certificates are in use distributed cleanup needs to know where to f
...
+
+**History:**
+
+* Added in 3.20.0, 3.18.2
+
+##### max_wal_size
+
+Sets the WAL size that triggers a checkpoint.
+
+Maximum size to let the WAL grow during automatic checkpoints. This is a soft limit; WAL size can exceed `max_wal_size` under special circumstances, such as heavy load, a failing `archive_command`, or a high `wal_keep_size` setting. If this value is specified without units, it is taken as megabytes. The default is 1 GB (`1024MB`). Increasing this parameter can increase the amount of time needed for crash recovery.
```
craigcomstock, line 1575: might want to add that for large data load scenarios a large value is recommended, such as in Federated Reporting: 20GB is a recommended starting point for anything but demo scenarios.

MPF.md
```diff
@@ -1510,6 +1510,100 @@ When custom certificates are in use distributed cleanup needs to know where to f
...
+
+**History:**
+
+* Added in 3.20.0, 3.18.2
+
+##### checkpoint_timeout
+
+Sets the maximum time between automatic WAL checkpoints.
+
+Maximum time between automatic WAL checkpoints. If this value is specified without units, it is taken as seconds. The valid range is between 30 seconds and one day. The default is five minutes (`5min`). Increasing this parameter can increase the amount of time needed for crash recovery.
```
craigcomstock, line 1593: Also mention why you would increase this value to our FR default of 30min: because if the timeout is too frequent the wal actually grows bigger due to full page writes needed after each checkpoint for new changes. See https://www.cybertec-postgresql.com/en/checkpoint-distance-and-amount-of-wal/

TL;DR: 

> The point now is: If checkpoints are close together there are many “first times” and full pages have to be written to the WAL quite frequently. However, if checkpoints are far apart the number of full page writes will drop dramatically leading to a lot less WAL.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -530,18 +530,40 @@ bundle agent federation_manage_files
...
+
+      "c[max_wal_size]" -> { "ENT-8617" }
+        string => ifelse( isvariable( "cfengine_enterprise_federation.postgres_config.max_wal_size"),
+                          $(cfengine_enterprise_federation.postgres_config.max_wal_size),
+                          "1GB");
+
+      "c[checkpoint_timeout]" -> { "ENT-8617" }
+        string => ifelse( isvariable( "cfengine_enterprise_federation.postgres_config.checkpoint_timeout"),
+                          $(cfengine_enterprise_federation.postgres_config.checkpoint_timeout),
+                          "5min");
```
craigcomstock, line 565: I really don't know what a good value for this is, given the un-predictable nature of when feeders dump and what cycle the checkpoint timeout might be in and whether wal gets exhausted :( Maybe something higher like 30min or 40min with the idea that we want to avoid automatic checkpoints during EVERY import cycle? I'm not sure. I suppose ideally a checkpoint might occur immediately AFTER an import actually, but postgres docs, https://www.postgresql.org/docs/current/sql-checkpoint.html, mention not to use that in "normal operations".

MPF.md
nickanderson: Since these tunables are only relevant in the context of federated reporting superhub I figured there was no need to mention that. None of these tunables will be applied if the hub is not enabled for federated reporting.

I figure if we want to suggest better defaults, we should instead change our defaults in policy. The hint that I left here is that the augments example values are using the values that we have recently suggested setting.

MPF.md
nickanderson: Since these tunables are only relevant in the context of federated reporting superhub I figured there was no need to mention that. None of these tunables will be applied if the hub is not enabled for federated reporting.

I figure if we want to suggest better defaults, we should instead change our defaults in policy. The hint that I left here is that the augments example values are using the values that we have recently suggested setting.

cfe_internal/enterprise/federation/federation.cf
nickanderson: Since these tunables are only relevant in the context of federated reporting superhub I figured there was no need to mention that. None of these tunables will be applied if the hub is not enabled for federated reporting.

I figure if we want to suggest better defaults, we should instead change our defaults in policy. The hint that I left here is that the augments example values are using the values that we have recently suggested setting.

MPF.md
nickanderson: Really?

I guess that both of the other headings here also explicitly say Federated Reporting, but it feels redundant given the heading levels.

vpodzime: Yeah, it's nested properly.

MPF.md
vpodzime: Yeah, we should have good defaults instead of telling people to change our defaults to better default values.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -530,18 +530,40 @@ bundle agent federation_manage_files
 
 bundle agent postgres_config
 # @brief Customize postgres config for superhub
+# @variable cfengine_enterprise_federation.postgres_config.shared_buffers - Sets the maximum number of locks per transaction.
```
vpodzime, line 533: Shouldn't `cfengine_enterprise_federation` be a namespace? What is it otherwise?

vpodzime (commented):
Looks good to me otherwise.

cfe_internal/enterprise/federation/federation.cf
nickanderson: Good eye! Will fix.

nickanderson:
@craigcomstock, @vpodzime please re-review

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -530,18 +530,40 @@ bundle agent federation_manage_files
...
+# @variable cfengine_enterprise_federation:postgres_config.log_lock_waits - Logs long lock waits.
+# @variable cfengine_enterprise_federation:postgres_config.max_wal_size - Sets the WAL size that triggers a checkpoint.
+# @variable cfengine_enterprise_federation:postgres_config.checkpoint_timeout- Sets the maximum time between automatic WAL checkpoints.
 {
   vars:
     am_superhub::
-      "c[shared_buffers]"
-        string => "1GB",
+      "c[shared_buffers]" -> { "ENT-8617" }
+        string => ifelse( isvariable( "cfengine_enterprise_federation.postgres_config.shared_buffers"),
```
vpodzime, line 542: These should also use a namespace, AFAICT.

nickanderson: bah, should have project wide search replace. DONE now.

vpodzime (commented):
Looks good to me, but squash the commits, please.

nickanderson:
Squashed.

vpodzime (approved):
ACK

PR #2397, opened 2022-06-06
CFE-3976 Use simple numeric comparison in apt_get package module

craigcomstock:
Assuming that apt-get -v always returns simple numeric versions
and that differences outside of the first two digits aren't significant
we just compare those first two digits as numbers.

Ticket: CFE-3796
Changelog: title


modules/packages/vendored/apt_get.mustache
```diff
@@ -32,7 +31,9 @@ apt_get_options = ["-o", "Dpkg::Options::=--force-confold",
                    "-o", "Dpkg::Options::=--force-confdef",
                    "-y"]
 
-if LooseVersion(apt_version) < LooseVersion("1.1"):
+# assume apt-get -v returns a simple n.n or n.n.n version so take the first
+# two numbers and convert to a float
+if float(".".join(apt_version.split(".")[0:2])) < float("1.1"):
```
olehermanse, line 36: I'd suggest not converting to float here, comparing 2 lists of ints does what you want, I think:

```suggestion
if [int(x) for x in apt_version.split(".")[0:2]] < [1, 1]:
```

olehermanse (changes requested):
Please update the commit message with an explanation of why, and fix the typo in the ticket number.

modules/packages/vendored/apt_get.mustache
craigcomstock: I'll try it out.

craigcomstock:
@cf-bottom exotics on jenkins please

modules/packages/vendored/apt_get.mustache
craigcomstock: I reverted to an even simpler solution, string lexical order. Should work in the only ints case as well as things like a, alpha, beta which I don't expect regardless.

craigcomstock:
Retry exotics build after fixing up tests_skip_needs_work definitions to not clobber each other. Should prohibit running package acceptance tests on non redhat, debian, x86/amd64 platforms as before as well as excluding aarch64 as required with the addition of that platform (ubuntu 22).

[Build Status](https://ci.cfengine.com/job/pr-pipeline/8065/)

craigcomstock:
Looking carefully at the CI result it would seem that exotics generally pass so I believe we can move forward with this PR.

modules/packages/vendored/apt_get.mustache
```diff
@@ -32,7 +31,8 @@ apt_get_options = ["-o", "Dpkg::Options::=--force-confold",
                    "-o", "Dpkg::Options::=--force-confdef",
                    "-y"]
 
-if LooseVersion(apt_version) < LooseVersion("1.1"):
+# assume apt-get -v returns faily simple version with integers separated by dots
+if [int(x) for x in apt_version.split(".")] < [1, 1]:
```
nickanderson, line 35: I'm not a pythonista, but this looks ok to me.
apt_get -v si processed a few lines up, getting the first line aplitting on spaces and taking the second (first if counting from 0) field. Then your if checks each part of the dotted version.
not clear to me though how the last dotted version is compared or what it's compared against.

```
❯ apt-get -v
apt 2.4.5 (amd64)
Supported modules:
*Ver: Standard .deb
 Pkg:  Debian APT solver interface (Priority -1000)
 Pkg:  Debian APT planner interface (Priority -1000)
*Pkg:  Debian dpkg interface (Priority 30)
 S.L: 'deb' Debian binary tree
 S.L: 'deb-src' Debian source tree
 Idx: EDSP scenario file
 Idx: EIPP scenario file
 Idx: Debian Source Index
 Idx: Debian Package Index
 Idx: Debian Translation Index
 Idx: Debian dpkg status file
 Idx: Debian deb file
 Idx: Debian dsc file
 Idx: Debian control file
```

craigcomstock: lists in python are compared left to right.

tests/acceptance/17_packages/README
```diff
@@ -0,0 +1,3 @@
+2022-06-07 added meta_skip.cf.sub here in masterfiles/tests/acceptance/17_packages
```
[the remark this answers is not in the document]
craigcomstock: Right. I worried about this a bit as well. Thus naming it meta_skip.cf. I was super annoyed when I had to edit so many files to make the same change over and over: aka violation of the "don't repeat yourself" principle.

I did add a README! :O

craigcomstock:
> From the commit message related to the geneerated test, make the generator warn complain if py2 is used?

Indeed, that would kind of nice as well.

tests/acceptance/17_packages/README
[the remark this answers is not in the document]
craigcomstock: I could rename meta_skip.cf to skip_unsupported_platforms.cf, even though it is both skip_unsupported and skip_needs_work. I also thought about adding a comment to each file or just having it be common.cf which would be even BETTER in terms of DRY but worse in terms of visibility. :(

tests/acceptance/17_packages/10_new/unsafe/README
```diff
@@ -0,0 +1,10 @@
+Run this generator and redirect the output to the resulting policy file.
```
craigcomstock, line 1: @nickanderson I added a README and added code to the generator to exit if not python 3.

olehermanse (changes requested):
@craigcomstock CFE-3796 is **not** the right ticket number, please fix it. (First commit and description).

tests/acceptance/17_packages/11_old/unsafe/package-inventory.cf
```diff
@@ -13,18 +13,13 @@ body common control
...
     bundlesequence => { default($(this.promise_files)) };
 }
 
 bundle agent init
 {
-  meta:
-      "test_skip_needs_work" string => "!redhat.!debian";
-      # RedHat 4 RPM has a bug which corrupts the RPM DB during our tests, so it is untestable.
-      # And available patches is an Enterprise feature.
-      "test_skip_unsupported" string => "redhat_4|centos_4|!enterprise";
```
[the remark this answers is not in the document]
craigcomstock: ah, right, I'll fix that.

vpodzime (commented):
The first commit should explain why the change is being made. And the last one should explain the reason for such a huge change in the generated test.

craigcomstock:
> The first commit should explain why the change is being made. And the last one should explain the reason for such a huge change in the generated test.

Check. Addressed and thank you. :)

vpodzime (approved):
ACK

craigcomstock:
cherry pick CFE-3976 apt package module script fix to
3.15 https://github.com/cfengine/masterfiles/pull/2413
3.18 https://github.com/cfengine/masterfiles/pull/2412

craigcomstock:
> The other two commits could be cherry picked as well, but can wait until after release since they are related to "test generation by developers" and "arm platform" which we won't be supporting in 3.15.6 and 3.18.2.

other two commits cherry picked now:
3.15: https://github.com/cfengine/masterfiles/pull/2436
3.18: https://github.com/cfengine/masterfiles/pull/2437

PR #2402, opened 2022-06-15
ENT-8823: Added policy to patch apachectl for more robust stopping on Enterprise Hubs

nickanderson:
The stock apachectl does not wait to see if a signal had the desired effect. In
the case of stopping and restarting this can result in httpd processes being
left running preventing new configuration from taking effect.

This change renders a patched apachectl which is instrumented to briefly wait
for httpd to be fully terminated and takes more aggressive actions to ensure
httpd is fully stopped before exiting.

This script is part of the package but needs to also be handled by the MPF to
cover the case of upgrade where policy from the newer version is running before
binaries from the newer version are running.

### DIFF
--- a/cfe_internal/enterprise/mission_portal.cf
+++ b/cfe_internal/enterprise/mission_portal.cf
@@ -16,8 +16,32 @@ bundle agent cfe_internal_enterprise_mission_portal
         comment => "Manage Apache Web server (on/off)";
 
 }
+bundle agent apachectl_patched_for_upgrade
+# @brief Ensure that apacehctl is patched so that it is able to re-start services
+#
+# @description This bundle addresses an issue where upgrades fail trying to stop
+# httpd. Versions prior to 3.20.0, 3.18.2, and 3.15.6 need to have apachectl
+# patched to make it wait for processes to shut down before exiting before
+# upgrading binaries to 3.20.0, 3.18.2, or 3.15.6. CFEngine packages for
+# versions after 3.20.0, 3.18.2, 3.15.6 ship with this patched apachectl which
+# should make this bundle a no-op.
+#
+# TODO Redact when 3.21.0 is the oldest supported version
+{
+  files:
+      "$(sys.bindir)/apachectl"
+        edit_template => "$(this.promise_dirname)/templates/apachectl.mustache",
+        handle => "apachectl_content",
+        template_method => "mustache",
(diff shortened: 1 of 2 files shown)


nickanderson:
@cf-bottom jenkins, pleae

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/8101/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/8101/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-8101/

craigcomstock (approved):

cfe_internal/enterprise/mission_portal.cf
```diff
@@ -43,6 +67,9 @@ bundle agent cfe_internal_enterprise_mission_portal_apache
 
   methods:
 
+      "apachectl patched for resilient stop " -> { "ENT-8823" }
+        usebundle => apachectl_patched_for_upgrade;
```
nickanderson, line 71: @Lex-2008 I know the ticket says:

> CFEngine versions where this should be done are probably these:
cfengine_3_18_0|cfengine_3_18_1|
cfengine_3_15.!(cfengine_3_15_6|cfengine_3_15_7)|
cfengine_3_19|cfengine_3_12|cfengine_3_10|cfengine_3_7|
cfengine_3_17|cfengine_3_16|cfengine_3_14|cfengine_3_13|cfengine_3_11|cfengine_3_9|cfengine_3_8|cfengine_3_6|cfengine_3_5

But, my logic is that all versions need this. If we ship a patched apachectl in the packages from 3.20.0, 3.18.2, and 3.15.6 forward the rendered template here should match that identically resulting in a no-op, promise kept. If they differ, of course this one will be rendered.

And once 3.21.0 is the oldest supported version, we should be able to redact this from policy and rely on the fact that the patched apachectl is in place from packages 3.20.0, 3.18.2, or 3.15.6 or later.

vpodzime: Assuming people won't run newer versions of CFEngine that might have further improved versions of *apachectl* with this version of MPF.

vpodzime (approved):
I'd personally be quite strict with the versions that really need this. But it's not a big deal.

Lex-2008 (approved):

cfe_internal/enterprise/mission_portal.cf
nickanderson: Or assuming that we update apachectl in the MPF as well.

PR #2407, opened 2022-06-21
ENT-8823: Fixed path to apachectl

nickanderson:
It's not with the rest of the binaries, it's in $WORKDIR/httpd/bin


vpodzime (approved):
ACK

PR #2410, opened 2022-06-21
ENT-8823: Fixed and refactored apachectl rendering to handle older versions of CFEngine (3.15)

nickanderson:


nickanderson:
@cf-bottom jenkins, please

PR #2413, opened 2022-06-23
CFE-3976 Use simple numeric comparison in apt_get package module (3.15)

craigcomstock:
Previously we used distutils.LooseVersion but distutils is being
deprecated in newer Python versions, such as that available in Ubuntu 22,
so we must migrate to something else.

After looking around a bit and considering our use of LooseVersion it
seems a simple version of this check is all that is needed.

Assuming that apt-get -v always returns simple numeric versions
we compare versions as a list of integers.

Ticket: CFE-3976
Changelog: title
(cherry picked from commit 34cfe809154280b1bb0fa8e1477d8b5c8ee6e8a0)

### DIFF
--- a/modules/packages/vendored/apt_get.mustache
+++ b/modules/packages/vendored/apt_get.mustache
@@ -7,7 +7,6 @@ import sys
 import os
 import subprocess
 import re
-from distutils.version import LooseVersion
 
 PY3 = sys.version_info > (3,)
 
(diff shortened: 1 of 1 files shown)


craigcomstock:
@cf-bottom jenkins

nickanderson:
@Lex-2008  we need a changelog entry from this one.

Lex-2008:
@nickanderson : https://github.com/cfengine/masterfiles/pull/2416

PR #2415, opened 2022-06-23
ENT-8847: Made cmdb update ignore locks (3.18)

nickanderson:
This change is intended to improve the smoothness with respect to unscheduled
agent runs in relation to CMDB data update. Specifically so that the default 1
minute locking period does not prevent an unscheduled agent run from updating
it's CMDB data.

For example, when Trigger agent run button is clicked, a request is made for the
host to do an agent run. If there was an agent less than one minute ago the CMDB
data would not update because of the lock.

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -62,8 +62,9 @@ bundle agent cfe_internal_update_policy
         usebundle => cfe_internal_update_policy_cpv;
 
     any::
-      "CMDB data update" -> { "ENT-6788" }
-        usebundle => cfe_internal_update_cmdb;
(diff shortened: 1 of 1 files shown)


nickanderson:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/8153/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/8153/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-8153/

vpodzime (approved):
ACK

PR #2425, opened 2022-06-30
ENT-8915: Made it possible to cfbs add the masterfiles repository directly

nickanderson:
I imagine this is most useful for testing or those who want to live on the edge.
Installing masterfiles directly prohibits cfbs upgrade functionality, commit sha
changing must be done manually.


nickanderson:
What do you think @olehermanse ?

olehermanse:
Makes sense to me :) I kinda thought we already did this 👨‍🔬

PR #2432, opened 2022-07-19
ENT-9010: Fixed directory in which windows agents source packages for upgrade (3.18)

nickanderson:
In 3.18.0 sys.flavor on windows changed so that it includes the major OS version
as is the case for most other platforms. That change broke the expectation of
where to find software during self upgrade. This change simply re-aligns the
source directory used. Since we only build one package for each architecture we
keep using a single directory instead of a separate directory per windows version.


nickanderson:
@ccoager How does this look to you?

ccoager:
@nickanderson This works for me on Windows.

PR #2435, opened 2022-07-25
Use '$(default:paths.something)' in federation.cf

vpodzime:
We define a separate namespace for the FR policy in federation.cf
and the 'paths' bundle is in the default namespace.

Ticket: ENT-8817
Changelog: None


nickanderson (approved):
Looks correct to me.

craigcomstock (approved):

PR #2438, opened 2022-07-25
ENT-8628: Add inventory for Raspberry Pi and DeviceTree devices

craigcomstock:
Raspberry Pi kernel has odd entries in /proc/cpuinfo: Model, Hardware, Revision

Also present is /proc/device-tree which generaly excludes the presence of usable dmi information from dmidecode.

Use device tree model, serial-number and compatible data for inventory and defining appropriate classes describing the host.

Ticket: ENT-8628
Changelog: title

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -633,6 +633,7 @@ bundle common inventory_control
     !freebsd::
 
       "dmidecoder" string => "/usr/sbin/dmidecode";
+      "proc_device_tree" string => "/proc/device-tree";
 
   classes:
       # setting this disables all the inventory modules except package_refresh
@@ -646,10 +647,15 @@ bundle common inventory_control
       "disable_inventory_lsb" expression => "disable_inventory";
       "disable_inventory_lsb" not => fileexists($(lsb_exec));
 
+      # If we have /proc/device-tree we should likely disable dmi completely
+      # as of 2022 systems with dmi dont have device-tree and vice versa.
+      "have_proc_device_tree" expression => fileexists($(proc_device_tree));
+
       # by default disable the dmidecode inventory if the general
       # inventory is disabled or the binary does not exist.  Note that
       # typically this is a very fast binary.
       "disable_inventory_dmidecode" expression => "disable_inventory";
+      "disable_inventory_dmidecode" expression => "have_proc_device_tree";
       "disable_inventory_dmidecode" not => fileexists($(dmidecoder));
 
       # by default disable the LLDP inventory if the general inventory
--- a/inventory/linux.cf
+++ b/inventory/linux.cf
@@ -7,6 +7,15 @@ bundle common inventory_linux
 #  systemd class based on linktarget of /proc/1/cmdline
 {
   vars:
+    have_proc_device_tree::
+      "proc_device_tree_model" string => readfile("/proc/device-tree/model"),
+      comment => "Read model from /proc/device-tree",
+      meta => { "inventory", "attribute_name=System version" };
+
+      "proc_device_tree_serial_number" string => readfile("/proc/device-tree/serial-number"),
+      meta => { "inventory", "attribute_name=System serial number" };
+
+
     has_proc_1_cmdline::
       "proc_1_cmdline_split" slist => string_split(readfile("/proc/1/cmdline", "512"), " ", "2"),
       comment => "Read /proc/1/cmdline and split off arguments";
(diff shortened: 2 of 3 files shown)


inventory/linux.cf
```diff
@@ -7,6 +7,19 @@ bundle common inventory_linux
...
+      comment => "Read model from /proc/device-tree",
+      meta => { "inventory", "attribute_name=System version" };
+
+      "proc_device_tree_serial_number" string => readfile("/proc/device-tree/serial-number"),
+      meta => { "inventory", "attribute_name=System serial number" };
+
+      # the compatible proc file here separates groups of entries with a null byte, so replace that with a comma and get several descriptive classes
+      # e.g. raspberrypi,4-model-b\000brcm,bcm2711\000
+      "proc_device_tree_compatible" string => execresult("$(paths.cat) /proc/device-tree/compatible | $(paths.sed) 's/\x00/,/'", useshell);
+      "proc_device_tree_compatible_list" slist => splitstring("$(proc_device_tree_compatible)", ",", 20);
```
nickanderson, line 21: Were you trying to get `raspberrypi_4_model_b` or `raspberrypi` and `4_model_b` classes a bit later https://github.com/cfengine/masterfiles/pull/2438/files#diff-3ee6fca00d4a49c79bc30911ddb39159e79c4d14241fc2aeb38dc8de0076eef8R74

nickanderson: if you want `raspberrypi_4_model_b` then maybe replace the null byte with something else to split on instead of comma, like `MYSPLIT`

nickanderson (commented):
Looks pretty good, just a question about some classes.

inventory/linux.cf
craigcomstock: I thought each individual thing made more sense to me: `raspberrypi` as a class seems VERY useful and of the two devices I sampled all of the individual bits seemed appropriate:

`raspberrypi, 4-model-b, brcm, bcm2711`

`pine64, pinephone-1.2, pinephone, allwinner, sun50i-a64`

It's fairly arbitrary but I thought it would be a good place to start.

controls/def.cf
```diff
@@ -633,6 +633,7 @@ bundle common inventory_control
     !freebsd::
 
       "dmidecoder" string => "/usr/sbin/dmidecode";
+      "proc_device_tree" string => "/proc/device-tree";
```
craigcomstock, line 636: I don't love this being here, feels like it should be in inventory/linux.cf, or push the have_proc_device_tree class here and reference in both any.cf and linux.cf.

inventory/linux.cf
```diff
@@ -48,12 +61,19 @@ bundle common inventory_linux
   classes:
 
     any::
+      "have_proc_device_tree" expression => fileexists($(inventory_control.proc_device_tree));
```
craigcomstock, line 64: this seems a duplication of the same line in any.cf: https://github.com/cfengine/masterfiles/pull/2438/files#diff-e517c74d74557aab06f37a22ffb428372e0172c21f4293c38ac79ecafbfde986R980, maybe move this to controls/def.cf?

inventory/linux.cf
olehermanse: Defining `raspberrypi`, `raspberrypi_4`, and `raspberrypi_4_model_b` make a lot of sense to me. The `4_model_b` without `raspberrypi` in front - not so much.

craigcomstock: agreed, was just trying a catch all without much fiddling.

craigcomstock: see https://tracker.mender.io/browse/ENT-9029, let's defer parsing this business for now. :)

craigcomstock:
working on a few refactorings I wanted to do... not quite working so will test/finish tomorrow :)

craigcomstock:
3.12.x travis failure is known issue: https://tracker.mender.io/browse/ENT-9028

controls/def.cf
```diff
@@ -646,10 +647,15 @@ bundle common inventory_control
...
 
+      # If we have /proc/device-tree we should likely disable dmi completely
+      # as of 2022 systems with dmi dont have device-tree and vice versa.
+      "have_proc_device_tree" expression => fileexists($(proc_device_tree));
+
       # by default disable the dmidecode inventory if the general
       # inventory is disabled or the binary does not exist.  Note that
       # typically this is a very fast binary.
       "disable_inventory_dmidecode" expression => "disable_inventory";
+      "disable_inventory_dmidecode" expression => "have_proc_device_tree";
```
nickanderson, line 658: I don't think this does what you perhaps hope it does.

This class protects the actuation of the bundle after the agent has started executing the bundle sequence, but that does not necessarily protect against vars and classes resolution in `bundle agent cfe_autorun_inventory_dmidecode`. All of those inventory autorun bundles are affected in the same way, defiing the class isn't necessarily going to prevent them from resolving those things, but they will have less opportunity.

To really disable it, those bundles need to be treated and have guards against resolving variables when the class is defined.
Really annoying, and I am not suggesting that you not do this or that you go the extra mile, just highlighting the fact.

nickanderson (approved):

controls/def.cf
craigcomstock: No, that makes sense. vars/classes are different. ack.

craigcomstock: Taking a further look I think I'll skip a bigger refactor. Let me know and we can just remove this line as it works fine without the inhibition as well.

craigcomstock:
cherry picked to 3.18.x https://github.com/cfengine/masterfiles/pull/2440

PR #2448, opened 2022-08-02
CFE-4017: Fixed services_autorun_inputs (3.15)

nickanderson:


nickanderson:
services_autorun_inputs has not been backported to 3.15, so not going to do it now.

PR #2485, opened 2022-09-09
ENT-8456: Adapt the cf-serverd.service template to the version from core

vpodzime:
Reflecting the changes from https://github.com/cfengine/core/commit/962bac57c2af.

Merge together:
https://github.com/cfengine/core/pull/5045
https://github.com/cfengine/masterfiles/pull/2485

### DIFF
--- a/templates/cf-serverd.service.mustache
+++ b/templates/cf-serverd.service.mustache
@@ -1,7 +1,7 @@
 [Unit]
 Description=CFEngine Enterprise file server
 After=syslog.target
-After=network.target
+After=network-online.target
 ConditionPathExists={{{vars.sys.bindir}}}/cf-serverd
 ConditionPathExists={{{vars.sys.workdir}}}/policy_server.dat
 ConditionPathExists={{{vars.sys.default_policy_path}}}
@@ -14,5 +14,5 @@ Restart=always
 RestartSec=10
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=network-online.target
 WantedBy=cfengine3.service


cf-bottom:

olehermanse (approved):
LGTM

craigcomstock (approved):

craigcomstock (changes requested):
Which core file does this try to duplicate? https://github.com/cfengine/core/blob/master/misc/systemd/cf-serverd.service.in is different than this one you propose.

Might be nice to add some links back and forth so we know what's going on here. I assume we take the service files from core for packaging? And then later after bootstrap, masterfiles could overwrite with what it has?

vpodzime:
> Might be nice to add some links back and forth so we know what's going on here.

The commit has a link to the related commit in core that changes the file.

vpodzime:
> I assume we take the service files from core for packaging? And then later after bootstrap, masterfiles could overwrite with what it has?

I'm not sure how this works and I don't want to care. :) I'm just trying to keep things consistent between the two places.

craigcomstock:
> > I assume we take the service files from core for packaging? And then later after bootstrap, masterfiles could overwrite with what it has?
> 
> I'm not sure how this works and I don't want to care. :) I'm just trying to keep things consistent between the two places.

Interesting. The link, https://github.com/cfengine/core/commit/3ccb427e933a5, is 404 for me and the file I referenced in core certainly is not the same as what you have here. I do think what you have here looks like an improvement though, so maybe double-back and fix the other spots?

craigcomstock (changes requested):
Fixup link to core change in comment/PR description, it is 404 for me. 

Also maybe fix the core version I linked to as well since it doesn't agree with this one.

vpodzime:
> Fixup link to core change in comment/PR description, it is 404 for me.

Oh, there was a rebase. It should be cfengine/core@962bac57c2af

vpodzime:
> Also maybe fix the core version I linked to as well since it doesn't agree with this one.

I'm changing both files at the same time. Here and in the related core PR.

craigcomstock (approved):

PR #2496, opened 2022-09-22
Remove mindepth argument from find since not all platforms support it

Lex-2008:
The change is OK now because the selection of the files is more careful
after a recent change, quoting @craigcomstock:
> before I added the search for name `*.py and mock_*` it was running
> `sed -ri` on every file under that dir.

Platforms which don't support it are AIX 5 and HP-UX.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -5,7 +5,7 @@ SUBDIRS += acceptance
 endif
 
 fix-python-hashbang:
-	test -x /usr/bin/python || find $(srcdir) -mindepth 2 \( -name '*.py' -o -name 'mock_*' -o -name 'test_*' \) -exec sed -ri '\~/usr/bin/python($$|[^0-9])~ s|/usr/bin/python|/usr/bin/python3|' '{}' \;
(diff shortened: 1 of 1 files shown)


Lex-2008:
test build: https://ci.cfengine.com/job/pr-pipeline/8633/

Lex-2008:
@craigcomstock was right: first commit is not good enough. This line:
```
test ... && test ... && find ...
```

returns non-zero exit status if any of `test` expressions return false - thus aborting the build. One solution is to add `|| true` at the end, another, as suggested in 2nd commit, to just drop `-mindepth` (we don't understand why it's there, anyway). We'll do that!

vpodzime (commented):
Please update the PR title accordingly. Also, the change is only OK now because the selection of the files is more careful after a recent change, right, @craigcomstock ?

Lex-2008:
yes, as @craigcomstock mentioned in PM on Slack,
> before I added the search for name `*.py and mock_*` it was running `sed -ri` on every file under that dir.

vpodzime:
> yes, as @craigcomstock mentioned in PM on Slack,

Please add this info to the commit message instead of *looks like it's not needed*.

Lex-2008:
Commit message amended! Can someone approve it, so we could merge it?

craigcomstock (approved):

PR #2509, opened 2022-09-26
Remove mindepth argument from find since not all platforms support it - 3.18.x

Lex-2008:
The change is OK now because the selection of the files is more careful
after a recent change, quoting @craigcomstock:
> before I added the search for name `*.py and mock_*` it was running
> `sed -ri` on every file under that dir.

Platforms which don't support it are AIX 5 and HP-UX.

(cherry-picked from commit https://github.com/Lex-2008/masterfiles/commit/8099f1a06e57ddae240a859dd5389c7736788897)

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -5,7 +5,7 @@ SUBDIRS += acceptance
 endif
 
 fix-python-hashbang:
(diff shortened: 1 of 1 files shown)


craigcomstock (changes requested):
@Lex-2008 please amend the commit comment and PR description to match the master PR. aka remove "doesn't look like" and such.

craigcomstock (approved):

PR #2515, opened 2022-10-05
CFE-4069: Added date to known paths for linux

nickanderson:
This change makes $(paths.date) available along with its associated classes when
the binary is present.

Ticket: CFE-4069
Changelog: Title

### DIFF
--- a/lib/paths.cf
+++ b/lib/paths.cf
@@ -138,6 +138,7 @@ bundle common paths
       "path[timedatectl]"    string => "/usr/bin/timedatectl";
 
     linux::
+      "path[date]"          string => "/usr/bin/date";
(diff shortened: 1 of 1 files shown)


lib/paths.cf
```diff
@@ -138,6 +138,7 @@ bundle common paths
       "path[timedatectl]"    string => "/usr/bin/timedatectl";
 
     linux::
+      "path[date]"          string => "/usr/bin/date";
```
craigcomstock, line 141: alpine which uses busybox has `/bin/date` but of course you can install a proper `date` in the right place most likely.

craigcomstock (approved):

PR #2525, opened 2022-10-24
ENT-9531: Added reports for enterprise users to migrate db retention settings

nickanderson:
This change is intended to ensure that customizations users have made related to
data retention are migrated to the new location where cf-reactor can make use of them.

Example output when retention settings have been modified:

    R: NOTICE: 'default:cfe_internal_hub_maintain.report_settings' has been customized. These settings have moved to Mission Portal as cf-reactor has taken over responsibility for maintaining this data.
    R: NOTICE: contexts history_length_days has been customized. Default: 7 Set: 71
    R: NOTICE: 'default:cfe_internal_hub_maintain.diagnostics_settings' has been customized. These settings have moved to Mission Portal as cf-reactor has taken over responsibility for maintaining this data.
    R: NOTICE: diagnostics history_length_days has been customized. Default: 1 Set: 21

Ticket: ENT-9531
Changelog: Title

### DIFF
--- a/lib/cfe_internal_hub.cf
+++ b/lib/cfe_internal_hub.cf
@@ -110,6 +110,108 @@ bundle agent cfe_internal_hub_maintain
       action => if_elapsed_day;
 }
 
+bundle agent cfe_internal_hub_maintain_notice_retention_settings_change
+# @brief Emits reports to inform users that configuration settings have been
+# modified and need to be migrated to new location.
+#
+# Database maintenance process of purging of old data is now the responsibility
+# of cf-remote. This bundle looks for user modifications of default retention
+# and informs they should migrate their configuration.
+{
+  classes:
+    "running_binary_with_new_db_retention_settings"
+      expression => "cfengine_3_22_0|feature_cf_reactor_report_retention";
+
+  vars:
+
+      # These variables were duplicated from cfe_internal_hub_maintain so that
+      # departure from defaults can be determined.
+
+      "default_report_settings"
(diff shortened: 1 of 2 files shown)


lib/cfe_internal_hub.cf
```diff
@@ -110,6 +110,108 @@ bundle agent cfe_internal_hub_maintain
...
+# @brief Emits reports to inform users that configuration settings have been
+# modified and need to be migrated to new location.
+#
+# Database maintenance process of purging of old data is now the responsibility
+# of cf-remote. This bundle looks for user modifications of default retention
+# and informs they should migrate their configuration.
+{
+  classes:
+    "running_binary_with_new_db_retention_settings"
+      expression => "cfengine_3_22_0|feature_cf_reactor_report_retention";
```
nickanderson, line 123: This was just a WAG for the context. Ideally, we will have a non version based class to key off of for this.

nickanderson:
Tagged WIP because this doesn't need to be prioritized for release and it may change based on other implementation details no yet determined.

lib/cfe_internal_hub.cf
vpodzime: Yeah, maybe a feature class? Or something similar.

vpodzime (approved):
ACK. We will see how things evolve before this is needed.

olehermanse (changes requested):
👍 , but don't merge until needed.

PR #2527, opened 2022-10-25
ENT-9535: Prevented management of runagent socket users when no users are listed

nickanderson:
This changes the guard for setting runagent socket ownership (which uses ACLs)
for the case when the list of runagent socket users is not an empty list. This
allows for disablement by simply setting an empty list via augments.

Ticket: ENT-9535
Changelog: Title

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -511,7 +511,7 @@ bundle common def
       "_have_control_agent_files_single_copy" -> { "CFE-3622"}
         expression => isvariable( "def.control_agent_files_single_copy" );
       "_have_control_executor_runagent_socket_allow_users"
-        expression => isvariable( "def.control_executor_runagent_socket_allow_users" );
+        expression => some( ".+", "def.control_executor_runagent_socket_allow_users" );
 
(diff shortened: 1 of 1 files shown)


craigcomstock (approved):

vpodzime (approved):
ACK. Good job with the investigation!

controls/def.cf
```diff
@@ -511,7 +511,7 @@ bundle common def
       "_have_control_agent_files_single_copy" -> { "CFE-3622"}
         expression => isvariable( "def.control_agent_files_single_copy" );
       "_have_control_executor_runagent_socket_allow_users"
-        expression => isvariable( "def.control_executor_runagent_socket_allow_users" );
+        expression => some( ".+.*", "def.control_executor_runagent_socket_allow_users" );
```
olehermanse, line 514: How is `.+.*` different from `.+`?

nickanderson: its just EXTRA more matching

nickanderson: Tidied

PR #2528, opened 2022-10-25
ENT-9535: Prevented management of runagent socket users when no users are listed (3.18)

nickanderson:
This changes the guard for setting runagent socket ownership (which uses ACLs)
for the case when the list of runagent socket users is not an empty list. This
allows for disablement by simply setting an empty list via augments.

Ticket: ENT-9535
Changelog: Title
(cherry picked from commit c8de0d254345c8bb37c22caacd13ae812c527c62)


cf-bottom:

PR #2531, opened 2022-11-04
ENT-9527: Added check to make sure cf-execd is running after attempting self upgrade on Windows (3.18)

nickanderson:
If the package upgrade fails, for example from trying to install a package that
was expected to be present but is missing, cf-execd will not be running, and the
host will not make further attempts to upgrade.

This change attempts to start cf-execd if it is not running after attempting the
package upgrade.

Ticket: ENT-9572
Changelog: Title
(cherry picked from commit af279aefc7c632bdcfd77400909538de1e3bd73a)


craigcomstock:
Ticket is actually ENT-9573

PR #2542, opened 2022-11-18
ENT-9600: Fixed case where cf-apache on Enterprise Hubs was not re-/started ?

nickanderson:
We have seen cases where httpd is not starting or being re-started as expected,
possibly as a result of function caching.

Ticket: ENT-9600
Changelog: Title

### DIFF
--- a/cfe_internal/enterprise/mission_portal.cf
+++ b/cfe_internal/enterprise/mission_portal.cf
@@ -165,6 +165,7 @@ bundle agent mission_portal_apache_from_stage(config, staged_config)
   classes:
       "systemd_supervised"
         expression => returnszero("$(paths.systemctl) -q is-enabled cf-apache > /dev/null 2>&1", "useshell"),
+        action => default:immediate,
         if => fileexists( $(paths.systemctl) );
 
   vars:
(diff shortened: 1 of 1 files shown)


craigcomstock:
Not sure if this PR helps, but with these changes PLUS removing `SSLCompression off` from the httpd.conf template makes things work on ubuntu-16! :)

```
  # TLS Compression should be disabled to avoid CRIME
  # https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2012-4929
# TODO: including this when our ssl library doesn't support??? breaks the config (ENT-9600 ubuntu-16 centos-7)
#  SSLCompression off
```

Not sure where we should go with this change. I added more specific notes in the ticket:

```
root@ubuntu1604:~# /var/cfengine/httpd/bin/httpd -t -f /var/cfengine/httpd/conf/httpd.conf.staged
AH00526: Syntax error on line 150 of /var/cfengine/httpd/conf/httpd.conf.staged:
Setting Compression mode unsupported; not implemented by the SSL library
root@ubuntu1604:~# sed -n '150p' /var/cfengine/httpd/conf/httpd.conf.staged
  SSLCompression off
```

nickanderson:
Perfect I think a better change is to adjust the https template accordingly. Plausibly that alone will resolve the issue.

nickanderson:
@craigcomstock The PR to supersede this one after our further discussion is https://github.com/cfengine/masterfiles/pull/2545

PR #2544, opened 2022-11-21
ENT-9621: Skip great package test on RHEL-9

craigcomstock:
Has issues with rpm -U --force behavior being
different than previous releases. As well as
possibly the same issues as rhel-8 had when we
chose to skip this test on rhel-8.

Fix this in CFE-4096

Ticket: ENT-9621
Changelog: none
(cherry picked from commit ffc6d446821fce73950f49bdc8ce501678ef2c25)


Lex-2008 (approved):

vpodzime (approved):
ACK

PR #2550, opened 2022-12-13
Use absolute path to httpd.pid

Lex-2008:
Otherwise cf-agent might not find it

Ticket: ENT-9603

### DIFF
--- a/cfe_internal/enterprise/mission_portal.cf
+++ b/cfe_internal/enterprise/mission_portal.cf
@@ -176,10 +176,10 @@ bundle agent mission_portal_apache_from_stage(config, staged_config)
       # httpd/httpd.pid in 3.15.5, 3.18.1 and, 3.19.0
 
       "httpd_pid_file" -> { "ENT-7966" }
-        string => ifelse( classmatch( "cfengine_3_1[0-4]" ), "httpd/logs/httpd.pid",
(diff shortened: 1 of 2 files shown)


cfe_internal/enterprise/mission_portal.cf
```diff
@@ -176,10 +176,10 @@ bundle agent mission_portal_apache_from_stage(config, staged_config)
       # httpd/httpd.pid in 3.15.5, 3.18.1 and, 3.19.0
 
       "httpd_pid_file" -> { "ENT-7966" }
-        string => ifelse( classmatch( "cfengine_3_1[0-4]" ), "httpd/logs/httpd.pid",
-                          classmatch( "cfengine_3_15_[0-4]" ), "httpd/logs/httpd.pid",
-                          "cfengine_3_18_0", "httpd/logs/httpd.pid",
-                          "httpd/httpd.pid" );
+        string => ifelse( classmatch( "cfengine_3_1[0-4]" ), "$(sys.workdir)/httpd/logs/httpd.pid",
```
nickanderson, line 179: Makes sense. No way that `fileexists()` would be able to know where to look with an unqualified path. But I think that this should come with a template change.

nickanderson (commented):
LGTM.

nickanderson (approved):

PR #2560, opened 2022-12-19
ENT-9686: Ensured manual edits to httpd.conf are reverted

nickanderson:
This change ensures that a manual change to httpd.conf is reverted as long as
the staged template render is still valid and that httpd gets re-started as
necessary to ensure it's running with the expected configuration.

Ticket: ENT-9686
Changelog: Title


Lex-2008 (approved):

cfe_internal/enterprise/mission_portal.cf
```diff
@@ -236,6 +247,13 @@ bundle agent mission_portal_apache_from_stage(config, staged_config)
         comment => concat( "We start apache after the new valid config is in ",
                            "place only if we have stopped apache already.");
 
+      "cf-apache" -> { "ENT-9686" }
+        service_policy => "restart",
+        if => and( "mission_portal_apache_config_repaired",
+                   not( "apache_stop_after_new_staged_config_repaired" ) ),
+        comment => concat( "We start apache after the new valid config is in ",
```
craigcomstock, line 254:
```suggestion
        comment => concat( "We restart apache after the new valid config is in ",
```

cfe_internal/enterprise/mission_portal.cf
```diff
@@ -236,6 +247,13 @@ bundle agent mission_portal_apache_from_stage(config, staged_config)
         comment => concat( "We start apache after the new valid config is in ",
                            "place only if we have stopped apache already.");
 
+      "cf-apache" -> { "ENT-9686" }
+        service_policy => "restart",
+        if => and( "mission_portal_apache_config_repaired",
+                   not( "apache_stop_after_new_staged_config_repaired" ) ),
+        comment => concat( "We restart apache after the new valid config is in ",
+                           "place only if we have stopped apache already.");
```
craigcomstock: Is this part of the comment accurate? If so, why not also on the non-systemd version?

craigcomstock: Yeah, I don't think this part of the comment is accurate.  In fact, it's the opposite, we restart apache after the new valid config is in place only if we have NOT stopped apache already. Right? Maybe add that here and above as well?

craigcomstock (approved):

PR #2564, opened 2022-12-20
ENT-9711: Updated reference to explicitly create runalerts-stamp directory

nickanderson:
Without /. a file will be created. Without create => true no action is promised.

Ticket: ENT-9711
Changelog: None

together:
https://github.com/cfengine/masterfiles/pull/2564
https://github.com/cfengine/core/pull/5136

### DIFF
--- a/templates/cf-runalerts.service.mustache
+++ b/templates/cf-runalerts.service.mustache
@@ -3,6 +3,7 @@ Description=CFEngine Enterprise SQL Alerts
 After=syslog.target
 ConditionPathExists={{{vars.sys.bindir}}}/runalerts.php
 ConditionFileIsExecutable={{{vars.sys.workdir}}}/httpd/php/bin/php
+ConditionPathIsDirectory={{{vars.sys.workdir}}}/httpd/php/runalerts-stamp
 
 PartOf=cfengine3.service
 After=cf-postgres.service
--- a/cfe_internal/enterprise/CFE_knowledge.cf
+++ b/cfe_internal/enterprise/CFE_knowledge.cf
@@ -56,12 +56,14 @@ bundle agent cfe_internal_setup_knowledge
       handle => "cfe_internal_setup_knowledge_dir_doc_root",
       perms => mog("0550", "root", $(def.cf_apache_group) );
 
-      "$(sys.workdir)/httpd/php" -> { "ENT-9703", "cfe_internal_setup_knowledge_dir_doc_root_runalerts_stamp" }
+      "$(sys.workdir)/httpd/php/." -> { "ENT-9703", "cfe_internal_setup_knowledge_dir_doc_root_runalerts_stamp" }
+        create => "true",
         comment => "php directory needs to be accessible (execute) to cfapache",
         handle => "cfe_internal_setup_knowledge_dir_doc_root_php",
         perms => mog("0550", "root", $(def.cf_apache_group) );
 
-      "$(sys.workdir)/httpd/php/runalerts-stamp" -> { "ENT-9703" }
+      "$(sys.workdir)/httpd/php/runalerts-stamp/." -> { "ENT-9703" }
+        create => "true",
         comment => "runalerts-stamp directory needs to be accessible (execute) to cfapache",
         handle => "cfe_internal_setup_knowledge_dir_doc_root_runalerts_stamp",
         perms => mog("0550", "root", $(def.cf_apache_group) );


templates/cf-runalerts.service.mustache
```diff
@@ -3,6 +3,7 @@ Description=CFEngine Enterprise SQL Alerts
 After=syslog.target
 ConditionPathExists={{{vars.sys.bindir}}}/runalerts.php
 ConditionFileIsExecutable={{{vars.sys.workdir}}}/httpd/php/bin/php
+ConditionPathIsDirectory={{{vars.sys.workdir}}}/httpd/php/runalerts-stamp
```
craigcomstock, line 6: Seems like we would want this service to fail with an error, not just "silently" as I understand from https://www.man7.org/linux/man-pages/man5/systemd.unit.5.html

```suggestion
AssertPathExists={{{vars.sys.bindir}}}/runalerts.php
AssertFileIsExecutable={{{vars.sys.workdir}}}/httpd/php/bin/php
AssertPathIsDirectory={{{vars.sys.workdir}}}/httpd/php/runalerts-stamp
```

vpodzime (approved):
ACK

templates/cf-runalerts.service.mustache
craigcomstock: I'll create a ticket. We can do this after release.

craigcomstock (approved):

olehermanse (approved):

craigcomstock:
Tested this change on an ubuntu 22 3.21.0 hub. A bit tricky to test actually as I had to enable unit file management and using host specific data needed to use the default namespace when enabling the class:

https://docs.cfengine.com/docs/3.21/reference-masterfiles-policy-framework.html#enable-management-of-components-on-systemd-hosts

`default:mpf_enable_cfengine_systemd_component_management`

Then the unit is rendered into `/etc/systemd/system` and if I remove the directory I see that the service is not started and an error is in status, but not in journal/syslog:

```
root@ubuntu2204:/etc/systemd/system# systemctl status cf-runalerts.service 
○ cf-runalerts.service - CFEngine Enterprise SQL Alerts                                                                                                
     Loaded: loaded (/etc/systemd/system/cf-runalerts.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Tue 2023-01-03 17:25:33 UTC; 1min 1s ago               
  Condition: start condition failed at Tue 2023-01-03 17:26:28 UTC; 5s ago
             └─ ConditionPathIsDirectory=/var/cfengine/httpd/php/runalerts-stamp was not met
    Process: 20913 ExecStart=/var/cfengine/httpd/php/bin/php /var/cfengine/bin/runalerts.php (code=killed, signal=TERM)
   Main PID: 20913 (code=killed, signal=TERM)                                                                                                          
        CPU: 523ms    
```

So it works but doesn't complain very loudly.

nickanderson:
> So it works but doesn't complain very loudly.

What doesn't complain loudly? Systemd?

nickanderson:
And, shall we proceed to merge this then?

craigcomstock:
> > So it works but doesn't complain very loudly.
> 
> What doesn't complain loudly? Systemd?

If we changed them to Asserts I think they would be more useful.

https://www.freedesktop.org/software/systemd/man/systemd.unit.html

> Unit files may also include a number of Condition…= and Assert…= settings. Before the unit is started, systemd will verify that the specified conditions and asserts are true. If not, the starting of the unit will be (mostly silently) skipped (in case of conditions), or aborted with an error message (in case of asserts). 

Changing to Asserts I get what I think is more useful/informative. It does return an exit code of  `1` here as well so would likely cause noise in the policy run. :+1: 

```
root@ubuntu2204:/var/log# rm -rf /var/cfengine/httpd/php/runalerts-stamp/
root@ubuntu2204:/var/log# systemctl restart cf-runalerts.service
Assertion failed on job for cf-runalerts.service.
root@ubuntu2204:/var/log# systemctl status cf-runalerts.service
○ cf-runalerts.service - CFEngine Enterprise SQL Alerts
     Loaded: loaded (/etc/systemd/system/cf-runalerts.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Tue 2023-01-03 17:36:55 UTC; 13s ago
     Assert: start assertion failed at Tue 2023-01-03 17:36:55 UTC; 13s ago
             AssertPathIsDirectory=/var/cfengine/httpd/php/runalerts-stamp was not met
    Process: 24873 ExecStart=/var/cfengine/httpd/php/bin/php /var/cfengine/bin/runalerts.php (code=killed, signal=TERM)
   Main PID: 24873 (code=killed, signal=TERM)
        CPU: 78ms

Jan 03 17:36:48 ubuntu2204.localdomain systemd[1]: Started CFEngine Enterprise SQL Alerts.
Jan 03 17:36:48 ubuntu2204.localdomain php[24873]: # Rendered by CFEngine
Jan 03 17:36:55 ubuntu2204.localdomain systemd[1]: Stopping CFEngine Enterprise SQL Alerts...
Jan 03 17:36:55 ubuntu2204.localdomain systemd[1]: cf-runalerts.service: Deactivated successfully.
Jan 03 17:36:55 ubuntu2204.localdomain systemd[1]: Stopped CFEngine Enterprise SQL Alerts.
Jan 03 17:36:55 ubuntu2204.localdomain systemd[1]: cf-runalerts.service: Starting requested but asserts failed.
Jan 03 17:36:55 ubuntu2204.localdomain systemd[1]: Assertion failed for CFEngine Enterprise SQL Alerts.
```

craigcomstock:
> And, shall we proceed to merge this then?

Yes. Let's merge! I created a ticket to investigate/change Conditions to Asserts: https://tracker.mender.io/browse/ENT-9743

craigcomstock:
cherry picked to 3.18.x https://github.com/cfengine/masterfiles/pull/2569

PR #2571, opened 2023-01-17
ENT-9778: Fixed body perms system_owned to account for Windows

nickanderson:
This change accounts for windows in the system_owned perms body.

Note: There is a limitation not being able to set file ownership to users other
than the executing user, so this would be expected to generate errors if
cf-agent is not run by Administrator.

Ticket: ENT-9778
Changelog: Title

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -1738,7 +1738,6 @@ body perms system_owned(mode)
 # ```
 {
       mode   => "$(mode)";
-      owners => { "root" };
 
 #+begin_ENT-951
 # Remove after 3.20 is not supported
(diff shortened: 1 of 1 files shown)


nickanderson:
@olehermanse, @craigcomstock 

What do you think about this. I am slightly on the fence about setting `owners` to `Administrator`. If a user that is not `Administrator` is executing `cf-agent` there will be errors produced. But, that seems correct, at least for `body perms system_owned`. We have a couple of other tickets to further investigate the limitation of not being able to set ownership to a user that is not the currently executing user.

nickanderson:
> 

Actually, users have now reported that both `Administrator` and `SYSTEM` result in the same errors when setting `owners`, thus I have made a change to omit setting `owners` for `windows`.

nickanderson:
User reports no errors with this current PR.

craigcomstock (approved):
looks good!

nickanderson:
@olehermanse what do you think? Shall I merge it like this with owners explicitly excluded since it's an improvement on the current state or shall I wait for further investigation and potential work in related tickets?

olehermanse (approved):

PR #2584, opened 2023-01-24
ENT-9806: Added Inventory for Policy Version

nickanderson:
This change adds inventory for the policy version as defined in body common
control. It's helpful to be able to easily look up what policy version something
is running (assuming a user has not modified the value set during release).

Ticket: ENT-9806
Changelog: Title


nickanderson:
image

craigcomstock (approved):

PR #2591, opened 2023-02-03
CFE-4146: Avoided deleting python symlink when sys.bindir is not /var/cfengine/bin

nickanderson:
If you build CFEngine with --enable-fhs then sys.bindir is going to be /usr/bin
and cleaning up the cfengine selected python symlink that could have been left
would result in cleaning up the system python symlink instead. This change
scopes the cleanup down to only when sys.bindir is /var/cfengine/bin.

Ticket: CFE-4146
Changelog: Title

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -139,13 +139,14 @@ bundle agent cfe_internal_setup_python_symlink(symlink_path)
         move_obstructions => "true",
         if => isvariable("python");
 
-      "$(sys.bindir)/python" -> { "CFE-3512" }
+      "$(sys.bindir)/python" -> { "CFE-3512", "CFE-4146" }
         delete => u_tidy,
-        if => islink( "$(sys.bindir)/python" ),
+        if => and( islink( "$(sys.bindir)/python" ),
+                   strcmp( "$(sys.bindir)", "/var/cfengine/bin")),
         comment => concat( "We don't want to leave a python that is potentially in $PATH ",
                            "after having re-named our python symlink that is used for various ",
-                           "modules");
-
+                           "modules. Additionally we want to be cautious that we don't delete ",
+                           "system python symlinks in the event the binary was built for FHS.");
 }
 
 bundle agent cfe_internal_update_policy_cpv


moschlar:
LGTM

nickanderson:
There is a naive pr to core trying to add a hard class when fhs is enabled: https://github.com/cfengine/core/pull/5176

This pr should be amended to OR the case of the class  if that pr is liked.

cfe_internal/update/update_policy.cf
```diff
@@ -139,13 +139,14 @@ bundle agent cfe_internal_setup_python_symlink(symlink_path)
...
         delete => u_tidy,
-        if => islink( "$(sys.bindir)/python" ),
+        if => and( islink( "$(sys.bindir)/python" ),
+                   strcmp( "$(sys.bindir)", "/var/cfengine/bin")),
         comment => concat( "We don't want to leave a python that is potentially in $PATH ",
                            "after having re-named our python symlink that is used for various ",
-                           "modules");
-
+                           "modules. Additionally we want to be cautious that we don't delete ",
+                           "system python symlinks in the event the binary was built for FHS.");
```
craigcomstock, line 149: I know what FHS is, but wonder if we don't need a glossary somewhere or just replace `FHS` with `https://wiki.linuxfoundation.org/lsb/fhs`.

craigcomstock (approved):

olehermanse (approved):

nickanderson:
This PR is waiting on https://github.com/cfengine/core/pull/5176

vali-um:
this would be highly appreciated.
alternatively you could never delete the symlink as it might interfere with system packages that set this symlink (eg. python-is-python3)

craigcomstock:
> this would be highly appreciated. alternatively you could never delete the symlink as it might interfere with system packages that set this symlink (eg. python-is-python3)

Thanks for the comment. As I understand the policy now it will only delete a symlink in `/var/cfengine/bin` so quite safe I would say. Should never fiddle with any system packages.

vali-um:
unfortunately, this is not true as (at least) debian and ubuntu build cfengine with --enable-fhs. so sys.bindir points to /usr/bin.
and the symlink /usr/bin/python is exatly what python-is-python3 uses

cfe_internal/update/update_policy.cf
nickanderson: We have https://docs.cfengine.com/docs/3.21/resources-faq-fhs.html

You think we should add the term here? https://docs.cfengine.com/docs/3.21/resources-additional-topics-glossary.html#top

nickanderson:
> unfortunately, this is not true as (at least) debian and ubuntu build cfengine with --enable-fhs. so sys.bindir points to /usr/bin. and the symlink /usr/bin/python is exatly what python-is-python3 uses

Yeah, I think @craigcomstock meant that "now" as if this PR is merged, which I am about to do. Will be great if you can report back here and or in the ticket https://northerntech.atlassian.net/browse/CFE-4146 if it improves the behavior for you.

nickanderson:
> There is a naive pr to core trying to add a hard class when fhs is enabled: [cfengine/core#5176](https://github.com/cfengine/core/pull/5176)
> 
> This pr should be amended to OR the case of the class if that pr is liked.

Decided to not wait for a change to core.

PR #2604, opened 2023-05-01
Simplified github workflows conditions, removed branches

craigcomstock:
If we specify branches it doesn't mean that branch PRs will pay attention to the yaml that is committed to master.

So we would have to cherry-pick back all the time.

So just operate on pull_request and push always, regardless of branch.

Ticket: none
Changelog: none


craigcomstock:
less change is better? let this one go :)

PR #2623, opened 2023-05-11
ENT-10212: Made allowconnects and allowallconnects configurable via Augments

nickanderson:


MPF.md
```diff
@@ -643,6 +643,59 @@ Here is an example setting the acl from augments:
 
 **See Also:** [Configure networks allowed to make collect calls (client initiated reporting)](#configure-networks-allowed-to-make-collect_calls-client-initiated-reporting)
 
+### Configure hosts that may connect to cf-serverd
+
+`allowconnects` is a list of IP addresses or subnets in `body server control` restricts hosts that are allowed to connect to `cf-serverd`. This is the first layer of access control in `cf-serverd`, a client coming from a host not covered by this list will not be able to connect to `cf-serverd` at all.
```
craigcomstock, line 648:
```suggestion
`allowconnects` is a list of IP addresses or subnets in `body server control` which restricts hosts that are allowed to connect to `cf-serverd`. This is the first layer of access control in `cf-serverd`, a client coming from a host not covered by this list will not be able to connect to `cf-serverd` at all.
```

MPF.md
```diff
@@ -643,6 +643,59 @@ Here is an example setting the acl from augments:
...
+
+In the MPF this defaults to include localhost and the value defined for `default:def.acl`.
+
+`allowconnects` can be customized by configuring `default:def.control_server_allowconnects` via Augments.
```
craigcomstock, line 652: so if you define this `allowconnects` does it clobber the previous "localhost and default:def.acl" or add to it?

MPF.md
```diff
@@ -643,6 +643,59 @@ Here is an example setting the acl from augments:
...
+}
+```
+
+**History:**
+
+* Added in 3.22.0
+
+### Configure hosts that may make multiple concurrent connections to cf-serverd
+
+`allowallconnects` is a list of IP addresses or subnets in `body server control` specifying hosts that are allowed to have more than one connection to `cf-serverd`.
```
craigcomstock, line 674: What happens to a host that is NOT allowed "more than one connection"? Do they only get one connection and so... what's the impact? They fail like during `update.cf` or just go slower or ???

MPF.md
craigcomstock: ah, now I see the code below, it "overwrites" you might mention that here in the docs.

craigcomstock (commented):
cool! some questions/comments :+1:

MPF.md
nickanderson: ACK

MPF.md
nickanderson: Yeah, I was a bit on the fence about extend vs overwrite when I started. I asked for opinions there but I didn't wait long before I made a choice and proceeded:

With full definition you can achieve extension by re-defining the defaults. With extension you can't really influence the defaults that include localhost and def.acl, so I guess full definition seems the best choice.

Adding a note as suggested.

MPF.md
nickanderson: Well, it depends. Most of cf-agent is serial, so only one connection at a time is typically done. But, if you have concurrent agents that are all trying to get something from the server there could be denials. If you `background` some files promise then you could get into concurrent copies. Also, if you are using client initiated reporting, a cf-agent process might tie up that single slot and then prevent a periodic client initiated report to connect and offer opportunity for collection.

nickanderson:
@craigcomstock Updates made, please re-review.

craigcomstock (approved):

PR #2627, opened 2023-05-16
ENT-10214: Improved federated reporting dump concurrency with database

craigcomstock:
It was found that with larger numbers of hosts and inventory data even with improvements to update_inventory() procedure in cfdb schema (change in nova) concurrency problems could still occur.

--serializable-deferrable option seems to help as well to cause pg_dump to wait for a consistent state in the database before proceeding with the dump.

Ticket: ENT-10214
Changelog: title

### DIFF
--- a/templates/federated_reporting/dump.sh
+++ b/templates/federated_reporting/dump.sh
@@ -55,7 +55,7 @@ in_progress_file="$CFE_FR_DUMP_DIR/$CFE_FR_FEEDER_$ts.sql.$CFE_FR_COMPRESSOR_EXT
 
 log "Dumping tables: $CFE_FR_TABLES"
 {
-  "$CFE_BIN_DIR"/pg_dump --column-inserts --data-only $(printf ' -t %s' $CFE_FR_TABLES) cfdb
+  "$CFE_BIN_DIR"/pg_dump --serializable-deferrable --column-inserts --data-only $(printf ' -t %s' $CFE_FR_TABLES) cfdb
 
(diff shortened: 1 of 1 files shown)


craigcomstock:
see https://www.postgresql.org/docs/current/app-pgdump.html

--serializable-deferrable

> Use a serializable transaction for the dump, to ensure that the snapshot used is consistent with later database states; but do this by waiting for a point in the transaction stream at which no anomalies can be present, so that there isn't a risk of the dump failing or causing other transactions to roll back with a serialization_failure. See [Chapter 13](https://www.postgresql.org/docs/current/mvcc.html) for more information about transaction isolation and concurrency control.
> 
> This option is not beneficial for a dump which is intended only for disaster recovery. It could be useful for a dump used to load a copy of the database for reporting or other read-only load sharing while the original database continues to be updated. Without it the dump may reflect a state which is not consistent with any serial execution of the transactions eventually committed. For example, if batch processing techniques are used, a batch may show as closed in the dump without all of the items which are in the batch appearing.
> 
> This option will make no difference if there are no read-write transactions active when pg_dump is started. If read-write transactions are active, the start of the dump may be delayed for an indeterminate length of time. Once running, performance with or without the switch is the same.

nickanderson (approved):
Makes sense to me based on the documentation snippet you posted, also I saw that it seems to work in the real world.

vpodzime (approved):
Sounds like a perfect match!

PR #2634, opened 2023-05-19
prepare.sh: Enabled setting release and version via env vars

olehermanse:
This makes it easier to "fix" masterfiles modules with wrong
version numbers. Can for example happen if there is something
non-standard with tags or .CFVERSION files. In such cases,
these env vars can be set from CFEngine Build steps, when
running the prepare.sh script, to force the correct version number.

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -14,17 +14,38 @@ fi
 
 cd "$(dirname $0)"
 
-echo "Running determine-version.sh ..."
-rm -f CFVERSION
-misc/determine-version.sh .CFVERSION > CFVERSION \
-    || echo "Unable to auto-detect CFEngine version, continuing"
+# Prefer EXPLICIT_VERSION variable over running the shell script,
+# this means you can set it explicitly, for example from CFEngine Build steps:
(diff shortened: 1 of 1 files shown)


nickanderson (approved):

prepare.sh
```diff
@@ -14,19 +14,40 @@ fi
...
+# CFVERSION file may look like this: 3.18.4-2 and this matches a tag
+# However, for the version variable here, used to put into policy files
+# we don't want the -2 part. That part should go into the other variable
+# called release:
+version=$(cat CFVERSION | awk -F"-" '{print $1}')
+# The code above should already have parsed tags / env var and put the correct
+# thing or default in CFRELEASE, no need to do more awk here:
 release=$(cat CFRELEASE)
-prefix="/var/cfengine/"
+prefix="/var/cfengine"
```
nickanderson, line 50: yes.

nickanderson (approved):

craigcomstock (approved):

vpodzime (approved):
ACK

Lex-2008 (approved):

PR #2651, opened 2023-06-12
ENT-10345: Aligned masterfiles module build steps with release tarball

nickanderson:
These changes align the result from cfbs build with the masterfiles package
tarball artifact built during release.

Ticket: ENT-10345
Changelog: Title

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -90,6 +90,16 @@ test-driver
 tests
 travis-scripts
 .travis.yml
+cfe_internal/core/watchdog/README.md
+cfe_internal/enterprise/ha/ha_info.json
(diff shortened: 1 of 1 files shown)


olehermanse (changes requested):
I think these should be a part of prepare.sh.

nickanderson:
> I think these should be a part of prepare.sh.

Moved.

nickanderson:
Removed cherry-pick label, cherry picks are up.

- 3.21 https://github.com/cfengine/masterfiles/pull/2652
- 3.18 https://github.com/cfengine/masterfiles/pull/2653

PR #2665, opened 2023-06-27
CFE-4223: do not filter $(sys.bindir) from python path

vali-um:
As $(sys.bindir) might point to /usr/bin filtering it can lead to missing an available python executeable.
The orginal reason to do so was to avoid a self referential link (https://github.com/cfengine/masterfiles/commit/d37c6d1929913d8cb9b269ce02ec823ade29b440)
Since the symlink is now named `cfengine-selected-python` (https://github.com/cfengine/masterfiles/commit/a864d5cced81b37cf0f0ed9e61d392bd6ed96194) it is not matched by the regex anymore and therefor not an issue.

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -90,9 +90,7 @@ bundle agent cfe_internal_setup_python_symlink(symlink_path)
 {
   vars:
       "path" string => getenv("PATH", 1024);
-      "path_folders" slist => filter("$(sys.bindir)",
-                                     splitstring("$(path)", ":", 128),
-                                     false, true, 128);
+      "path_folders" slist => splitstring("$(path)", ":", 128);
 
     windows::
       "abs_path_folders" -> {"CFE-2309"}


cfe_internal/update/update_policy.cf
```diff
@@ -90,9 +90,7 @@ bundle agent cfe_internal_setup_python_symlink(symlink_path)
 {
   vars:
       "path" string => getenv("PATH", 1024);
-      "path_folders" slist => filter("$(sys.bindir)",
-                                     splitstring("$(path)", ":", 128),
-                                     false, true, 128);
+      "path_folders" slist => splitstring("$(path)", ":", 128);
```
nickanderson, line 93: @vali-um there is some history behind this.

The symlink to the version of python to use was originally called `python`.

https://github.com/cfengine/masterfiles/commit/7e8fb18f4d0a31499d4b9feb5828c18e028001db#diff-0db031f8fa35f6bee0b584e73e273f3abbd8134a5958e12694209f52ca8da693R194

This caused issues for some people when `/var/cfengine/bin` was in PATH and so `python` might have an unexpected target.

We renamed this symlink to be more explicit, and as part of that we want to make sure we don't have the `python` symlink in `/var/cfengine/bin`.

https://github.com/cfengine/masterfiles/commit/a864d5cced81b37cf0f0ed9e61d392bd6ed96194

Do you think this is still needed now that https://github.com/cfengine/masterfiles/pull/2591 has been merged?

nickanderson (changes requested):

cfe_internal/update/update_policy.cf
vali-um: I forgot to write down my conclusion yesterday:

Yes, i think the change is still needed and has, as far as i can tell, no undesired effects

nickanderson (approved):

nickanderson:
Thanks @vali-um , merged, backported to 3.18, 3.21

PR #2699, opened 2023-08-18
ENT-10480: Added ability to extend without overriding filename patterns to copy during policy update

nickanderson:
This change adds the ability to extend the default list of filename patterns to
copy during policy update. Prior to this you could only re-define the list of
filename patterns which adds a maintenance burden of tracking the upstream
defaults through policy framework upgrades.


craigcomstock (approved):
absolutely! very good! 🎉 🏆

PR #2700, opened 2023-08-18
ENT-10481: Added ability to enable backup archives during policy update

nickanderson:
By default CFEngine creates backups of files when they are modified. These
backups are created by default in the same directory as the file that changed.
Files in $(sys.inputdir) are affected by this as well, but the default policy
update and sync settings result in these backups being culled since they do not
exist at the distribution point.

This change adds the capability to enable backing up files that change in
$(sys.inputdir) to a single location.

Note: A similar setting already exists that applies to promises.cf (normal agent
runs). But it seems undesirable that the same configuration applies to the
update policy, resulting in backups of all changes to policy locally on the
host, hence the introduction of this new configuration option.

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1691,7 +1691,49 @@ For example:
 }
 ```
 
-**History**: Added in 3.10.1
+**Notes:**
+
+- This applies to `promises.cf`.
+
+
+**History:**
+
+- Introduced in CFEngine 3.10.1
+
+### Configure default repository for file backups during policy update
+
+By default the agent creates a backup of a file before it is edited in the same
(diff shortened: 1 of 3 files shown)


nickanderson:
> Yeah, I don't love the term "default repository" but I see it is already "baked in" so pretty much have to keep it.
> 
> https://docs.cfengine.com/docs/3.21/reference-components-cf-agent.html#default_repository

I agree, it was annoying while I was working on it. Perhaps we could add an alias that we prefer. backup_dir or something. I created a ticket in case someone gets inspired. https://northerntech.atlassian.net/browse/CFE-4237

PR #2713, opened 2023-08-31
CFE-4245: Fixed lib/testing.cf tap output

craigcomstock:
Though one issue remains: testing_todo() and testing_skip()
don't quite act correctly. They should report ok|not ok followed
by # SKIP or # TODO at the end of the test point line.

https://testanything.org/tap-version-14-specification.html

see https://northerntech.atlassian.net/browse/CFE-4246

Ticket: CFE-4245
Changelog: title

### DIFF
--- a/lib/testing.cf
+++ b/lib/testing.cf
@@ -154,7 +154,7 @@ bundle agent testing_generic_report(format, outfile)
       "tests_passed" data => '[]';
       "tests_passed"
         data => mergedata(tests_passed,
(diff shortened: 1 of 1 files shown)


nickanderson (approved):
Looks ok to me. Perhaps as you mentioned in chat, we should include an example TAP output.

Anything we add to the policy file can be rendered directly in the docs: https://docs.cfengine.com/docs/master/reference-masterfiles-policy-framework-lib-testing.html

craigcomstock:
cherry picked
https://github.com/cfengine/masterfiles/pull/2714
https://github.com/cfengine/masterfiles/pull/2715

PR #2717, opened 2023-09-01
ENT-9931: Guard against /sys/hypervisor/uuid not being readable

craigcomstock:
On a centos-6 PV instance in Amazon EC2 I see /sys/hypervisor/uuid exists

-r--r--r-- 1 root root 4096 Sep  1 14:30 /sys/hypervisor/uuid

but is not readable even though permissions seem to be ok.

cat: /sys/hypervisor/uuid: No such file or directory

While SELinux is enforced it doesn't seem to be involved since this is a special sysfs filesystem and there are no entries for the failure in the audit log.

Ticket: ENT-9931
Changelog: title

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -617,6 +617,15 @@ bundle common cfe_autorun_inventory_aws
         scope => "namespace",
         if => isvariable("cfe_autorun_inventory_dmidecode.dmi[bios-vendor]");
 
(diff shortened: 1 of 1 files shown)


larsewi (approved):
🚀

nickanderson (approved):

PR #2721, opened 2023-09-06
ENT-9931: Redirected all output when checking /sys/hypervisor/uuid with cat

craigcomstock:
This will only occur in pre 3.22.0 versions of CFEngine.

In 3.22.0 and after the isreadable() function is used.

Problem found with our no-noise-test showing the output of the uuid.

Ticket: ENT-9931
Changelog: none


nickanderson (approved):

craigcomstock:
cherry picks
https://github.com/cfengine/masterfiles/pull/2724
https://github.com/cfengine/masterfiles/pull/2723

PR #2722, opened 2023-09-06
ENT-10623: Marked 30_custom_promise_types/24_expireafter as flakey fail for aarch64 debian_11

craigcomstock:
Ticket: ENT-10623
Changelog: none


nickanderson (approved):
I have some memory that there was an issue using /multiple/ of these things together, but I can't find reference to that in https://github.com/cfengine/core/blob/master/tests/acceptance/README so it may simply be a hallucination.

PR #2725, opened 2023-09-07
ENT-9931: Fixed /sys/hypervisor/uuid is readable test for pre 3.22.0 versions

craigcomstock:
The 2>&1 and >/dev/null were in the wrong order.

Ticket: ENT-9931
Changelog: none

### DIFF
--- a/inventory/any.cf
+++ b/inventory/any.cf
@@ -622,7 +622,7 @@ bundle common cfe_autorun_inventory_aws
         expression => isreadable("/sys/hypervisor/uuid", 1);
 @else
       "sys_hypervisor_uuid_readable" -> { "ENT-9931" }
-        expression => returnszero("${paths.cat} /sys/hypervisor/uuid 2>&1 >/dev/null", "useshell");
+        expression => returnszero("${paths.cat} /sys/hypervisor/uuid >/dev/null 2>&1", "useshell");
 @endif
 
     !disable_inventory_aws.sys_hypervisor_uuid_readable::


craigcomstock:
This fix doesn't really affect master since it is 3.23 and there is a macro for post-3.22 to use isreadable() function instead of `cat`

cherry picks
https://github.com/cfengine/masterfiles/pull/2727
https://github.com/cfengine/masterfiles/pull/2726

inventory/any.cf
```diff
@@ -622,7 +622,7 @@ bundle common cfe_autorun_inventory_aws
         expression => isreadable("/sys/hypervisor/uuid", 1);
 @else
       "sys_hypervisor_uuid_readable" -> { "ENT-9931" }
-        expression => returnszero("${paths.cat} /sys/hypervisor/uuid 2>&1 >/dev/null", "useshell");
+        expression => returnszero("${paths.cat} /sys/hypervisor/uuid >/dev/null 2>&1", "useshell");
```
nickanderson, line 625: I have no opinion really, but you prefer no space after the first `>` ?

nickanderson: Also, `${}` is a bit a-typical for MPF which usually uses `$()`.

nickanderson (approved):
A couple comments, but :ok:

inventory/any.cf
nickanderson: I think that CFEngine would skip the promise that contained an unresolved variable.

craigcomstock: indeed it would!

PR #2732, opened 2023-09-19
ENT-10656: Added missing build steps to repository module to align with CFEngine Build index

nickanderson:
Now if you cfbs add https://github.com/cfengine/masterfiles you will get the
same build steps as those used when you cfbs add masterfiles from the CFEngine
Build index.

Ticket: ENT-10656
Changelog: Title

### DIFF
--- a/cfbs.json
+++ b/cfbs.json
@@ -8,7 +8,18 @@
       "tags": ["supported", "base"],
       "repo": "https://github.com/cfengine/masterfiles",
       "by": "https://github.com/cfengine",
-      "steps": ["run ./prepare.sh -y", "copy ./ ./"]
+      "steps": ["run ./prepare.sh -y",
+                "delete cfe_internal/core/watchdog/README.md",
+                "delete cfe_internal/enterprise/ha/ha_info.json",
+                "delete .github",
+                "delete inventory/README.md",
+                "delete lib/README.md",
+                "delete LICENSE",
+                "delete modules/promises",
+                "delete .no-distrib",
+                "delete services/autorun/README.md",
+                "delete templates/README.md",
+                "copy ./ ./"]
     }
   }
 }


nickanderson:
@olehermanse I am not sure what to set `EXPLICIT_RELEASE` and `EXPLICIT_VERSION` to here.

It influences not only the version string for the policy itself (`version`) in `body common control` of `promises.cf` and `update.cf` but also the default binary version of CFEngine to target during self upgrade.

olehermanse:
@nickanderson IMO, it would be wrong to set an exact version here explicitly, when you are between versions your version number should reflect this. If we want the self upgrade policy to handle this better, that's what needs to be fixed, I think.

cfbs.json
```diff
@@ -8,7 +8,18 @@
       "tags": ["supported", "base"],
       "repo": "https://github.com/cfengine/masterfiles",
       "by": "https://github.com/cfengine",
-      "steps": ["run ./prepare.sh -y", "copy ./ ./"]
+      "steps": ["run EXPLICT_VERSION=ChangeMe EXPLICIT_RELEASE=ChangeMe ./prepare.sh -y",
```
olehermanse, line 11:
```suggestion
      "steps": ["run ./prepare.sh -y",
```

olehermanse, line 11:
```suggestion
      "steps": ["run ./prepare.sh -y",
```

olehermanse (changes requested):

cfbs.json
```diff
@@ -8,7 +8,18 @@
...
+                "delete cfe_internal/enterprise/ha/ha_info.json",
+                "delete .github",
+                "delete inventory/README.md",
+                "delete lib/README.md",
+                "delete LICENSE",
+                "delete modules/promises",
+                "delete .no-distrib",
+                "delete services/autorun/README.md",
+                "delete templates/README.md",
+                "copy ./ ./"]
```
olehermanse, line 22: @nickanderson actually, this is not needed. The current (newest) version of prepare.sh has all of these deletes inside of it.

See:

https://github.com/cfengine/masterfiles/blob/998f71dcbf2f982d2d0783e58bd22288efddf377/prepare.sh#L61-L100

This matches the latest version in the index (3.22 non LTS, which is not the default, since we default to latest LTS):

https://github.com/cfengine/build-index/blob/bf1ae15c83ef3ac2ff7acfd89e58a28d29d2e5c5/versions.json#L898-L901

So I guess the latest improvements to the `prepare.sh` script have not been backported to 3.21.x. Feel free to do this, to prepare mpf for the upcoming 3.21 release, if you'd like.

olehermanse (changes requested):

cfbs.json
olehermanse: Looks like that backporting has also been done already: https://github.com/cfengine/masterfiles/pull/2643

PR #2747, opened 2023-10-19
ENT-10793: Added rendering of custom mustache templates to $(sys.workdir)/modules

nickanderson:
Merge together:
- https://github.com/cfengine/documentation/pull/3110

With this change mustache templates found in modules/mustache (relative to the
root of the policy set) are rendered to $(sys.workdir)/modules using datastate()
from the update.cf policy entry.

Ticket: ENT-10793
Changelog: Title


craigcomstock (approved):
seems cool. :+1:

PR #2748, opened 2023-10-19
ENT-10921: Improved failure logging during federated reporting schema import

craigcomstock:
Saves full log in /var/cfengine/outputs

TODO: clean up old /var/cfengine/outputs/*-schema-attach-*.log files

Ticket: ENT-10789
Changelog: title

### DIFF
--- a/templates/federated_reporting/import.sh
+++ b/templates/federated_reporting/import.sh
@@ -127,18 +127,25 @@ for file in $dump_files; do
   if [ ! -f "${file}.failed" ]; then
     hostkey=$(basename "$file" | cut -d. -f1)
     logfile="$WORKDIR"/outputs/"$hostkey"-schema-attach-$(date +%F-%T)-failure.log
-    "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
-                        -c "SET SCHEMA 'public'; SELECT attach_feeder_schema('$hostkey', ARRAY[$table_whitelist]);" \
-      > "$logfile" 2>&1 || one_failed=1
-      if [ "$one_failed" = "0" ]; then
-        rm -f "$logfile"
-      else
-        any_failed=1
-        log "Attaching schemas: FAILED for $hostkey, check $logfile for details"
-        log "last 10 lines of $logfile"
-        tail -n 10 "$logfile"
-      fi
-      one_failed=0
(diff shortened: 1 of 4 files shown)


craigcomstock:
Should wait for affected users to report that this change is helpful in debugging.

nickanderson (approved):

templates/federated_reporting/import.sh
```diff
@@ -124,9 +125,17 @@ log "Attaching schemas"
 for file in $dump_files; do
   if [ ! -f "${file}.failed" ]; then
     hostkey=$(basename "$file" | cut -d. -f1)
+    logfile="$WORKDIR"/outputs/"$hostkey"-schema-attach-$(date +%F-%T)-failure.log
     "$CFE_BIN_DIR"/psql -U $CFE_FR_DB_USER -d cfdb --set "ON_ERROR_STOP=1" \
                         -c "SET SCHEMA 'public'; SELECT attach_feeder_schema('$hostkey', ARRAY[$table_whitelist]);" \
```
craigcomstock, line 130: @nickanderson how would you integrate/configure debug logging for federated imports?

We can add
```
                        -c "SET client_min_messages TO DEBUG5" \
```

In this line and any other psql commands we run here and capture logs as I have done here to track down problems easier.z 

Maybe an augment to set the log level or just "debug" as a boolean would be easier.

nickanderson: First thing that comes to mind would be a new variable, defaulted to NOTICE but overridable via augments.

Then someone could set that variable to any of the allowed values:

DEBUG5, DEBUG4, DEBUG3, DEBUG2, DEBUG1, LOG, NOTICE, WARNING, and ERROR.

nickanderson: Oh, that's a shell script, not a renderd mustache file.

craigcomstock: my proposal here doesn't work yet... bash quoting :'(

craigcomstock: I tested what I have here now and think this looks good. I bailed on trying to figure out bash escaping, not worth the effort so just copy/pasted the command twice.

nickanderson (approved):

craigcomstock:
cherry picks
https://github.com/cfengine/masterfiles/pull/2781
https://github.com/cfengine/masterfiles/pull/2780

PR #2760, opened 2023-11-07
CEF-3982: Aligned systemd service templates with core

nickanderson:


templates/cf-apache.service.mustache
```diff
@@ -17,4 +17,3 @@ UMask=0177
 
 [Install]
 WantedBy=multi-user.target
-WantedBy=cfengine3.service
```
olehermanse, line 20: Could put link to related PR / commit in commit message.

Changelog entry does not explain the result of this change to a user.

olehermanse (changes requested):

templates/cf-apache.service.mustache
nickanderson: changed.

olehermanse (approved):
LGTM

vpodzime (approved):
ACK

PR #2764, opened 2023-11-14
ENT-10904: Added --quiet to psql commands where needed in federated reporting scripts

craigcomstock:
If an environment has SET commands in psqlrc or ~/.psqlrc then output can include the SET response and break parsing.

Ticket: ENT-10904
Changelog: title


craigcomstock:
cherry picks
https://github.com/cfengine/masterfiles/pull/2779
https://github.com/cfengine/masterfiles/pull/2778

PR #2765, opened 2023-11-15
ENT-10664: Added documentation about specifying self upgrade target version

nickanderson:


MPF.md
```diff
@@ -392,26 +392,37 @@ the appropriate platform directory. Clients will automatically download and
...
+  }
 }
 ```
 
 **Notes:**
 
 - This policy is specific to CFEngine Enterprise.
-- The negative look ahead regular expression is useful because it automatically
-  turns off on hosts after they reach the target version.
+- If using a regular expression based on CFEngine version, use a negative look ahead to disable self upgrade when the host reaches the desired version. e.g. `cfengine_3_10_(?!2$)\\d+` matches hosts running CFEngine 3.10 but not 3.10.2 specifically.
```
craigcomstock, line 421: maybe change this to 3.21.2 or 3.18.something aka update the example?

nickanderson: Changed, squashed.

craigcomstock (approved):

PR #2776, opened 2023-11-17
ENT-10918: Fixed support for automatically installing semanage on el9 for federated reporting

nickanderson:
This change let's rocky 9 and redhat 9 hosts know which package contains
semanage so that it can be automatically installed for managing permissions in
federated reporting.

Ticket: ENT-10918
Changelog: Title

PR #2777, opened 2023-11-17
ENT-10918: Fixed support for automatically installing semanage on el9 for federated reporting (3.21)

nickanderson:
This change let's rocky 9 and redhat 9 hosts know which package contains
semanage so that it can be automatically installed for managing permissions in
federated reporting.

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1617,7 +1617,7 @@ This directive can be used to control which versions of the SSL/TLS protocol wil
 
 **History:**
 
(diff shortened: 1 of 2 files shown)


cf-bottom:

vpodzime (commented):
No need for the second commit from #2776?

nickanderson:
> No need for the second commit from #2776?

No, it was needed. I hadn't finished squashing the first PR. It's in here now.

PR #2785, opened 2023-11-28
ENT-10951: Automatically include $(sys.policy_hub) in acl, allowconnects, allowallconnects

nickanderson:
This change more specifically includes the hubs ip address in the default acl, allowconnects and allowallconnects.

sys.policy_hub/16 remains the default for def.acl, but new _derived suffixed variables which are used in the final configuration have been added.

Previously, setting acl, allowconnects, or allowallconnects and omitting the policy server would prevent the policy server from accessing it's own resources. Recovery from this mistake was tedious. Now, the policy server ip address is joined with the configured values for acl, allowconnects, and allowallconnects in the new _derived suffixed variables so that the policy servers IP is always included unless a class disabling the behavior is defined.

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -79,6 +79,22 @@ bundle common def
         if => and(not(isvariable("override_data_acl")), not(isvariable("acl"))),
         meta => { "defvar" };
 
+    !disable_always_accept_policy_server_acl::
+      "acl_derived" -> { "ENT-10951" }
+        slist => {
+                   @(def.acl),
+                   "$(sys.policy_hub)"
+        },
+        comment => "Define an acl for the machines to be granted accesses";
+
+    disable_always_accept_policy_server_acl::
+      "acl_derived" -> { "ENT-10951" }
+        slist => {
+                   @(def.acl)
+        },
+        comment => "Define an acl for the machines to be granted accesses";
+
+    any::
 
       "control_server_allowconnects" -> { "ENT-10212" }
         slist => { "127.0.0.1" , "::1", @(def.acl) },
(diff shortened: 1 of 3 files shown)


controls/def.cf
```diff
@@ -79,6 +79,22 @@ bundle common def
         if => and(not(isvariable("override_data_acl")), not(isvariable("acl"))),
         meta => { "defvar" };
 
+    !disable_always_accept_policy_server_acl::
+      "acl_derived" -> { "ENT-10951" }
+        slist => {
+                   "@(def.acl)",
+                   "$(sys.policy_hub)"
```
olehermanse, line 86: Why are you using the `sys.policy_hub` variable here, but `sys.policy_server` in other places?

controls/def.cf
```diff
@@ -94,6 +94,22 @@ bundle common def
                            " allowallconnects in body server control if not",
                            " already specified" );
 
+    !disable_always_accept_policy_server_allowallconnects::
+      "control_server_allowallconnects_derived" -> { "ENT-10951" }
+        slist => { @(def.control_server_allowallconnects), "$(sys.policy_server)" },
+        comment => concat( "We want to define the default setting for",
```
craigcomstock, line 134: Kind of feel like the comment should mention including `$(sys.policy_server)` or not including it below.

craigcomstock (changes requested):
definitely squash your commits, it seems you have at least 3 commits that change the same code? Otherwise, looks good!

controls/def.cf
craigcomstock: I see you fixed this in a later commit. 👍

controls/def.cf
nickanderson: Oh, just a bug. This pr should have been a draft.

nickanderson:
> definitely squash your commits, it seems you have at least 3 commits that change the same code? Otherwise, looks good!

yeah, should have been marked as draft.

controls/def.cf
nickanderson: fixed.

nickanderson:
OK, fixed up, squashed. Ready for review.

controls/def.cf
olehermanse: @nickanderson this one is still here ^

olehermanse (changes requested):
@nickanderson commit message and PR title still says sys.policy_server (not sys.policy_hub).

Looks mostly good otherwise.

nickanderson:
> @nickanderson commit message and PR title still says sys.policy_server (not sys.policy_hub).
> 
> Looks mostly good otherwise.

fixed.
LMK when you have test driven.

olehermanse:
> fixed.
> LMK when you have test driven.

@nickanderson still looks the same, did you forget to push?

olehermanse (changes requested):
Commit message and PR title still say `sys.policy_server`.

nickanderson:
> > fixed.
> > LMK when you have test driven.
> 
> @nickanderson still looks the same, did you forget to push?

I forgot to auth when I pushed :D took your suggestions, squashed.

craigcomstock (approved):

PR #2797, opened 2023-12-12
ENT-11096: Policy now manages Mission Portals httpd.conf ownership and permissions

nickanderson:
See also: https://github.com/cfengine/buildscripts/pull/1345

Ticket: ENT-11096
Changelog: Title


vpodzime (approved):
ACK

PR #2799, opened 2023-12-12
ENT-8785: Added recommendation for installing gnu parallel on federated reporting superhubs

nickanderson:
Ticket: ENT-8785
Changelog: Title

### DIFF
--- a/cfe_internal/recommendations.cf
+++ b/cfe_internal/recommendations.cf
@@ -9,6 +9,17 @@ bundle agent MPF_class_recommendations
         if => "cfengine_internal_purge_policies";
 }
 
+bundle agent federated_reporting_superhub_recommendations
+{
+  meta:
+      "tags" slist => { "cfengine_recommends" };
+
+  reports:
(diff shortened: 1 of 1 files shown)


cfe_internal/recommendations.cf
```diff
@@ -9,6 +9,17 @@ bundle agent MPF_class_recommendations
...
 
+bundle agent federated_reporting_superhub_recommendations
+{
+  meta:
+      "tags" slist => { "cfengine_recommends" };
+
+  reports:
+    cfengine_enterprise_federation:am_superhub::
+      "CFEngine recommends installing gnu parallel on federated reporting superhubs."
+        if => not(isgreaterthan(length(packagesmatching( "parallel", ".*", ".*", ".*")), 0 ));
```
craigcomstock, line 20: seems like we could do better than this expression, something like `package_installed("parallel")` where are our user-defined/custom functions like custom promises?

but yes, this seems fine, though we should likely update the docs as well right?

craigcomstock (approved):

PR #2804, opened 2023-12-14
CFE-2724: Removed quotes from `arglist` attribute in `commands` promises

larsewi:
Now that whitespaces are preserved in the `arglist` attribute of the `commands` promises, we no longer need to use extra quotes. In fact, these quotes become a part of the argument, causing some methods to fail in the MPF.

Merge together with:
 - https://github.com/cfengine/core/pull/5378
 - https://github.com/cfengine/enterprise/pull/775

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -757,7 +757,11 @@ bundle agent superhub_schema
       "$(cfengine_enterprise_federation:config.bin_dir)/psql_wrapper.sh"
         arglist => {
                      "cfdb",
+@if minimum_version(3.24)
+                     "select superhub_schema('$(sys.key_digest)');",
+@else
                      `"select superhub_schema('$(sys.key_digest)');"`,
+@endif
                    },
         classes => psql_wrapper_exit_codes;
 }
@@ -775,7 +779,11 @@ bundle agent ensure_feeders
       "$(cfengine_enterprise_federation:config.bin_dir)/psql_wrapper.sh"
         arglist => {
                      "cfdb",
+@if minimum_version(3.24)
+                     "select ensure_feeders($(feeders_arg));"
+@else
                      `"select ensure_feeders($(feeders_arg));"`
+@endif
                    },
         classes => psql_wrapper_exit_codes,
         if => isgreaterthan(length(feeders), 0);


cfe_internal/enterprise/federation/federation.cf
```diff
@@ -757,7 +757,7 @@ bundle agent superhub_schema
       "$(cfengine_enterprise_federation:config.bin_dir)/psql_wrapper.sh"
         arglist => {
                      "cfdb",
-                     `"select superhub_schema('$(sys.key_digest)');"`,
+                     "select superhub_schema('$(sys.key_digest)');",
```
nickanderson, line 760: Assuming this change is done for 3.24.0 ... When I upgrade the first thing I will do is upgrade my policy set. So I will be running e.g. 3.18.6 with this policy for some time before I upgrade to 3.24.0.

Ideally we could handle this automatically. Seems like it could be difficult to detect reliably, but this is not an elegant transition.

People will have to duplicate all the cases where they have double quoting in argslist where the behavior changes and make sure that policy uses the right quoting based on the cfengine version.

nickanderson (changes requested):
This change as it stands would not work during upgrades while I am running old binaries and new policy.

cfe_internal/enterprise/federation/federation.cf
larsewi: Yes, it would be a breaking change. Do you think we should not make it?

larsewi:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/10202/)

cfe_internal/enterprise/federation/federation.cf
larsewi: Ping @nickanderson

nickanderson: Do I understand correctly? ...

- Users must locate all use of `argslist` within their policy
- Users must add the macro treatment (or similar) as done in this file **if** the promise using `argslist` does **not** execute within a shell.
- Users must NOT add the macro treatement (or similar) if the promise using `argslist` uses a shell

Am I overlooking anything?

larsewi: Sounds about right :)

nickanderson: Ok, we just need a good change in behavior post to outline the details I guess.

nickanderson (approved):

larsewi:
https://northerntech.atlassian.net/browse/CFE-4303

PR #2813, opened 2023-12-28
Check if cf-apache.service is active to see if it's systemd-managed

vpodzime:
We don't enable the cf-apache.service, it's started by the cfengine3 umbrella service as its (weak) dependency. So only the cfengine3 service is enabled normally.

Ticket: ENT-11130
Changelog: None


vpodzime:
jenkins with deployment tests only: [Build Status](https://ci.cfengine.com/job/pr-pipeline/10225/)

PR #2816, opened 2024-01-02
Fixes for FR policy fixing up SELinux context of ~cftransport/.ssh

vpodzime:
Multiple issues discovered and fixed. Plus one piece of polishing.

### DIFF
--- a/cfe_internal/enterprise/federation/federation.cf
+++ b/cfe_internal/enterprise/federation/federation.cf
@@ -378,7 +378,7 @@ bundle agent transport_user
     # _stdlib_path_exists_<command> and paths.<command> are defined is masterfiles/lib/paths.cf
     selinux_enabled.cftransport_fcontext_missing.default:_stdlib_path_exists_semanage::
       "$(default:paths.semanage) fcontext -a -t ssh_home_t '$(home)/.ssh(/.*)?'";
-    selinux_enabled.incorrect_ssh_context.default:_stdlib_exists_restorecon::
+    selinux_enabled.incorrect_ssh_context.default:_stdlib_path_exists_restorecon::
       "$(default:paths.restorecon) -R -F $(home)/.ssh/";
 
     any::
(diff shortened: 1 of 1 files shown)


vpodzime:
Manually tested to verify that it fixes issues on a CentOS 7 hub enabled as a feeder. Really strange it hasn't been breaking FR tests.

vpodzime:
> Manually tested to verify that it fixes issues on a CentOS 7 hub enabled as a feeder.

It just requires two agent runs to properly do the full setup. https://github.com/cfengine/masterfiles/pull/2817 adds a more complicated change making sure a single agent run is enough, but we may want to stick to this simpler change for backports.

cfe_internal/enterprise/federation/federation.cf
```diff
@@ -300,29 +300,34 @@ bundle agent transport_user
           expression => not(returnszero("$(default:paths.semanage) fcontext -l | grep '$(home)/.ssh(/.*)?'", "useshell")),
           if => fileexists("$(home)");
     enabled.selinux_enabled::
+        # For all the files below it must be true that if they exist they need
```
craigcomstock, line 303: I appreciate your explanation here but have to say this block has become rather unreadable. I wonder if we could do better? The logic reads to me as correct after some time looking at it.

craigcomstock (approved):

cfe_internal/enterprise/federation/federation.cf
vpodzime: Well, I know how to make this a lot nicer in Python. Or any language with functions and loops, really. But no clue how to make it nicer in CFEngine policy. Maybe @nickanderson can help polish this later.

vpodzime: We could use a list of files and then create an array with those files as keys and the results of the SELinux labels as values and then search for a false value in `getindices()` on the array. But that's similarly ugly, AFAICT.

PR #2817, opened 2024-01-02
Follow up for FR SSH SELinux context fixes

vpodzime:
Needs #2816 to be merged first. Adds one more commit that makes bigger changes in the policy, probably too disruptive for backporting.


craigcomstock (approved):
I don't see why we couldn't back port these changes. A nice refactoring. :+1:

craigcomstock:
I removed the cherry-pick label because we decided these changes were too disruptive.

vpodzime:
Added the label back, I think we might want to reconsider this for the next round of releases.

PR #2827, opened 2024-01-23
CFE-4313: Adjusted package_methods generic and yum to allow for packages with dashes

craigcomstock:
Trying to install python3 would result in attempting to install all python3-*.* packages which includes hundreds of python modules.

Ticket: CFE-4313
Changelog: title

### DIFF
--- a/lib/packages.cf
+++ b/lib/packages.cf
@@ -1003,7 +1003,7 @@ body package_method yum
       package_list_arch_regex    => "$(rpm_knowledge.rpm3_arch_regex)";
 
       package_installed_regex => ".*";
-      package_name_convention => "$(name)-$(version).$(arch)";
+      package_name_convention => "$(name)";
 
       # just give the package name to rpm to delete, otherwise it gets "name.*" (from package_name_convention above)
       package_delete_convention => "$(name)";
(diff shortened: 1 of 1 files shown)


lib/packages.cf
```diff
@@ -1003,7 +1003,7 @@ body package_method yum
       package_list_arch_regex    => "$(rpm_knowledge.rpm3_arch_regex)";
 
       package_installed_regex => ".*";
-      package_name_convention => "$(name)-$(version).$(arch)";
+      package_name_convention => "$(name)";
```
nickanderson, line 1006: Wouldn't this result in being unable to specify the version and/or architecture for a package?

nickanderson (commented):
hummm

lib/packages.cf
craigcomstock: Likely a better fix would be to change the C code to anchor the `$(version).$(arch)` portion to the end of the name. I can investigate that instead.

craigcomstock:
I took a peek at the C code in core/cf-agent/verify_packages.c and decided it's too complex for a quick fix. So will close and leave CFE-4313 open/todo.

PR #2834, opened 2024-02-01
CFE-4323: FreeBSD service management now uses one prefixed service commands

nickanderson:
On FreeBSD service commands only work if a service has been enabled in
/etc/rc.conf. If it has not been enabled then a services promise with a
service_policy of "start" or "stop" results in a failure to change the system
state during each agent execution. Switching to "one" prefixed commands, e.g.
onestatus, onestop, and onestart allows services that are not enabled to still
be managed and does not interfere with services that have been enabled in
/etc/rc.conf.

Ticket: CFE-4323
Changelog: Title

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -139,6 +139,9 @@ bundle agent standard_services(service,state)
 # methods:
 #     "SSHD should be running" usebundle => standard_services("sshd", "start");
 # ```
+#
+# **Notes:**
+# FreeBSD uses "one" prefixed commands, e.g. onestatus, onestart, onestop so that services that are not enabled can still be managed
 {
   vars:
       "c_service" string => canonify("$(service)");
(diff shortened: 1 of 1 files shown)


lib/services.cf
```diff
@@ -171,15 +174,17 @@ bundle agent standard_services(service,state)
 
       "have_init" expression => fileexists($(init));
 
-    chkconfig.have_init.freebsd::
-      "running" -> { "CFE-3513" }
-        expression => returnszero("$(init) onestatus > /dev/null", "useshell");
-
-    chkconfig.have_init.!freebsd::
+    chkconfig.have_init::
```
nickanderson, line 177: As previously discussed in https://github.com/cfengine/masterfiles/pull/1892 checkconfig does not relate to freebsd, so this was removed for clarity.

lib/services.cf
```diff
@@ -139,6 +139,9 @@ bundle agent standard_services(service,state)
 # methods:
 #     "SSHD should be running" usebundle => standard_services("sshd", "start");
 # ```
+#
+# **Notes:**
+# FreeBSD uses "one" prefixed commands, e.g. onstatus, onestart, onestop so that services that are not enabled can still be managed
```
craigcomstock, line 144: maybe?

```suggestion
# FreeBSD uses "one" prefixed commands, e.g. onestatus, onestart, onestop so that services that are not enabled can still be managed
```

craigcomstock (dismissed):
one comment typo I think, otherwise looks good.

craigcomstock (commented):
squash commits?

craigcomstock (approved):

PR #2841, opened 2024-02-22
Watchdog to ignore stale pidfile

mikeweilgart:
Fixes CFE-4335.


mikeweilgart:
I don't have an easy way to test it, but in isolation the code seems to work.

/proc exists at least on all Linux + Solaris, though I'm unable to find confirmation that it's required by POSIX.  For the basic point of, "how old is this process?" (checking mtime with ls -t) it should work.

Options to head and ls only include those specified by POSIX.

Hat tip to https://stackoverflow.com/q/40205638/5419599 where I got the idea for this code.

mikeweilgart:
Does not work on MacOS, not sure how broadly watchdog is intended to work.

mikeweilgart:
Also, shouldn't the pidfile be in /var/run/ ?  That exists on all modern unixes, including macos, and gets wiped out on boot which is the most common reason for stale pidfiles generally (existing past boot).  Reference: https://www.pathname.com/fhs/2.2/fhs-5.13.html

vpodzime (approved):
Looks good to me.

nickanderson:
@flynn1973 Can you confirm if this change improves the situation for you?

flynn1973:
a quick run of the changed script shows a missing "then" ;-)

https://github.com/cfengine/masterfiles/blob/70eb84ad845af716fdc53b1299544e7b521a563d/cfe_internal/core/watchdog/templates/watchdog.mustache#L60

error:

root@aixtest01: /root # /var/cfengine/bin/watchdog_changed
/var/cfengine/bin/watchdog_changed[52]: syntax error at line 64 : `else' unexpected


after adding it the missing statemen the script runs without errors....all in all this change will definitely improve the situation..good idea.

nickanderson:
@flynn1973 Patched.

nickanderson:
@mikeweilgart do you want to squash these commits together and amend the commit body to include something like 

```
Ticket: CFE-4335
Changelog: AIX watchdog now handles stale pids
```

nickanderson:
@mikeweilgart thanks, one more thing, can you re-word the commit subject to be past tense?

Something like "Made AIX watchdog ignore stale pid files".

nickanderson:
@mikeweilgart I think this watchdog is specific to AIX. It might work on Linux as well, but the Linux one is much more simple.

The aix watchdog uses this template here: https://github.com/cfengine/masterfiles/blob/master/cfe_internal/core/watchdog/watchdog.cf#L174-L179

and the linux watchdog, well the watchdog used when /etc/cron.d is present uses https://github.com/cfengine/masterfiles/blob/master/cfe_internal/core/watchdog/watchdog.cf#L225-L227

Would be good to combine them, but that would be separate work.

nickanderson:
Thanks @mikeweilgart !

PR #2843, opened 2024-02-22
Updated MPF docs to consistently use Augments syntax v2

nickanderson:


nickanderson:
Preview here: http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-99/output/_site/reference-masterfiles-policy-framework.html

nickanderson:
squashed.

craigcomstock (changes requested):
I tend to think this makes things look more complicated than they need to be.

I know it is "possible" but I wouldn't recommend it unless you need the extra capabilities which none of these examples need right?

craigcomstock (approved):

MPF.md
```diff
@@ -88,9 +88,13 @@ With an augments like this:
...
-  "vars": {
-    "house": "Gryffindor",
-    "mpf_update_policy_master_location": "/srv/cfengine/$(sys.flavor)/$(def.house)"
+  "variables": {
+    "default:def.house": {
+      "value": "Gryffindor"
+    },
+    "default:def.mpf_update_policy_master_location": {
+      "value": "/srv/cfengine/$(default:sys.flavor)/$(default:def.house)"
+    }
```
olehermanse, line 97:
```suggestion
    "house": "Gryffindor",
    "mpf_update_policy_master_location": "/srv/cfengine/$(default:sys.flavor)/$(default:def.house)"
    }
```

olehermanse (approved):

PR #2849, opened 2024-02-28
CFE-4342: When failing to detect platform, inventory attribute "OS" now defaultss to PRETTY_NAME from os-release as a fallback

olehermanse:


olehermanse:
See: https://github.com/cfengine/masterfiles/pull/2850

PR #2853, opened 2024-03-21
CFE-4359: Update regex strings to raw to squelch warnings introduced with python 3.12

wilpig:
Starting with python 3.12 using backslashes generate warnings and will be escalated to errors in a future release.

https://docs.python.org/3/library/re.html

> Regular expressions use the backslash character ('\') to indicate special forms or to allow special characters to be used without invoking their special meaning. This collides with Python’s usage of the same character for the same purpose in string literals; for example, to match a literal backslash, one might have to write '\\\\' as the pattern string, because the regular expression must be \\, and each backslash must be expressed as \\ inside a regular Python string literal. Also, please note that any invalid escape sequences in Python’s usage of the backslash in string literals now generate a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning) and in the future this will become a [SyntaxError](https://docs.python.org/3/library/exceptions.html#SyntaxError). This behaviour will happen even if it is a valid escape sequence for a regular expression.
> 
> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with 'r'. So r"\n" is a two-character string containing '\' and 'n', while "\n" is a one-character string containing a newline. Usually patterns will be expressed in Python code using this raw string notation.
> 

See CFE-4359 for warning examples.

### DIFF
--- a/modules/packages/vendored/apt_get.mustache
+++ b/modules/packages/vendored/apt_get.mustache
@@ -193,12 +193,12 @@ def list_updates(online):
         #                               name                   old version       new version
         #                                 |                         |                 |
         #                         /-------+-------\              /--+--\       /------+-------\
-        match = re.match("^Inst\s+(?P<name>[^\s:]+)(?::\S+)?\s+\[[^]\s]+\]\s+\((?P<version>\S+)" +
+        match = re.match(r"^Inst\s+(?P<name>[^\s:]+)(?::\S+)?\s+\[[^]\s]+\]\s+\((?P<version>\S+)" +
 
         #                repository(ies)      arch (might be optional)
         #                       |               |
         #                    /--+-\   /---------+---------\
-                         "(?:\s+\S+)*?(\s+\[(?P<arch>[^]\s]+)\])?\).*", line)
+                         r"(?:\s+\S+)*?(\s+\[(?P<arch>[^]\s]+)\])?\).*", line)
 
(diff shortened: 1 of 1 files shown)


larsewi (approved):
Great, thanks @wilpig 🚀 @cf-bottom can you please run this through Jenkins? 😇

olehermanse:
@cf-bottom jenkins, please :)

craigcomstock:
> Alright, I triggered a build:
> 
> [Build Status](https://ci.cfengine.com//job/pr-pipeline/10553/)
> 
> **Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10553/
> 
> **Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10553/

I checked and saw two flaky failures in CI so good to go from that point of view I would say.

craigcomstock (approved):

vpodzime (approved):
I'm surprised `\s` ever worked with normal strings. Definitely a fix no matter what Python version runs this...

larsewi:
I'm not authorized to merge. @nickanderson can you merge please ? :)

nickanderson:
cherry-picked to 3.21.x: https://github.com/cfengine/masterfiles/pull/2866

PR #2854, opened 2024-03-26
CFE-4345: Added freebsd package_module and package_inventory since we have packages module available (3.18)

craigcomstock:
- Added ability to influence default package manager and inventory via Augments
- Added freebsd package_module and package_inventory since we have pkg packages module available

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1735,6 +1735,53 @@ For example:
 
 - Introduced in CFEngine 3.23.0, 3.21.3, 3.18.6
 
+### Configure default package manager
+
+The MPF specifies the package module to use for managing packages and collecting software inventory based on the detected platform. Define `default:def.default_package_module` as a data structure keyed with values matching the value of `sys.flavor` for the platforms you wish to target.
+
+**Example:**
+
+```json
+{
+  "variables": {
+    "default:def.default_package_module": {
+      "value": {
+        "ubuntu_20": "snap",
+        "aix_7": "yum"
+      },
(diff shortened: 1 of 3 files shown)


MPF.md
```diff
@@ -1735,6 +1735,53 @@ For example:
...
+      },
+        "comment": "This variable provides the ability to override the default package manager to use for a platform. Keys are based on the value of $(sys.flavor) for the targeted platform."
+    }
+  }
+}
+```
+
+**History:**
+
+* Added in CFEngine 3.24.0
```
nickanderson, line 1760:
```suggestion
* Added in CFEngine 3.24.0, 3.21.5, 3.18.8
```

nickanderson (approved):
We should update the history.

MPF.md
```diff
@@ -1735,6 +1735,53 @@ For example:
...
+        },
+        "comment": "This variable provides the ability to extend the default package managers to inventory for a platform. Keys are based on the value of $(sys.flavor) for the targeted platform."
+      }
+  }
+}
+```
+
+**History:**
+
+* Added in CFEngine 3.24.0
```
craigcomstock, line 1783: @nickanderson would we not update this one also? And then update master with these version changes ;) so I will wait for your answer about this one and then re-do the changes from master down... (the version bits here)

nickanderson: Yes, this one too. I guess I just missed it. Doesn't matter to me which way the things are picked from as long as they all match.

PR #2855, opened 2024-03-26
CFE-4345: Added freebsd package_module and package_inventory since we have pkg packages module available (3.21)

craigcomstock:
- Added ability to influence default package manager and inventory via Augments
- Added freebsd package_module and package_inventory since we have pkg packages module available

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1807,6 +1807,53 @@ For example:
 
 - Introduced in CFEngine 3.23.0, 3.21.3
 
+### Configure default package manager
+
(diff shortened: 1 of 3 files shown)


MPF.md
```diff
@@ -1807,6 +1807,53 @@ For example:
...
+      },
+        "comment": "This variable provides the ability to override the default package manager to use for a platform. Keys are based on the value of $(sys.flavor) for the targeted platform."
+    }
+  }
+}
+```
+
+**History:**
+
+* Added in CFEngine 3.24.0
```
nickanderson, line 1832:
```suggestion
* Added in CFEngine 3.24.0, 3.21.5, 3.18.8
```

nickanderson (approved):
I think we should update the history here too.

PR #2857, opened 2024-03-27
ENT-11480: Removed hour delay between CFEngine Enterprise PostgreSQL recommendation checks

nickanderson:
This change switches from the warn_only action body to the policy(p) action body
so that the PostgreSQL recommendation checks are consistent and not appearing to
flap being checked only once an hour (since the warn_only action body sets
ifelapsed to 60). This also aligns with other recommendation policies.

Ticket: ENT-11480
Changelog: Title


vpodzime (approved):
Looks good to me. Thanks!

PR #2860, opened 2024-03-29
CFE-4364: Made autorun inputs sort in lexical order

nickanderson:
When autorun is in use, lexical order is applied for bundle actuation. This
change applies lexical ordering to the list of policy files that are found and
included in inputs. Sorting the files explicitly ensures more predictable
behavior and can assist in debugging.

Ticket: CFE-4364


olehermanse (approved):

craigcomstock (approved):
cool.

PR #2862, opened 2024-04-02
Allow disabling recommendations using a CMDB class

vpodzime:
CMDB classes default to the `data` namespace. So if the user adds `cfengine_recommendations_disabled` as a class in their CMDB data for the hub machine, checking the class in the default namespace is not good enough to disable recommendations.

### DIFF
--- a/controls/def.cf
+++ b/controls/def.cf
@@ -667,7 +667,7 @@ bundle common def
         expression => some( ".+", "def.control_executor_runagent_socket_allow_users" );
 
       "cfengine_recommendations_enabled"
-        expression => "!cfengine_recommendations_disabled";
+        expression => "!cfengine_recommendations_disabled.!data:cfengine_recommendations_disabled";
 
       ### Enable special features policies. Set to "any" to enable.
 


controls/def.cf
```diff
@@ -667,7 +667,7 @@ bundle common def
         expression => some( ".+", "def.control_executor_runagent_socket_allow_users" );
 
       "cfengine_recommendations_enabled"
-        expression => "!cfengine_recommendations_disabled";
+        expression => "!cfengine_recommendations_disabled.!data:cfengine_recommendations_disabled";
```
nickanderson, line 670: I am not a big fan of having multiple classes that do the same thing generally. This class is suggested because it's not clear how to define the class in MP CMDB UI, seems to me better to improve this elsewhere than add another class.

nickanderson (commented):
I would look to improve this in some different way.

perhaps one or more of: 
- Emit a report that tells people to define `default:cfengine_recommendations_disabled`
- Update the docs to reference `default:`
- Improve CMDB UI

nickanderson:
> I would look to improve this in some different way.
> 
> perhaps one or more of:
> 
>     * Emit a report that tells people to define `default:cfengine_recommendations_disabled`
> 
>     * Update the docs to reference `default:`
> 
>     * Improve CMDB UI

@vpodzime what do you think about this proposal? https://github.com/cfengine/masterfiles/pull/2863
It does two of the suggested things above.

controls/def.cf
vpodzime: Yes, I completely agree. This is really just my "user story" persona trying to scratch my own itch.

vpodzime:
> @vpodzime what do you think about this proposal? #2863
> It does two of the suggested things above.

Let's continue with the discussion there.

PR #2863, opened 2024-04-02
ENT-11523: Improved instructions for disabling recommendations

nickanderson:
This change introduces a reports promise (`R: Note: Recommendations can be disabled by defining 'default:cfengine_recommendations_disabled'`) that is activated if any of the existing recommendations are emitted.

For example:

```
root@hub:~/cfbs# cf-agent -Kf update.cf; cf-agent -K
 warning: edit_line warning promised: - #shared_buffers = 128MB                 # min 128kB
 warning: edit_line warning promised: - #effective_cache_size = 4GB
 warning: Should edit file '/var/cfengine/state/pg/data/postgresql.conf', but only warning promised
 warning: Warnings encountered when actuating files promise '/var/cfengine/state/pg/data/postgresql.conf'
R: CFEngine Recommended Settings:
R: shared_buffers = 4850MB
R: effective_cache_size = 13590MB
 warning: Method 'postgresql_conf_recommendations' invoked repairs, but only warnings promised
R: Note: All recommendations can be disabled by defining 'default:cfengine_recommendations_disabled'
root@hub:~/cfbs# cf-agent -K --define default:cfengine_recommendations_disabled
root@hub:~/cfbs#
```

It also clarifies the documentation a bit more, explicitly calling out the `default` namespace.

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -2539,12 +2539,12 @@ Maximum time between automatic WAL checkpoints. If this value is specified witho
 ## Recommendations
 
 The MPF includes policy that inspects the system and makes recommendations about
-the configuration of the system. When `cfengine_recommendations_enabled` is
-defined bundles tagged `cfengine_recommendation` are executed in lexical order.
-`cfengine_recommendations_enabled` is defined by default when
-`cfengine_recommendations_disabled` is **not** defined.
+the configuration of the system. When `default:cfengine_recommendations_enabled` is
+defined bundles tagged `cfengine_recommends` are executed in lexical order.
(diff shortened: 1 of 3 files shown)


vpodzime (approved):
Looks good to me. However, I still think we should do more for users which we want to keep in Mission Portal mostly/only. Which is probably a matter of improving the CMDB UX, but one could argue that classes like `cfengine_recommendations_disabled` don't even belong to CMDB data, they are more like MPF configuration and would perhaps better fit into the inputs for the masterfiles module or something like that.

craigcomstock (approved):

PR #2864, opened 2024-04-04
ENT-11526: Changed mission-portal apache restart to graceful to minimize service interruptions

craigcomstock:
In our CI system deployment tests we go through test steps quickly.
Fairly often the apache config is updated by policy and a restart is initiated.
By changing to apachectl graceful we should avoid some of the server availability during tests problems we have been seeing.

Ticket: ENT-11526
Changelog: title

with https://github.com/cfengine/core/pull/5480

### DIFF
--- a/cfe_internal/enterprise/mission_portal.cf
+++ b/cfe_internal/enterprise/mission_portal.cf
@@ -228,7 +228,7 @@ bundle agent mission_portal_apache_from_stage(config, staged_config)
                            "place only if we have stopped apache already.");
 
       "LD_LIBRARY_PATH=$(sys.workdir)/lib:$LD_LIBRARY_PATH $(sys.workdir)/httpd/bin/apachectl" -> { "ENT-9686" }
-        args => "restart",
+        args => "graceful",
         if => and( "mission_portal_apache_config_repaired",
                    not( "apache_stop_after_new_staged_config_repaired" ) ),
         contain => in_shell,
--- a/templates/cf-apache.service.mustache
+++ b/templates/cf-apache.service.mustache
@@ -10,6 +10,7 @@ PartOf=cfengine3.service
 Type=forking
 ExecStart={{{vars.sys.workdir}}}/httpd/bin/apachectl start
 ExecStop={{{vars.sys.workdir}}}/httpd/bin/apachectl stop
+ExecReload={{{vars.sys.workdir}}}/httpd/bin/apachectl graceful
 PIDFile={{{vars.sys.workdir}}}/httpd/httpd.pid
 Restart=always
 RestartSec=10


craigcomstock:
@cf-bottom jenkins please, thanks.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10580/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10580/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10580/

craigcomstock:
bootstrap-tests is a known issue fixed for ENT-8264 in https://github.com/cfengine/mission-portal/pull/2321

98-delete-project failure
```
6	:	1712282346.958	     0.460	0	0	22	0	curl -k -i --show-error --fail -u admin:admin -H "content-type: application/json" -X DELETE https://54.216.44.96/api/build/projects/1 && echo
```

is a bit like what I was trying to fix but in this case it is a 500 error which would not be solved by a graceful restart of apache.

```
54.216.44.96	  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
54.216.44.96	                                 Dload  Upload   Total   Spent    Left  Speed
54.216.44.96	
54.216.44.96	  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
54.216.44.96	  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
54.216.44.96	curl: (22) The requested URL returned error: 500 Internal Server Error
```

Will check if there is another ticket for 500 errors during this test.

craigcomstock:
have re-opened 500 error tickets: ENT-11182, ENT-11391, ENT-11313

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10602/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10602/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10602/

cfe_internal/enterprise/mission_portal.cf
```diff
@@ -228,7 +228,7 @@ bundle agent mission_portal_apache_from_stage(config, staged_config)
                            "place only if we have stopped apache already.");
 
       "LD_LIBRARY_PATH=$(sys.workdir)/lib:$LD_LIBRARY_PATH $(sys.workdir)/httpd/bin/apachectl" -> { "ENT-9686" }
-        args => "restart",
+        args => "graceful",
```
nickanderson, line 231: Is this synchronous or async? I was thinking graceful is async. If it's async then whatever changes we expected to take effect from the apache reload won't be in effect at the end of this promise execution. I dunno if that is a problem, but it was something that came to mind.

nickanderson (commented):
I can see how this might make sense. I would be concerned that graceful is aysnc and then after the promise runs we wouldn't be in the expected state until some undetermined period of time later.

cfe_internal/enterprise/mission_portal.cf
craigcomstock: Good point. I will check into it and/or test it.

craigcomstock: from [URL](https://httpd.apache.org/docs/current/programs/apachectl.html)

```
The apachectl script returns a 0 exit value on success, and >0 if an error occurs. For more details, view the comments in the script.
```

It would seem it is sync.

Also interestingly it checks the config so it would seem we don't need to do that in policy as well?

```
This command automatically checks the configuration files as in configtest before initiating the restart to make sure Apache doesn't die.
```

This checking of the configuration files as in configtest is the same behavior for both restart and graceful according to the docs mentioned above.

I will test with bad config and see what happens.

vpodzime: We still need to know that the config is broken if that is the case, right?

vpodzime (approved):
Makes sense. :+1:

cfe_internal/enterprise/mission_portal.cf
vpodzime: > I dunno if that is a problem

I don't think it should be a problem. Even if it's async, I think it guarantees that **new** connections will be handled according to the new config when they are actually processed.

craigcomstock: @nickanderson @vpodzime I tried our patched /var/cfengine/httpd/bin/apachectl out and it seems graceful is as synchronous as the restart command. It will fail if config is broken like this:

```
root@ubuntu2004:/var/cfengine/share/GUI# /var/cfengine/httpd/bin/apachectl graceful
AH00526: Syntax error on line 1 of /var/cfengine/httpd/conf/httpd.conf:
Invalid command 'BROKEN', perhaps misspelled or defined by a module not included in the server configuration
root@ubuntu2004:/var/cfengine/share/GUI# echo $?
1
```

Looking through the source code at https://github.com/apache/httpd it seems that apachectl/httpd -k merely send good ol' UNIX signals to the httpd process.

in server/mpm_unix.c
```

    if (!strcmp(dash_k_arg, "graceful")) {
        if (!running) {
            printf("httpd not running, trying to start\n");
        }
        else {
            *exit_status = send_signal(otherpid, AP_SIG_GRACEFUL);
            return 1;
        }
    }
```

and send_signal() just returns the result of a kill() call

```
static int send_signal(pid_t pid, int sig)
{
    if (kill(pid, sig) < 0) {
        ap_log_error(APLOG_MARK, APLOG_STARTUP, errno, NULL, APLOGNO(00057)
                     "sending signal to server");
        return 1;
    }
    return 0;
}
```

So we aren't getting back any information about httpd or it's state other than if kill() succeeds then we know that apachectl/httpd could find the proper PID and was able to send the signal.

craigcomstock:
Looks like many platforms didn't have httpd running during bootstrap-tests restart test and sequential-tests during upgrades showed an error that may indicate why:

```
3.249.208.255	Failed to start cf-apache.service: Unit cf-apache.service is not loaded properly: Exec format error.
```

Will check it out.

craigcomstock:
typo in core, fixed, @cf-bottom jenkins retry please.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10603/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10603/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10603/

craigcomstock:
rhel-9 hub in 99-cfbs-init test said
```
54.74.76.234	curl: (22) The requested URL returned error: 500 Internal Server Error
```

Looking in support tarball and var/cfengine/httpd/logs/error_log I think this is a known issue

```
[Thu Apr 11 23:41:58.598136 2024] [php:warn] [pid 58068] [client 91.232.32.37:46788] PHP Warning:  Unknown: Failed to open stream: Permission denied in Unknown on line 0
[Thu Apr 11 23:41:58.598753 2024] [php:error] [pid 58068] [client 91.232.32.37:46788] PHP Fatal error:  Failed opening required '/var/cfengine/httpd/htdocs/api/dispatch.php' (include_path='.:') in Unknown on line 0
```

Known issue: https://northerntech.atlassian.net/browse/ENT-11182

---

Selenium failure: InventoryTest::testExcludeIncludeFilter, looks like a timing fluke in selenium test to me. Logged a ticket: https://northerntech.atlassian.net/browse/ENT-11546

craigcomstock:
Will wait to cherry pick for a week or so to see how/if this improves our CI. Will set a reminder for myself for 2024-04-19.

PR #2865, opened 2024-04-11
CFE-4374: Fix comparison that caused control_executor_mailfilter_*_configured to never be set

jmechnich:
It would be great if the fix could be backported to the LTS version as well.


cf-bottom:

nickanderson:
Thanks @jmechnich I amended your commit message here: https://github.com/cfengine/masterfiles/pull/2867 Closing this in favor of that.

nickanderson:
@jmechnich merged and back-ported to 3.21 here: https://github.com/cfengine/masterfiles/pull/2866

Thanks for the fix!

PR #2873, opened 2024-04-17
CFE-4359: Update regex strings to raw to squelch python 3.12 warnings (yum and zypper)

amousset:
Same as https://github.com/cfengine/masterfiles/pull/2853, for yum and zypper

### DIFF
--- a/modules/packages/vendored/yum.mustache
+++ b/modules/packages/vendored/yum.mustache
@@ -145,14 +145,14 @@ def list_updates(online):
             lastline = ""
 
         line = lastline + line
-        match = re.match("^\S+\s+\S+\s+\S+", line)
+        match = re.match(r"^\S+\s+\S+\s+\S+", line)
         if match is None:
             # Keep line
             lastline = line
             continue
 
         lastline = ""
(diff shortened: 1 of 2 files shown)


modules/packages/vendored/yum.mustache
```diff
@@ -145,14 +145,14 @@ def list_updates(online):
             lastline = ""
 
         line = lastline + line
-        match = re.match("^\S+\s+\S+\s+\S+", line)
+        match = re.match(r"^\S+\s+\S+\s+\S+", line)
```
craigcomstock, line 148: The warning you refer to is this I assume:

```
>>> match = re.match("^\S+\s+\S+\s+\S+", line)
<stdin>:1: SyntaxWarning: invalid escape sequence '\S'
```

Yes, as explained [here](https://docs.python.org/3/library/re.html)

> Also, please note that any invalid escape sequences in Python’s usage of the backslash in string literals now generate a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning) and in the future this will become a [SyntaxError](https://docs.python.org/3/library/exceptions.html#SyntaxError)

craigcomstock (approved):

PR #2890, opened 2024-05-24
ENT-11136: Refactor and improve policy fixing up SELinux context on FR SSH files (3.21)

nickanderson:
We need to make sure both `semanage fcontext -a` and `restorecon`
are done before we run `ssh-keygen`. Otherwise `ssh-keygen` may
be blocked by SELinux when trying to save/write the generated
key at first and only succeeds in a later agent run.

Ticket: ENT-11136
Changelog: Federated reporting policy fixes SELinux context of
           the ~cftransport/.ssh directory and its contents in a
           single agent run.
(cherry picked from commit ff392a9587ebc6bad4511750c6b01198e214208d)


nickanderson:
@cf-bottom jenkins, please

vpodzime (approved):
I still like this change :grin:

craigcomstock (approved):
looks ok, please test :)

PR #2796, opened 2023-12-12
ENT-10962: Fixed mission portal restoring changes from /var/cfengine/share/GUI

nickanderson:

PR #2892, opened 2024-05-24
ENT-10962: Fixed mission portal restoring changes from /var/cfengine/share/GUI (3.18)

nickanderson:


nickanderson:
For https://github.com/cfengine/masterfiles/pull/2796

PR #2902, opened 2024-06-05
ENT-11809: Allow access for cfapache to read host_specific.json

larsewi:
Allowed read access for cfapache to `$(sys.workdir)/cmdb/SHA=*/host_specific.json` through ACLs.

Ticket: ENT-11809
Changelog: Commit

**TODO:**
 - [x] Manual testing - read file as cfapache user
 - [ ] Pass CI - [Build Status](https://ci.cfengine.com/job/pr-pipeline/10894/)

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -321,6 +321,14 @@ bundle agent cfe_internal_update_cmdb
 @endif
 }
 
+body acl u_acl_cmdb_host_specific_json
+# @brief Allow read access for user cfapache
+# Mission Portal needs read access to these files (See ticket ENT-11809)
+{
+  acl_method => "append";
+  aces => { "user:cfapache:r" };
+}
+
 bundle agent cfe_internal_update_cmdb_data_distribution
 # @brief Ensure data is ready for agents to download
 {
@@ -390,6 +398,7 @@ bundle agent cfe_internal_update_cmdb_data_distribution
       # Write out the data for each host that had a data change
       "$(sys.workdir)/cmdb/$(_i)/host_specific.json"
         create => "true", # CFE-2329, ENT-4792
+        acl => u_acl_cmdb_host_specific_json, # ENT-11809
         template_data => mergedata("_get_cmdb_data_response_d[data][$(_i)]" ), # mergedata() is necessary in order to pick out a substructure, parsejson() is insufficient because expanding a key results in iteration of /values/ under that key
         template_method => "inline_mustache",
         edit_template_string => string_mustache( "{{$-top-}}", "_get_cmdb_data_response_d[data][$(_i)]" ),


cfe_internal/update/update_policy.cf
```diff
@@ -390,6 +398,7 @@ bundle agent cfe_internal_update_cmdb_data_distribution
       # Write out the data for each host that had a data change
       "$(sys.workdir)/cmdb/$(_i)/host_specific.json"
         create => "true", # CFE-2329, ENT-4792
+        acl => u_acl_cmdb_host_specific_json, # ENT-11809
         template_data => mergedata("_get_cmdb_data_response_d[data][$(_i)]" ), # mergedata() is necessary in order to pick out a substructure, parsejson() is insufficient because expanding a key results in iteration of /values/ under that key
         template_method => "inline_mustache",
```
vpodzime, line 403: I hope/believe we no longer use this promise to **render** these files. Is this the only place that manages permissions of these files?

larsewi: As far as I can tell, yes :)

vpodzime: I believe it's definitely time to drop it, I believe `cf-reactor` has proven itself good in keeping those files up to date and this policy doesn't take group CMDB into account so it couldn't be used anyway. However, we should have a promise for correct permissions on these files.

nickanderson: The bundle itself is guarded with !have_cf_reactor|cmdb_data_files_updates_done_in_policy.

```
    policy_server.enterprise_edition.(!have_cf_reactor|cmdb_data_files_updates_done_in_policy)::
      "cfe_internal_update_cmdb_data_distribution";
```

So, cf-reactor is taking care of this. The policy needs to remain until there are no supported versions who don't use cf-reactor for this.

nickanderson: That also means that the acl on this file will not be managed, since cf-reactor is taking care of things.

nickanderson (changes requested):
Since `bundle agent cfe_internal_cmdb_data_distribution` is not run when cf-reactor is responsible for managing the files, I don't believe this ACL will have any effect in current versions.

```
  methods:

    policy_server.enterprise_edition.(!have_cf_reactor|cmdb_data_files_updates_done_in_policy)::
      "cfe_internal_update_cmdb_data_distribution";
```

So, the ACL on that file should I think be handled elsewhere.

larsewi:
As @nickanderson mentioned, these changes has no effect. Hence, I'm closing this PR.

cfe_internal/update/update_policy.cf
larsewi: Thanks for the explanation 🚀

PR #2908, opened 2024-06-06
Fixed missing software inventory on RedHat based systems

craigcomstock:
A previous fix to add package inventory for AIX and adding ability to influence default package manager and inventory via augments introduced this bug.

The commit was: 87292179415a0d4770281715a850ef49d2991842

There was a typo that used the class cfe_yum_package_module_supported instead of cfe_python_for_package_modules_supported.

The first class is undefined anywhere in policy. The second class it the correct one.

Ticket: ENT-11850
Changelog: none

### DIFF
--- a/promises.cf.in
+++ b/promises.cf.in
@@ -98,7 +98,7 @@ body common control
 
       # We only define package_inventory on redhat like systems that have a
       # python version that works with the package module.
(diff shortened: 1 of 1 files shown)


vpodzime (approved):
ACK

craigcomstock:
acceptance_tests relies on nightlies and apparently ubuntu 22 testing-pr failed to produce a package.

PR #2918, opened 2024-07-04
Fixed error caused by function call in comment attribute

larsewi:
```
$ cf-agent -KI
   error: Comments can only be scalar objects, not 'call' in '$(cfe_internal_hub_vars.docroot)/static/.'
```

Ticket: None
Changelog: None


cfe_internal/enterprise/CFE_knowledge.cf
```diff
@@ -159,11 +159,9 @@ bundle agent cfe_internal_setup_knowledge
         depth_search => recurse_basedir("inf"),
         file_select => not_dir,
         perms => mog("0600", "$(def.cf_apache_user)", "$(def.cf_apache_group)"),
-        comment => concat( "Ensure permissions for",
-                           " $(cfe_internal_hub_vars.docroot)/static/*.",
-                           " This is where exported and scheduled reports",
-                           " generated by Mission Portal (temp files to ",
-                           " download or email)" );
+        comment => "Ensure permissions for $(cfe_internal_hub_vars.docroot)/static/*.
```
nickanderson, line 162: Yeah this should fix it, back to the way the comment was before I updated it to concat.

Alternatively, would remove the variable from the comment. I think it actually reads better without the path repeated (it's already in the promiser).

nickanderson (approved):
But, ill fix that later.

PR #2919, opened 2024-07-04
Fix comment of the $(cfe_internal_hub_vars.docroot)/static/ perms promise

vpodzime:
Ticket: ENT-11941
Changelog: None


vpodzime:
Doesn't work.

PR #2931, opened 2024-07-24
CFE-4426: Added inline docs showing valid values for method (field_operation) in body edit_field quoted_var

nickanderson:
This is just a small change so that a user doesn't have to hunt down the valid
options to know what's available.

Ticket: CFE-4426

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -1243,7 +1243,8 @@ body edit_field fstab_options(newval, method)
 body edit_field quoted_var(newval,method)
 # @brief Edit the quoted value of the matching line
 # @param newval The new value
-# @param method The method by which to edit the field
+# @param method The method by which to edit the field (append|prepend|alphanum|set|delete)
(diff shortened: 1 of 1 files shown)


lib/files.cf
```diff
@@ -1243,7 +1243,7 @@ body edit_field fstab_options(newval, method)
 body edit_field quoted_var(newval,method)
 # @brief Edit the quoted value of the matching line
 # @param newval The new value
-# @param method The method by which to edit the field
+# @param method The method by which to edit the field (append|prepend|alphanum|set|delete)
```
craigcomstock, line 1246:
```suggestion
# @param method The method by which to edit the field (append|prepend|alphanum|set|delete)
   see https://docs.cfengine.com/latest/reference-promise-types-files-edit_line-field_edits.html#field_operation
```

craigcomstock (commented):
maybe add a link to the SOURCE of the nice information :)

craigcomstock (approved):

PR #2936, opened 2024-07-30
Use --rows-per-insert instead of merge_inserts.awk

vpodzime:
Older versions pg_dump didn't have this option so we had to use our custom AWK script to get a significant speed-up of loading dumps from feeders on a superhub. All versions of PostgreSQL CFEngine Enterprise is shipped with now have pg_dump supporting the new --rows-per-insert option.

Changelog: Data dumping on Federated Reporting feeders no longer
           uses an AWK filter to merge INSERT lines in the dumps
Ticket: None


vpodzime:
jenkins: [Build Status](https://ci.cfengine.com/job/pr-pipeline/11078/)

craigcomstock (approved):
Nice. I don't mind AWK and I celebrate simplification. One less tool is a "good thing" (TM).

vpodzime:
> Nice. I don't mind AWK and I celebrate simplification. One less tool is a "good thing" (TM).

We still give users a chance to user their own custom AWK and SED scripts. :)

PR #2948, opened 2024-10-08
CFE-4442: Clarified need to specify namespace when defining custom policy update bundle

nickanderson:
This change clarifies the need to specify the namespace when defining a custom policy update bundle both in documentation in reports emitted by policy.

If namespace is not specified the bundle will not be found and will be unable to run.

### DIFF
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -42,7 +42,7 @@ bundle agent cfe_internal_update_policy
 
       # Look for a bundle that matches what the user wants
       "found_matching_user_specified_bundle"
-        slist => bundlesmatching( "$(def.mpf_update_policy_bundle)" );
+        slist => bundlesmatching( "^$(def.mpf_update_policy_bundle)$" );
 
   methods:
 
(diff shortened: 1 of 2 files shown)


nickanderson:
@craigcomstock could you take this for a spin? 

Also, the Warning message that I added, it's currently classed to `inform_mode|verbose_mode|DEBUG|DEBUG_cfe_internal_update_policy::` Do you think that's OK, or do you think that it should be emitted in any context?

cfe_internal/update/update_policy.cf
```diff
@@ -42,7 +42,7 @@ bundle agent cfe_internal_update_policy
 
       # Look for a bundle that matches what the user wants
       "found_matching_user_specified_bundle"
-        slist => bundlesmatching( "$(def.mpf_update_policy_bundle)" );
+        slist => bundlesmatching( "^$(def.mpf_update_policy_bundle)$" );
```
craigcomstock, line 45: This has no effect because `bundlesmatching()` is anchored. I suppose it is good for clarity though. :+1:

craigcomstock (approved):

nickanderson:
Yes, this should be picked to 3.24.x and 3.21.x. I have yet to do it.

nickanderson:
backported.

PR #2949, opened 2024-10-09
ENT-4400: Allowed blob image-src in the Mission Portal CSP rules

aleksandrychev:
Highchart export requires it

### DIFF
--- a/cfe_internal/enterprise/templates/httpd.conf.mustache
+++ b/cfe_internal/enterprise/templates/httpd.conf.mustache
@@ -215,7 +215,7 @@ LogLevel warn
     object-src 'none'; \
     frame-src 'self'; \
     child-src 'self'; \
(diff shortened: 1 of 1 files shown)


craigcomstock (approved):

craigcomstock (changes requested):
Won't this require the same change in buildscripts repo?

aleksandrychev:
> Won't this require the same change in buildscripts repo?

@craigcomstock I don't see any of CSP rules there https://github.com/cfengine/buildscripts/blob/master/deps-packaging/apache/httpd.conf

aleksandrychev:
> Won't this require the same change in buildscripts repo?

added whole rules set https://github.com/cfengine/buildscripts/pull/1504

craigcomstock (approved):

PR #2959, opened 2024-10-18
Added support for AIX System Resource Controller services promises

craigcomstock:
On AIX the default will use AIX System Resource Controller commands to manage service state.

```
bundle agent main
{
  services:
    "sendmail"
      service_policy => "start";
}
```

service_policy options supported are: start, stop and reload.

Ticket: CFE-4447
Changelog: title

### DIFF
--- a/lib/services.cf
+++ b/lib/services.cf
@@ -111,10 +111,12 @@ bundle agent standard_services(service,state)
 #
 # Else, if chkconfig is present, it will be used.
 #
-# Else, if the service command is available, if will be used.
+# Else, if the service command is available, it will be used.
 #
-# Else, if the svcadm command is available, if will be used. Note you
+# Else, if the svcadm command is available, it will be used. Note you
 # have to supply the full SMF service identifier.
+#
+# Else, if lssrc command is available (AIX), it will be used.
 #
 # Else, control is passed to `classic_services`.
 #
@@ -170,7 +172,9 @@ bundle agent standard_services(service,state)
       "chkconfig" expression => "!systemd._stdlib_path_exists_chkconfig";
       "sysvservice" expression => "!systemd.!chkconfig._stdlib_path_exists_service";
       "smf" expression => "!systemd.!chkconfig.!sysvservice._stdlib_path_exists_svcadm";
-      "fallback" expression => "!systemd.!chkconfig.!sysvservice.!smf";
+      # AIX System Resource Controller https://www.ibm.com/docs/en/aix/7.2?topic=concepts-system-resource-controller
+      "aix_src" expression => "_stdlib_path_exists_lssrc";
+      "fallback" expression => "!systemd.!chkconfig.!sysvservice.!smf.!aix_src";
 
       "have_init" expression => fileexists($(init));
 
@@ -268,6 +272,8 @@ bundle agent standard_services(service,state)
       classes => kept_successful_command;
 
   methods:
+    aix_src::
+      "aix_service" usebundle => aix_services($(service), $(state));
     fallback::
       "classic" usebundle => classic_services($(service), $(state));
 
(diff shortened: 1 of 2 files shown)


lib/services.cf
```diff
@@ -1098,3 +1104,74 @@ bundle agent classic_services(service,state)
...
+# Another failure case. Likely on next agent run (default of 5 minutes) the service will have stopped or we will try again.
+#    error: Finished command related to promiser '/usr/bin/stopsrc -s sendmail' -- an error occurred, returned 1
+#   notice: Q: "...in/stopsrc -s s": 0513-056 Timeout waiting for command response. If you specified a foreign host,
+# Q: "...in/stopsrc -s s": see the /etc/inittab file on the foreign host to verify that the SRC daemon
+# Q: "...in/stopsrc -s s": (srcmstr) was started with the -r flag to accept remote requests.
+bundle agent aix_services(service, desired_state)
+{
+  vars:
+    # current state can be: active, inoperative, stopping
+    "current_state" string => execresult("/usr/bin/lssrc -s $(service) | tail -1 | awk '{print $NF}'", useshell);
```
nickanderson, line 1150: I have some small amount of worry about cached result here, it *might* benefit from action => immediate but I think it should be ok without it since $(service) comes from the parameter and that won't be set until real 3 pass actuation and not during pre-eval.

lib/services.cf
```diff
@@ -1098,3 +1104,74 @@ bundle agent classic_services(service,state)
...
+      strcmp("$(desired_state)", "restart")
+    );
+    "needs_stop" expression => and(
+      strcmp("$(current_state)", "active"),
+      strcmp("$(desired_state)", "stop")
+    );
+
+  commands:
+    needs_start::
+      "/usr/bin/startsrc -s $(service)";
```
nickanderson, line 1167: Perhaps use $(paths.startsrc)

lib/services.cf
```diff
@@ -1098,3 +1104,74 @@ bundle agent classic_services(service,state)
...
+    needs_restart::
+      "/usr/bin/stopsrc -s $(service); /usr/bin/startsrc -s $(service)"
```
nickanderson, line 1169: perhaps use $(paths.stopsrc) and $(paths.startsrc)

lib/services.cf
```diff
@@ -1098,3 +1104,74 @@ bundle agent classic_services(service,state)
...
+      contain => useshell;
+    needs_stop::
+      "/usr/bin/stopsrc -s $(service)";
```
nickanderson, line 1172: perhaps use $(paths.stopsrc)

nickanderson (approved):
This looks OK to me. I commented a bit and have some non-blocking suggestions.

lib/services.cf
craigcomstock: Ok. :) Yeah, ideally folks shouldn't have to worry about it right?

What would be the problem about "cached result"?
I know that the commands in execresult are run a few times and cached in pre-eval and then in the real passes right?
Seems OK to me regardless as things will "work out"/stabilize eventually.

nickanderson: The problem with the cached result is that the state from pre-eval could be used instead of the state from when this is actuated.

Again, I don't think it applies here because $(service) would not be defined during pre-eval since it's only available as a parameter which will only happen during the normal 3 pass evaluation.

However, if this bundle happens to be called for the same service multiple times within the policy, then the cache is likely to apply which could cause some issue if you for example do:

start (but say it was already running), stop, start

In that second start the cached value from the first start would be used and with the knowledge it was running it may not try to start again.

craigcomstock: Yeah. I suppose it would be good to include a warning of some sort if you try to change the state of a service that has already been managed in the current agent run... kind of like a global class that signified e.g. "sshd_service_managed" and if you re-entered the aix_services() bundle and that class was set then log a warning that you shouldn't try to enforce more than one state on a service in a policy run.

craigcomstock: I made a ticket: https://northerntech.atlassian.net/browse/CFE-4448

craigcomstock:
@nickanderson do you want me to back-port this?

nickanderson:
Yeah I'd bring it to 3.24.x

craigcomstock:
https://github.com/cfengine/masterfiles/pull/2960 (3.24)

PR #2963, opened 2024-11-22
CFE-4450: Allow passing environment variables to package managers

amousset:
Allow passing options to the package managers modules through `env:NAME=VERSION` options.

Taking the easiest path. If it is considered acceptable I'll add a test case and documentation.

### DIFF
--- a/modules/packages/vendored/apt_get.mustache
+++ b/modules/packages/vendored/apt_get.mustache
@@ -295,8 +295,17 @@ def package_arguments_builder(is_apt_install):
         elif line.startswith("options="):
             global apt_get_options
             option =  line.split("=", 1)[1].rstrip()
+
             if option:
-              apt_get_options.append(option)
+                if option.startswith("env:"):
+                    env_option = option.split(":", 1)[1]
+                    env_var = env_option.split("=", 1)
+                    if len(env_var) == 2:
+                        env_name = env_var[0]
+                        env_value = env_var[1]
+                        os.environ[env_name] = env_value
+                else:
+                    apt_get_options.append(option)
 
     if name:
         args.extend(one_package_argument(name, arch, version, is_apt_install))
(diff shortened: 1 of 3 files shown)


nickanderson:
Do you have some explicit examples of the use case for each of the package managers as well?

amousset:
For APT, for example controling [needrestart's behavior](https://manpages.ubuntu.com/manpages/focal/man1/needrestart.1.html#environment). 

But the main use case is for non-standard packages that allow customizing the configuration through env vars used in the postinst script. Gitlab is one of them (https://docs.gitlab.com/omnibus/installation/index.html#set-up-the-initial-account):

```
sudo GITLAB_ROOT_EMAIL="<gitlab_admin@example.com>" GITLAB_ROOT_PASSWORD="<strongpassword>" EXTERNAL_URL="http://gitlab.example.com" apt install gitlab-ee
```

nickanderson:
From my side I think it makes sense, I will invite @olehermanse @craigcomstock to bike shed a bit.

craigcomstock (commented):
Seems fine to me. Seeing this change here kind of begs to have a common python script to share such repeated code.

Also, it's a shame to not have the same for non-python modules if they even need it. For example, apk for alpine, but I have never needed to use env vars so probably fine. We can leave the complex usage to the complex systems: apt, yum and zypper. ;)

Definitely was looking for documentation and examples (a test case is usually a great example) before I read your description. So with those two pieces this looks good to me!

Thanks for the contribution!

amousset:
I created some basic documentation in https://github.com/cfengine/documentation/pull/3374. I also took a look at the tests but it seems there are no tests covering the `options` attribute, and that there is no testing mechanism for "real module/fake package manager" that would be convenient here (there is "fake module/no PM" and "real module/real PM"). I'll try to take a closer look.

PR #2964, opened 2024-11-27
CFE-4452: Made system_log_level configurable via Augments

nickanderson:
This change exposes the variable default:def.control_common_system_log_level to
be configurable via Augments. This way you can adjust the level of logs that go
to the system log without having to modify vendored policy.

Ticket: CFE-4452
Changelog: Title

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1408,6 +1408,26 @@ Example definition in augments file:
 
 * Added in 3.22.0, 3.21.2
 
+### Configure the minimum log level for system log
+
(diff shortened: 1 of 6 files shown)


olehermanse (approved):
Tested it, works!

nickanderson:
> Tested it, works!

brushoff-mostinterestingman

nickanderson:
Should cherry pick to 3.24 and maybe 3.21, but not until after the upcoming releases are done.

PR #2980, opened 2025-01-06
ENT-12560: Add paths for the dmsetup, fdisk, and lshw commands

jimrt:
These hardware related commands - whose locations differ slightly between distros - are often used by CMDB discovery tools. By adding them to paths.cf, we make it easy to template out a sudoers file.

Changelog: Title


nickanderson (approved):
LGTM, thanks!

PR #2992, opened 2025-03-04
Added case if http/2(php-fpm binary present) no need for php_version variable for httpd.conf template

craigcomstock:
Ticket: ENT-11440
Changelog: none


cf-bottom:

nickanderson (approved):

vpodzime (approved):

craigcomstock:
http/2 is only in master so no need to cherry pick.

PR #3002, opened 2025-05-05
Fixed some ArchLinux missing paths

m-bouissou:
Simple PR to add a few missing paths for basic system commands for ArchLinux


craigcomstock:
Thanks for the contribution @m-bouissou. We are in the middle of making releases right now so will hold off on this for a bit.

PR #3003, opened 2025-05-05
Moved standard library paths bundle from common to agent type to simplify SELinux policy

craigcomstock:
This should simplify selinux policy where we want the agent to have broad access but other components like serverd, execd, hub to have more limited access.

The paths bundle checks for the existence of many commands and this operation can cause SELinux AVCs.

Ticket: ENT-12954
Changelog: title

### DIFF
--- a/cfe_internal/enterprise/CFE_hub_specific.cf
+++ b/cfe_internal/enterprise/CFE_hub_specific.cf
@@ -15,7 +15,7 @@
 ##################################################################
 
 @if minimum_version(3.12.0)
-bundle common cfe_hub_specific_file_control
+bundle agent cfe_hub_specific_file_control
 {
   vars:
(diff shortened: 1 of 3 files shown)


craigcomstock:
@cf-bottom jenkins please, thanks!

craigcomstock:
ubu24 hub had trouble installing, some problem with postgres starting. Looks like a flake.

nickanderson (commented):
Not sure how I feel about this one. My initial reaction is no. But, i think it's true that most use of paths is from cf-agent.

I need to ponder on it. I guess we will if we have not yet confirmed that we aren't using any of the paths in MPF stock and probably scour build modules as well.

Plausibly could add something, sort of like the extend paths capability to get certain paths into common availability if needed. I wonder, will this change negatively affect the ability of the paths bundle to be extended without modification?

PR #3013, opened 2025-06-13
CFE-4543: Made protocol_version configurable via Augments

nickanderson:
This change implements easy configuration of protocol_version in body common control for the standard MPF entries.

Ticket: CFE-4543

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1306,6 +1306,38 @@ Example definition in augments file:
 }
 ```
 
+### Specify the CFEngine protocol version to use
+
+By default CFEngine will negotiate the newest protocol version available. Configuring `protocol_version` will restrict the protocol to the specified version.
+
+```json
+{
+  "variables": {
+    "default:def.control_common_protocol_version": {
+      "value": "filestream"
+      }
+  }
+}
+```
+
+**Notes:**
+
+- Valid values for `protocol_version` can be extracted from the syntax-description output of `cf-promises`.
+
+  For example:
+
+  ```command
+  cf-promises --syntax-description=json | jq -r '.bodyTypes.common.attributes.protocol_version.range'
+  ```
+
+  ```output
+  (1|classic|2|tls|3|cookie|4|filestream|latest)
+  ```
+
+**History:**
+
+- Added in CFEngine 3.27.0
+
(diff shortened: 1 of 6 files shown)


MPF.md
```diff
@@ -1306,6 +1306,24 @@ Example definition in augments file:
 }
 ```
 
+### Specify the CFEngine protocol version to use
+
+By default CFEngine will negotiate the newest protocol version available. Configuring `protocol_version` will restrict the protocol too the specified version.
```
craigcomstock, line 1311:
```suggestion
By default CFEngine will negotiate the newest protocol version available. Configuring `protocol_version` will restrict the protocol to the specified version.
```

MPF.md
```diff
@@ -1306,6 +1306,24 @@ Example definition in augments file:
...
+
+```json
+{
+  "variables": {
+    "default:def.control_common_protocol_version": {
+      "value": "filestream"
```
craigcomstock, line 1317: Can we get a reference to the supported values?

craigcomstock (commented):
I think we need to make maintenance easier by cross-referencing core code as I mention in one of my comments. Could be a follow-up ticket as it would involve some C refactoring most likely and maybe a bit tricky there.

controls/def.cf
```diff
@@ -399,6 +399,12 @@ bundle common def
         comment => concat( "The minimum log level required for log messages to go to the",
                            " system log (e.g. syslog or Windows Event Log).",
                            " (critical|error|warning|notice|info)" );
+
+      "control_common_protocol_version_defined" -> { "CFE-4543" }
+        expression => isvariable( "default:def.control_common_protocol_version" ),
+        comment => concat( "Defines the protocol version to use for all outgoing",
+                           " connections.",
+                           " (1|classic|2|tls|3|cookie|4|filestream|latest)" );
```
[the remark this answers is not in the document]
nickanderson: Instead of embedding a comment to core code, I chose to embed a comment with a command to extract the current value from the syntax description with jq.

MPF.md
nickanderson: Absolutely!

craigcomstock (approved):

PR #3026, opened 2025-08-07
Added Ubuntu 24.04 and Debian 12 to standalone_self_upgrade.

rexberg:
I patched this in my MPF and it has been working fine.


nickanderson (approved):
Thank you!

PR #3034, opened 2025-08-14
ENT-13172: Stopped automatic mirroring of WORKDIR/share/GUI to Mission Portal docroot

nickanderson:
There are very many files under share/GUI. Mirroring these files has been
observed to take multiple 10s of seconds. Instead of actively mirroring by
default, now the files will only be mirrored when the class
default:mpf_enable_mission_portal_docroot_sync_from_share_gui is defined.

Ticket: ENT-13172

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1925,16 +1925,17 @@ memory related probes on policy servers:
 
 * Added in 3.10.2, 3.11.0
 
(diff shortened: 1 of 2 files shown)


nickanderson:
seems good to me from my manual testing. Significant improvement.

PR #3037, opened 2025-08-28
CFE-4582: Added fsattrs bodies for the immutable bit

nickanderson:
This introduces a new fsattrs immutable body for use with the files promise
type. When applied, it ensures that the promised file has the immutable (i)
attribute set. The agent will transparently manage the attribute, removing it
before any necessary file modifications and restoring it afterwards.

Ticket: CFE-4582
Changelog: Title

### DIFF
--- a/lib/files.cf
+++ b/lib/files.cf
@@ -1936,6 +1936,75 @@ body delete tidy
       rmdirs   => "true";
 }
 
+@if minimum_version(3.27)
+##-------------------------------------------------------
+## fsattrs
+##-------------------------------------------------------
+
+body fsattrs immutable(state)
+# @brief Ensure promised file is immutable or is not immutable
+# @param state A string representing a boolean choice in the range `(true|yes|on)` for true and `(false|no|off)` for false.. When true file immutability is enforced, when false file immutability is removed.
+#
+# **Example:**
+#
+# ```cf3
+# files:
+#   "/etc/motd"
+#     content => "There are, in fact, rules. You have been notified.",
+#     fsattrs => immutable("true"),
+#     perms => mog( "644", "root", "root" );
+# ````
+{
+      # When true the end state of the file is immutable. If the promise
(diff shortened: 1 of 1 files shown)


lib/files.cf
```diff
@@ -1936,6 +1936,31 @@ body delete tidy
       rmdirs   => "true";
 }
 
+##-------------------------------------------------------
+## fsattrs
+##-------------------------------------------------------
+
+body fsattrs immutable
+# @brief Ensure promised file is immutable
```
nickanderson, line 1944: Humm, maybe this should be immutable(true_or_false)

so you could fsatters => immutable("true");

could then have additional bodies if desired for is_immutable and is_not_immutable.

lib/files.cf
```diff
@@ -1936,6 +1936,75 @@ body delete tidy
       rmdirs   => "true";
 }
 
+@if minimum_version(3.27)
+##-------------------------------------------------------
+## fsattrs
+##-------------------------------------------------------
+
+body fsattrs immutable(CF_BOOL)
```
craigcomstock, line 1944:
```suggestion
body fsattrs immutable(state)
```

I would focus on the parameter name as what it represents not what type it is.

lib/files.cf
```diff
@@ -1936,6 +1936,75 @@ body delete tidy
...
+{
+      # When true the end state of the file is immutable. If the promise
+      # necessitates a change in the file, the agent will handle immutability
+      # transparently with the bit being removed just before modification
+      # and restored immediately after the atomic change is made.
+
+        immutable => "$(CF_BOOL)"; # The i attribute supported by chattr
+}
+
+body fsattrs is_immutable
```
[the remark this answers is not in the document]
nickanderson: yeah, I had considered that but instead preferred for the others to be more stand alone.

craigcomstock (approved):

PR #3041, opened 2025-09-04
Inventory Unique NFS Servers and Recommend consistency in definition

nickanderson:
Tickets: ENT-13223, ENT-13224

### DIFF
--- a/cfe_internal/recommendations.cf
+++ b/cfe_internal/recommendations.cf
@@ -107,6 +107,46 @@ bundle agent ignore_interfaces_rx_reccomendations
 }
 @endif
 
+bundle agent nfs_mount_recommendations
+# @brief Recommendations about configured NFS servers
+{
+  meta:
+
+      "tags" slist => { "cfengine_recommends" };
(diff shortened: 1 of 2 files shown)


cfe_internal/recommendations.cf
```diff
@@ -107,6 +107,46 @@ bundle agent ignore_interfaces_rx_reccomendations
...
+
+  vars:
+      "nfs_server_list"
+        slist => getvalues( "default:inventory_linux.nfs_server"),
+        depends_on => { "cfe_internal_inventory_mounted_nfs_server" };
+
+  classes:
+
+      # If we end up emitting the recommendation, then we define a class so that
+      # instructions about disabling these reports
```
craigcomstock, line 125:
```
      # If we end up emitting the recommendation, then we define a class so that
      # instructions about disabling these reports are also emitted.
```

craigcomstock (changes requested):
looks good. I think the comment needs a bit more text to be complete.

PR #3047, opened 2025-09-22
ENT-13243: Stopped enforcing permissions of public docroot scripts

nickanderson:
This change stops enforcing permissions of files in the scripts subdirectory of
the public docroot. The volume of files in this directory do not make it
sensible to check permissions during each agent execution.

To improve performance of policy runs on hubs this specific enforcement is
skipped unless a class is set. These files permission should be fully handled by
the package.

Ticket: ENT-13243

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -1954,6 +1954,28 @@ application docroot consists of only packaged files.
 * Added `default:mpf_enable_mission_portal_docroot_sync_from_share_gui` in
   CFEngine 3.27.0
 
+### Enable permission enforcement for files under `WORKDIR/httpd/htdocs/public/scripts`
+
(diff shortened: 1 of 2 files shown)


MPF.md
```diff
@@ -1954,6 +1954,28 @@ application docroot consists of only packaged files.
...
+        "enterprise_edition.am_policy_hub::"
+      ]
+    }
+  }
+}
+```
+
+**History:**
+
+* Stopped enforcing permissions for this directory by default in CFEngine 3.27.0.
```
craigcomstock, line 1975: seeing "this directory" here made me want a more specific reference to which directory above as I commented.

PR #3069, opened 2025-11-11
ENT-11784: dnf package module

nickanderson:
- Uses dnf python library for interfacing with dnf to install and remove packages
- Use rpm library for listing currently installed packages (it's seconds faster)

Ticket: ENT-11784

### DIFF
--- a/modules/packages/vendored/dnf.mustache
+++ b/modules/packages/vendored/dnf.mustache
@@ -1,0 +1,533 @@
+#!/usr/bin/env python3
+# Note: See lib/packages.cf `package_module dnf` use of the
+# `interpreter` attribute to use cfengine-selected-python.
+
+"""DNF Package Module for CFEngine.
+
+This module provides a Python-based interface between CFEngine and the DNF package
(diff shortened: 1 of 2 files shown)


modules/packages/vendored/dnf.mustache
```diff
@@ -0,0 +1,508 @@
...
+    """
+    for option in options:
+        option = option.strip()
+        if '=' in option:
+            key, value = [x.strip() for x in option.split('=', 1)]
+            if key == OPT_ENABLE_REPO:
+                if value in base.repos:
+                    base.repos[value].enable()
+            elif key == OPT_DISABLE_REPO:
+                if value in base.repos:
```
larsewi:
```suggestion
                if base.repos is not None and value in base.repos:
```

larsewi (changes requested):
Please fix pyright issues and also format with black. Then I will continue review afterwards

larsewi (approved):
LGTM 🚀

PR #3088, opened 2025-12-13
inventory/linux: Use Python script to check version

btriller:
Get rid of shells and other command and use Python itself for version check.
Get major/minor via tuple for compatibility with very ancient Python versions (sys.version_info.major etc introduced in 2.7) [1].

[1] https://docs.python.org/2/library/sys.html#sys.version_info

### DIFF
--- a/inventory/linux.cf
+++ b/inventory/linux.cf
@@ -75,8 +75,8 @@ bundle common inventory_linux
                            " acceptable ( 3.x or 2.4 or greater ) for package",
                            " modules. We use this guard to prevent errors",
                            " related to missing python modules."),
-        expression => returnszero("$(sys.bindir)/cfengine-selected-python -V 2>&1 | grep ^Python | cut -d' ' -f 2 | ( IFS=. read v1 v2 v3 ; [ $v1 -ge 3 ] || [ $v1 -eq 2 -a $v2 -ge 4 ] )",
-                                  useshell);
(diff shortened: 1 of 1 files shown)


craigcomstock (approved):
Looks reasonable. Thanks for the contribution! :)

inventory/linux.cf
```diff
@@ -75,8 +75,8 @@ bundle common inventory_linux
                            " acceptable ( 3.x or 2.4 or greater ) for package",
                            " modules. We use this guard to prevent errors",
                            " related to missing python modules."),
-        expression => returnszero("$(sys.bindir)/cfengine-selected-python -V 2>&1 | grep ^Python | cut -d' ' -f 2 | ( IFS=. read v1 v2 v3 ; [ $v1 -ge 3 ] || [ $v1 -eq 2 -a $v2 -ge 4 ] )",
-                                  useshell);
+        expression => returnszero("$(sys.bindir)/cfengine-selected-python -c 'import sys; vi = sys.version_info; exit(not (vi[0] > 2 or (vi[0] == 2 and vi[1] > 3)))'",
```
nickanderson: Does this work for you as well?
```suggestion
        expression => returnszero("$(sys.bindir)/cfengine-selected-python -c 'import sys; exit(sys.version_info < (2, 4))'",
```

nickanderson (commented):
Seems ok, but maybe it can be simplified further?

inventory/linux.cf
btriller: TIL yes, that works too.

nickanderson (approved):

PR #3099, opened 2026-01-12
ENT-13667: Fixed bad regex in packages promise method for pip

larsewi:
The packages promise method for pip fails with error:
```
error: Regular expression error: 'invalid range in character class' in expression '^([[:alnum:]-_]+\s\([\d.]+\))' (offset: 13)
```

This is probably a regression after upgrading from PCRE to PCRE2.

The regex engine complains about the hyphen, which has a special meaning
within the square brackets. The bug is easily fixed by escaping the
hyphen.

Ticket: ENT-13667
Changelog: Title

Back ported to https://github.com/cfengine/masterfiles/pull/3100

### DIFF
--- a/lib/packages.cf
+++ b/lib/packages.cf
@@ -414,9 +414,9 @@ bundle common pip_knowledge
   vars:
       "call_pip" string => "$(paths.path[pip])";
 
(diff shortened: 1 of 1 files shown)


nickanderson (approved):
Makes sense.

PR #3070, opened 2025-11-11
CFE-2852: dnf_group package module

nickanderson:
Ticket: CFE-2852

PR #3118, opened 2026-04-10
fix: update policy modules_presence bundle should work regardless of findfiles behavior related to trailing directory separator for directories

craigcomstock:
We don't want to process the directory anyway.
We want only files.

Ticket: CFE-4623
Changelog: none


craigcomstock:
@nickanderson fixed this in a different way here: https://github.com/cfengine/masterfiles/pull/3070/changes/ac3d657fd69367b562859ed27331cd70e8b1b778 so no need for this fix.

PR #3119, opened 2026-04-10
cfe-4623: fix modules_presence update bundle with regard to findfiles() returning directories with or without trailing slashes

craigcomstock:
Ticket: CFE-4623
Changelog: none

### DIFF
--- a/.github/workflows/bootstrap_policy_run_check.yml
+++ b/.github/workflows/bootstrap_policy_run_check.yml
@@ -3,11 +3,29 @@ on:
 
 jobs:
   bootstrap_policy_run_check:
+    strategy:
+      fail-fast: false
+      matrix:
+        cfengine_version: [ "3.21.8", "3.24.3", "3.27.0", "master" ]
+
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
           path: masterfiles
+      - name: Get Togethers
+        uses: cfengine/together-javascript-action@main
+        id: together
+        with:
+          myToken: ${{ secrets.GITHUB_TOKEN }}
+      - name: Checkout Core
+        if: ${{ steps.together.outputs.core != null }}
+        uses: actions/checkout@v4
+        with:
+          repository: cfengine/core
+          path: core
+          ref: ${{steps.together.outputs.core || github.base_ref || github.ref}}
+          submodules: recursive
       - name: Install, bootstrap, policy run
-        run: masterfiles/ci/docker-bootstrap-policy-run.sh
+        run: masterfiles/ci/docker-bootstrap-policy-run.sh ${{ matrix.cfengine_version }}
--- a/cfe_internal/update/update_policy.cf
+++ b/cfe_internal/update/update_policy.cf
@@ -801,20 +801,17 @@ bundle agent modules_presence
     # TODO: Remove this once sys.moduledir is always available (3.26+ only)
     "moduledir" string => ifelse(isvariable("sys.moduledir"), "$(sys.moduledir)", "$(sys.workdir)/modules");
     "_vendored_dir" string => "$(this.promise_dirname)$(const.dirsep)..$(const.dirsep)..$(const.dirsep)modules$(const.dirsep)packages$(const.dirsep)vendored$(const.dirsep)";
+    "_vendored_dir_filter" string => "$(this.promise_dirname)$(const.dirsep)..$(const.dirsep)..$(const.dirsep)modules$(const.dirsep)packages$(const.dirsep)vendored";
     "_override_dir" string => "$(this.promise_dirname)$(const.dirsep)..$(const.dirsep)..$(const.dirsep)modules$(const.dirsep)packages$(const.dirsep)";
     "_custom_template_dir" string => "$(this.promise_dirname)$(const.dirsep)..$(const.dirsep)..$(const.dirsep)modules$(const.dirsep)mustache$(const.dirsep)";
     "_vendored_paths" slist => findfiles("$(_vendored_dir)*.mustache");
     "_custom_template_paths" slist => findfiles("$(_custom_template_dir)*.mustache"), if => isdir( "$(_custom_template_dir)" );
-    "_package_paths"
-        with => ifelse( or( cf_version_between( "3.24.0", "3.24.3"),
-                            cf_version_at("3.26.0"),
-                            cf_version_at("3.27.0")), "$(const.dirsep)",
-                        ""),
-        slist => filter("$(_override_dir)vendored$(with)", _package_paths_tmp, "false", "true", 999);
-
+    "_package_paths_tmp" slist => findfiles("${_override_dir}*"),
+      comment => "We get a temp list of files that we have to filter out the vendored sub directory.";
 
     windows::
-      "_package_paths_tmp" slist => findfiles("$(_override_dir)*");
+      "_package_paths" slist => filter("\Q${_vendored_dir_filter}\E.*", _package_paths_tmp, "true", "true", 999),
+        comment => "The list will include the vendored sub-directory that we don't want so remove it.";
       "_vendored_modules" slist => maplist(regex_replace("$(this)", "\Q$(_vendored_dir)\E(.*).mustache", "$1", "g"), @(_vendored_paths));
       "_override_modules" slist => maplist(regex_replace("$(this)", "\Q$(_override_dir)\E(.*)", "$1", "g"), @(_package_paths));
       # replace single backslashes in a windows path with double-backslashes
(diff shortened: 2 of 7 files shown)


cf-bottom:

craigcomstock:
this fixes valgrind-checks in jenkins also: [Build Status](https://ci.cfengine.com/job/pr-pipeline/13512/)

cfe_internal/update/update_policy.cf
```diff
@@ -808,7 +808,7 @@ bundle agent modules_presence
     "_package_paths"
         with => ifelse( or( cf_version_between( "3.24.0", "3.24.3"),
                             cf_version_at("3.26.0"),
-                            cf_version_at("3.27.0")), "$(const.dirsep)",
+                            cf_version_minimum("3.27.0")), "$(const.dirsep)",
```
nickanderson: I guess it will be nice to document the exact last version it's present by switching to between when it's fixed.

nickanderson (approved):
Cool, I think we could tweak the condition further

cfe_internal/update/update_policy.cf
craigcomstock: Yeah, we need to be specific and clear. findfiles() returned directories with trailing slashes for a long time, then didn't (3.24.0 through 3.27.0) and then went back to the old behavior of trailing slashes at 3.28.0 forward. That's how I understand it since we only recently bumped the libntech version to get the change from there in master.

nickanderson: I don't understand why it's not failing the pr with the update in master. Shouldn't the trailing slash be present from master now?

nickanderson (approved):
I pushed a change to the version detection to your pr. should be good and ready for squash.

olehermanse:
The core PR this relies on has been merged; https://github.com/cfengine/core/pull/6069

.github/workflows/bootstrap_policy_run_check.yml
```diff
@@ -3,11 +3,29 @@ on:
 
 jobs:
   bootstrap_policy_run_check:
+    strategy:
+      fail-fast: false
+      matrix:
+        cfengine_version: [ "3.21.8", "3.24.3", "3.27.0", "master" ]
```
olehermanse, line 9: Since this action supports "together" PRs, I guess we should not use cf-remote master (nightlies) here.

When we make PRs which actually require a PR in both masterfiles and core at the same time, I think using the nightly in cf-remote will be outdated / wrong in most cases.

.github/workflows/bootstrap_policy_run_check.yml
```diff
@@ -3,11 +3,29 @@ on:
...
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
           path: masterfiles
+      - name: Get Togethers
+        uses: cfengine/together-javascript-action@main
+        id: together
+        with:
+          myToken: ${{ secrets.GITHUB_TOKEN }}
```
olehermanse, line 21: What is this token used for?

.github/workflows/bootstrap_policy_run_check.yml
olehermanse: Another option, maybe crazy, could be to use nightlies when there is no together, but use repo checkout when the user has specified a together link.

olehermanse:
Currently, the bootstrap_policy_run_check succeds on master, but valgrind tests fails. This is because cf-remote uses a (outdated) nightly build, while the valgrind test uses a checkout of master.

AFAICT the valgrind failure is correct, the issue has not been fixed for master, it is only working on older nightly builds.

.github/workflows/bootstrap_policy_run_check.yml
craigcomstock: it allows the togethers action to operate properly

craigcomstock:
> Currently, the bootstrap_policy_run_check succeds on master, but valgrind tests fails. This is because cf-remote uses a (outdated) nightly build, while the valgrind test uses a checkout of master.
> 
> AFAICT the valgrind failure is correct, the issue has not been fixed for master, it is only working on older nightly builds.

as far as I can tell, the build 345 that cf-remote is using should have the fix for findfiles() in libntech. That fix went in Apr 7 and build 345 is from Apr 10. https://github.com/cfengine/core/pull/6067

.github/workflows/bootstrap_policy_run_check.yml
craigcomstock: I imagined that valgrind check could take care of the build-from-source including togethers and this bootstrap test is more of a "check against reasonably used binary versions".

craigcomstock: I did consider that "master" in this list might translate to a from-source-and-togethers build but... valgrind does handle that case already. Probably this job should be renamed "bootstrap test against released (and master nightly) packages"

nickanderson (approved):
🏆

craigcomstock:
jenkins no tests canaries only: [Build Status](https://ci.cfengine.com/job/pr-pipeline/13526/)

.github/workflows/ci.yml
```diff
@@ -12,9 +12,9 @@ jobs:
...
-    uses: ./.github/workflows/tests.yml
-  valgrind:
-    needs: bootstrap_policy_run_check
-    uses: ./.github/workflows/valgrind.yml
+#  tests:
+#    needs: bootstrap_policy_run_check
+#    uses: ./.github/workflows/tests.yml
+#  valgrind:
+#    needs: bootstrap_policy_run_check
+#    uses: ./.github/workflows/valgrind.yml
```
olehermanse: Why are these disabled?

olehermanse (changes requested):
Looks good overall, just 1 question about why some things were commented out - looks like a leftover.

.github/workflows/ci.yml
craigcomstock: ah, right, will re-enable.

craigcomstock: fixed

craigcomstock:
I manually tested with a script like this to make sure override package modules were copied with a script like this where I copied all of the vendored package modules into test-module-overrides/ and added a comment including TEST_OVERRIDE

```
#!/usr/bin/env bash
set -ex
./autogen.sh
sudo rm -rf /var/cfengine/modules
sudo rm -rf /var/cfengine/masterfiles
sudo make install
sudo cp test-module-overrides/* /var/cfengine/masterfiles/modules/packages/
sudo /var/cfengine/bin/cf-agent -IB $(hostname -i) | tee bootstrap.log
sudo /var/cfengine/bin/cf-agent -Kdf update.cf -DDEBUG | tee update.log
```

todo: automate something like the above script: https://northerntech.atlassian.net/browse/CFE-4656

ci/docker-bootstrap-policy-run.sh
```diff
@@ -5,27 +5,41 @@ set -ex
...
-if docker images | grep mpf; then
-  docker rmi mpf
+if docker images | grep "$image_name"; then
+  docker rmi "$image_name"
+fi
+
+if [ -d "${NTECH_ROOT}"/core ]; then
+  docker build -t "$image_name" --build-arg CFENGINE_VERSION="$CFENGINE_VERSION" -f "${NTECH_ROOT}"/masterfiles/ci/bootstrap-policy-run.source.Dockerfile  "${NTECH_ROOT}"
+else
+  docker build -t "$image_name" --build-arg CFENGINE_VERSION="$CFENGINE_VERSION" -f "${NTECH_ROOT}"/masterfiles/ci/bootstrap-policy-run.cfremote.Dockerfile  "${NTECH_ROOT}"
```
olehermanse, line 29: Just a thought I had yesterday - we could choose to do both in this case (if not together, test with both cf-remote and repo). But it will be more costly / slow, maybe not worth it :)

olehermanse (approved):
🚀

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13527/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13527/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13527/

cfe_internal/update/update_policy.cf
```diff
@@ -801,28 +801,24 @@ bundle agent modules_presence
     # TODO: Remove this once sys.moduledir is always available (3.26+ only)
     "moduledir" string => ifelse(isvariable("sys.moduledir"), "$(sys.moduledir)", "$(sys.workdir)/modules");
     "_vendored_dir" string => "$(this.promise_dirname)$(const.dirsep)..$(const.dirsep)..$(const.dirsep)modules$(const.dirsep)packages$(const.dirsep)vendored$(const.dirsep)";
+    "_vendored_dir_filter" string => "$(this.promise_dirname)$(const.dirsep)..$(const.dirsep)..$(const.dirsep)modules$(const.dirsep)packages$(const.dirsep)vendored";
```
nickanderson, line 804: Dpes this need wrapped with \Q \E  to make it literal when used as a regex?

craigcomstock: yeah, probably, good catch. I will test on windows. :)

nickanderson (approved):

nickanderson:
Waiting on separate windows test.

cfe_internal/update/update_policy.cf
craigcomstock: I tested this change on windows and it seems to work well.

PR #3129, opened 2026-04-15
Reformatted some initial files with cfengine format

olehermanse:


inventory/debian.cf
```diff
@@ -5,88 +5,126 @@ bundle common inventory_debian
...
+          "\s*#[^\n]*",
+          # comments
+          "=",
+          # split
+          100,
+          # maxentries
+          2048,
+        );
+
+      # maxbytes
```
olehermanse, line 32:
```suggestion
          2048,
          # maxbytes
        );

```

PR #3135, opened 2026-04-20
CFE-4648: Add smtpport support to masterfiles

nickanderson:
Instruments support for the smtpport attribute in executor control,
matching the existing pattern for smtpserver.

- Added def.smtpport variable with default value "25"
- Added smtpport to body executor control in cf_execd.cf
- Documented smtpport in MPF.md with augments example
- Can be overridden via augments like other mail settings

### DIFF
--- a/MPF.md
+++ b/MPF.md
@@ -900,6 +900,27 @@ This setting can be customized via Augments, for example:
 }
 ```
 
+### smtpport
+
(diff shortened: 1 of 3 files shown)


nickanderson:
@basvandervlies ^

basvandervlies:
@nickanderson  nice! Thanks for adding as configurable parameter to def.json

larsewi (approved):

